PR #1, opened 2016-08-30
Initial setup

millerts:
These changes are just getting the initial scaffolding in place so feature development can begin.


.gitignore
```diff
@@ -0,0 +1,3 @@
+**/*.iml
+**/.idea/**
+**/target
```
jtreuting: I've used gitignore.io in the past to generate these and it seems to cover some of the more random ones that come up.  Also, would be nice to include entries for eclipse for other open source folks that might want to contribute.  

https://www.gitignore.io/api/java%2Cintellij%2Ceclipse%2Cmaven

The only issue i have with this auto-generated one is that it doesn't do the ** like you have, so might want to alter the one from above with ** after.

PR #5, opened 2016-09-09
Feature/open exam

millerts:
This contains:

* Exam follows pattern of other services
* DB migrations via a maven profile
* HealthController
* getExam goes against the database
* added missing tests and integration tests


README.md
```diff
@@ -1,3 +1,25 @@
...
+### Normal Build Actions
+
+* **mvn clean install** 
+	* standard build
+	* runs unit tests and builds jars
+* **mvn clean install -Dintegration-tests.skip=false**
+	* Uses in memory database and stands up Spring container
+	* Runs any integration tests that have *IntegrationTests* in the class name
+
+#### Database mirations
```
jeffjohnson9046: "mirations" should be "migrations"

client/src/main/java/tds/exam/Exam.java
```diff
@@ -1,18 +1,110 @@
...
+import java.math.BigInteger;
+import java.time.LocalDateTime;
 import java.util.UUID;
 
 /**
  * Class representing an exam
  */
 public class Exam {
-    private final UUID id;
+    private int id;
```
jtreuting: This might be preference, but I like using examId here instead of a generic id for each type.  But that is more tied to wanting the database to use that name for the PK then having it map to the object directly.  Might be more of a .NET approach too.

Thoughts?

client/src/main/java/tds/exam/Exam.java
```diff
@@ -1,18 +1,110 @@
...
+    private UUID uniqueKey;
```
jtreuting: what is the difference between id and unique key?

millerts: id is the numeric primary auto incrementing key.  uniqueKey is the unique varbinary UUID for an exam.  @jeffjohnson9046 and I discussed getting rid of the UUID for exams but that would make it hard to migrate existing data if there is some out there.

Please help with the name.  Due to the time constraints I didn't add it but in the near future that int id will be exposed to users.  So maybe that should be examId?

jtreuting: I personally like the auto increment id to user the name of the class in it, so i like examId

service/src/main/resources/db/migration/V11473361910__exam_create_table_exam.sql
```diff
@@ -0,0 +1,23 @@
+/***********************************************************************************************************************
+  File: V11473361910_exam_create_table_exam.sql
+
+  Desc: Creates the initial exam database table
+
+***********************************************************************************************************************/
+USE exam;
+
+CREATE TABLE IF NOT EXISTS exam(
+  id INT NOT NULL AUTO_INCREMENT,
```
jeffjohnson9046: So... I'm going back & forth on this surrogate key + "natural key" thing a bit.  As we've said, none of us would normally use a `UUID` as a primary key.  However, I'm concerned that the term "id" is going to become overloaded.  For example, I almost put a comment on the `ExamController.getExamById` method citing that it should be named `getExamByKey`.  However, to the caller/consumer `uniqueKey` == `id`.  So I could see some confusion when a person refers to the `id` but really means the `uniqueKey` or vice versa.  Because words mean things and language is important.

Some other [devil's advocate] thoughts:
* We already effectively have a unique id primary key: the UUID that represents the exam.  Granted, it's a key that's been difficult to deal with.  In some ways, it seems superfluous to have a surrogate key that auto-increments.
* We'll be taking up a bit more space in the database server's memory; now the primary key (an `int`) _and_ the uniqueKey (a `varbinary(16)`, which we'll be querying against almost all of the time), where as before we'd just have the `varbinary(16)` unique key.  This means fewer index records can fit in memory, which could cause more I/O against disk.
  * On the other hand, using the integer `id` internally to establish and enforce relationships within the database will be a huge win for identifying associated records and debugging/troubleshooting.
  * The mitigating factor for this is the nature of our system; rarely will we be reading "old" records.  I honestly wouldn't expect a "hot"/"primed" system (i.e. one that is in the midst of a Session) to be forcing the database server to go to disk.

Not advocating any action now; just sorta thinking out loud more than anything.  I suppose the pros outweigh the cons (`ints` and `bigints` being easier to reason about than a blob stored in a field, insert performance should be better), but... I dunno.

client/src/main/java/tds/exam/Exam.java
millerts: Sounds good.

service/src/test/java/tds/exam/controllers/ExamControllerTest.java
```diff
@@ -1,38 +0,0 @@
...
-    public void setUp() {
-        examService = mock(ExamService.class);
-        controller = new ExamController(examService);
-    }
-
-    @After
-    public void tearDown() {}
-
-    @Test
-    public void anExamCanBeReturnedForWithValidId() {
```
jeffjohnson9046: Choose either "For" or "With" in this method name.

service/src/main/resources/db/migration/V11473361910__exam_create_table_exam.sql
jeffjohnson9046: The `id` column should probably be an `BIGINT`, just in case.  Alternately, we could consider using an `UNSIGNED INT`; I think we'll only want positive values in there, and an `UNSIGNED INT` should give us some breathing room; especially if we have an archiving strategy to clear out the tables.

jtreuting:
looks good

PR #7, opened 2016-09-21
Feature/open exam

millerts:
This only contains the code for checking if an existing or a new exam can be created.  It will always return a stubbed exam.


client/src/main/java/tds/exam/OpenExamRequest.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @NotNull
+    @Min(0)
+    private int numberOfDaysToDelay;
+
+    private Long proctorId;
+
+    private String guestAccommodations;
+
+    //TODO - find what this means
```
millerts: I will resolve this TODO before closing the open exam task.

service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -17,4 +19,6 @@
      * @return {@link Exam} otherwise null
      */
     Optional<Exam> getExam(UUID uuid);
+
+    Response<Exam> openExam(OpenExamRequest openExamRequest);
```
millerts: I will add in the missing JavaDocs when this actually opens an exam in the db

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
+                throw new IllegalArgumentException(String.format("Could not find student for %s", openExamRequest.getStudentId()));
+            }
+        }
+
+        Session currentSession = sessionOptional.get();
+
+        //Previous exam is retrieved in lines 5492 - 5530 and 5605 - 5645 in StudentDLL
+        Optional<Exam> previousExamOptional = examQueryRepository.getLastAvailableExam(openExamRequest.getStudentId(), openExamRequest.getAssessmentId(), openExamRequest.getClientName());
+
+        boolean canOpenPreviousExam = false;
```
jtreuting: This variable has the same name as a function in this class as well.  i didn't even know you could do that, but I think it could be confusing.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
+        if (previousExamOptional.isPresent()) {
+            Pair<Boolean, Optional<ValidationError>> canOpenPreviousExamPair = canOpenPreviousExam(previousExamOptional.get(), currentSession);
+
+            if (canOpenPreviousExamPair.getRight().isPresent()) {
+                return new Response<Exam>(canOpenPreviousExamPair.getRight().get());
```
jtreuting: I don't love using Pair because it feels hard to read when referring to `getRight()` and `getLeft()`.  Using an internal class with getters like `getCanOpen()` and `getValidationErrors()` feels cleaner to me.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
+        }
+
+        Exam exam;
+        if (canOpenPreviousExam) {
+            //Open previous exam
+            LOG.debug("Can open previous exam");
+            exam = new Exam.Builder().withId(previousExamOptional.get().getId()).build();
+        } else {
+            //Line 5602 in StudentDLL
+            Optional<Extern> externOptional = sessionService.getExternByClientName(openExamRequest.getClientName());
```
jtreuting: This is probably tangential to this PR, but getting an Extern doesn't really mean much just looking at it without knowing more details about the database and how it is used.  I think it is a naming thing.  Maybe we could add a comment about what getting the Extern will help us determine since just reading the code doesn't really show that, or maybe more appropriately we could change the name of Extern at some point.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
     }
 
     @Override
     public Optional<Exam> getExam(UUID id) {
         return examQueryRepository.getExamById(id);
     }
+
+    @Override
+    public Response<Exam> openExam(OpenExamRequest openExamRequest) {
+        Optional<Session> sessionOptional = sessionService.getSession(openExamRequest.getSessionId());
```
jtreuting: I tried to see if it was considered convention to name Optional variables with the Optional suffix and couldn't find any sort of answer.  I can see it making sense because a logical thing to do next is to call `sessionOptional.get()` and set that to a variable which would obviously need a different name.  That being said, it feels a little old school like how variable names used to be `strName` or `blnIsOpen`.  So wondering if it is the norm in Java I guess.

When doing my Googling on this, I found a few places that suggested method naming convention for when Optional is returned.  One suggestion was that methods starting with get would return the object or throw, and find would return Optional if not found.  So that would mean naming this `sessionService.findSession()` instead.  This is probably a conversation for offline if we want to discuss.  I find it an interesting idea.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
+        Instant startTime = Instant.now();
+
+
+
+        return null;
+    }
+
+    private Pair<Boolean, Optional<ValidationError>> canOpenPreviousExam(Exam previousExam, Session currentSession) {
+        //Port of Student.DLL lines 5526-5530
+        if (ExamStatusCode.STAGE_CLOSED.equals(previousExam.getStatus().getStage())) {
```
jtreuting: `previousExam.getStatus()` could return null so this would throw in that case.  Is that what we would expect, or if there is no status would we expect to return false that it can't be opened?  Probably is exceptional and should throw.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1,25 +1,198 @@
...
+        //Port of Student.DLL lines 5555-5560
+        if (ExamStatusCode.STAGE_INACTIVE.equals(previousExam.getStatus().getStage())) {
+            return Pair.of(true, Optional.empty());
+        }
+
+        /*
+        The below code is a straight port of the legacy StudentDLL._CanOpenExistingOpportunity_SP.  It was a little hard to follow the application
+        reasons for this logic. StudentDLL lines 5569 - 5589
+         */
+        boolean daysSinceLastChange = false;
```
jtreuting: The name of this makes me think it should be an integer.  The bool is really something like `isOneOrMoreDaysSincePreviousExam` (now that is descriptive!)

service/src/main/java/tds/exam/services/impl/SessionServiceImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        UriComponentsBuilder builder =
+            UriComponentsBuilder
+                .fromHttpUrl(String.format("%s/%s", examServiceProperties.getSessionUrl(), sessionId));
+
+        Optional<Session> sessionOptional = Optional.empty();
+        try {
+            final Session session = restTemplate.getForObject(builder.toUriString(), Session.class);
+            sessionOptional = Optional.of(session);
+        } catch (RestClientException rce) {
+            LOG.debug("Exception thrown when retrieving session", rce);
```
jtreuting: LOG.error instead since this is a problem that we would want know about in production if the services can't communicate.

service/src/main/java/tds/exam/web/advice/ExceptionAdvice.java
```diff
@@ -37,10 +37,26 @@
...
+    @ExceptionHandler(IllegalStateException.class)
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
+    @ResponseBody
+    ResponseEntity<ExceptionMessageResource> handleIllegalStateException(final IllegalStateException ex) {
+        return new ResponseEntity<>(
+            new ExceptionMessageResource(HttpStatus.INTERNAL_SERVER_ERROR.toString(), ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+
+    @ExceptionHandler(IllegalArgumentException.class)
+    @ResponseStatus(value = HttpStatus.BAD_REQUEST)
```
jtreuting: Combine this handler with the previous 400 handler for MethodArguemntTypeMismatchExpetion.class.  You can pass an Array of Exceptions to the ExceptionHandler

client/src/main/java/tds/exam/ExamStatusCode.java
```diff
@@ -0,0 +1,55 @@
...
+        return description;
+    }
+
+    private ExamStatusCode(Builder builder) {
+        status = builder.status;
+        stage = builder.stage;
+        description = builder.description;
+    }
+
+    public static class Builder {
```
millerts: No, we don't.  I wrote this class before we had our discussion.  I'll change it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: No, it isn't convention.  Honestly, it was the suggested one from IntelliJ so I just used it.  The Optional itself is so inconsequential I didn't spend much time thinking about it.  Since the code no longer checks if session or student are empty at the some time I will use the orThrow here instead.

That being said it doesn't fix the previousExamOptional below.  I like 'find' or 'fetch'.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: word.  I'll fix it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: I'm all for renaming it.  I kept it the same name as the legacy code.  After looking at the data in the view it seems like it represents Client & Session configuration record. ExternalClientSessionConfiguration?

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -31,18 +42,72 @@ public ExamQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDat
...
+                    .withStatus(rs.getString("status"))
+                    .withDescription(rs.getString("description"))
+                    .withStage(rs.getString("stage"))
+                    .build())
+                .build();
+        }
+
+        private Instant mapTimezoneToInstant(ResultSet rs, String columnLabel) throws SQLException {
+            Timestamp t = rs.getTimestamp(columnLabel);
+            return t != null ? t.toLocalDateTime().toInstant(ZoneOffset.UTC) : null;
```
millerts: @jeffjohnson9046 and I have already done it in each of our code bases.  I'll add a class in common.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Actually, this method is not called unless the previous Exam can be found (line 71).  Since this is a private method I don't do a check.  If we want to make sure we check for null I'll add a not null check and throw at the beginning of the method, but I should do it for all arguments not just exam.

jtreuting: I was talking about the result from previousExam.getStatus() and not previousExam itself.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
jeffjohnson9046: Yeah - we'll probably need a different class in common to handle mapping `Timestamp`s to `Instant`s.  Since we're using builders or constructors for immutable objects, the `InstantPropertyEditorSupport` won't help us here (since the JDBC template expects setters to be present for hydrating the POJO from the database).

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: I'd consider naming it `atLeastOneDaySinceChange` or even `atLeastOneDaySincePreviousAttempt` (assuming we want to change `timesTaken` to `attempt`).  Then the `if` statement sorta reads like the intent.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -90,7 +90,7 @@ public Exam mapRow(ResultSet rs, int rowNum) throws SQLException {
                 .withSessionId(UuidAdapter.getUUIDFromBytes(rs.getBytes("session_id")))
                 .withAssessmentId(rs.getString("assessment_id"))
                 .withStudentId(rs.getLong("student_id"))
-                .withTimeTaken(rs.getInt("times_taken"))
+                .withAttempts(rs.getInt("times_taken"))
```
jeffjohnson9046: I think if the property on the `Exam` class is renamed to `attempts`, we should also rename the column in the database from `times_taken` to `attempts`.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -186,10 +186,10 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
 
             if (previousExam.getDateCompleted() != null) {
                 Duration duration = Duration.between(previousExam.getDateChanged(), Instant.now());
-                if (LegacyComparer.lessThan(previousExam.getTimeTaken(), openExamRequest.getMaxOpportunities()) &&
+                if (LegacyComparer.lessThan(previousExam.getAttempts(), openExamRequest.getMaxOpportunities()) &&
```
jeffjohnson9046: While we're at it, I think it might make sense to rename `getMaxOpportunities` on the `OpenExamRequest` to `getMaxAttempts` or `getMaximumAttempts`.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -186,10 +186,10 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
                     LegacyComparer.greaterThan(duration.get(DAYS), openExamRequest.getNumberOfDaysToDelay())) {
                     return Optional.empty();
-                } else if (LegacyComparer.greaterOrEqual(previousExam.getTimeTaken(), openExamRequest.getMaxOpportunities())) {
+                } else if (LegacyComparer.greaterOrEqual(previousExam.getAttempts(), openExamRequest.getMaxOpportunities())) {
                     return Optional.of(new ValidationError(ValidationErrorCode.MAX_OPPORTUNITY_EXCEEDED, "Max number of opportunities for exam exceeded"));
```
jeffjohnson9046: Maybe change "Max number of opportunities for exam exceeded" to "Max number of attempts for exam exceeded"?

millerts:
All requested changes have been made

PR #10, opened 2016-10-07
Feature/check approval

jeffjohnson9046:
Implementation of the `checkApproval` method as described by TDS-210.

**NOTES**
* Added the `ExamApprovalResource` for consistency, even though we're likely moving away from returning data to the caller when the entity -> resource/"view" mapping is one to one (as it is in this case).  Will come back and clean this up later.
* Changes I made in `ExamQueryRepositoryImpl`:
  * `getExamById`:  Added `ORDER BY e.id DESC LIMIT 1`.  There can obviously be multiple records with the same exam id (e.g. when status changes), so we want to get the most recent "version" of the exam (i.e. the exam record with the highest auto-incremented ID).
  * `getLastAvailableExam`: changed `ORDER BY e.exam_id` to `ORDER BY e.id`.  The `exam.exam_id` is `VARBINARY(16)`/`UUID`, thus might not necessarily be in the right sequence.  The `exam.id` column is auto-incrementing, so ordering by that should get the most recent exam based on the query criteria.
* Changes I made in `shouldOpenPreviousExamIfSessionIdSame`:
  * Had to make the session id for `currentSession` and `previousSession` the same.  Looking at the code (current and legacy), this can happen and is one of the conditions for successfully opening a previous exam.  I hope this is what that unit test is trying to express.


service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -77,30 +86,37 @@ public ExamQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDat
...
+                "   exam e \n" +
+                "JOIN \n" +
+                "   exam_status_codes esc \n" +
+                "   ON esc.status = e.status \n" +
+                "WHERE \n" +
+                "   e.date_deleted IS NULL \n" +
+                "   AND e.student_id = :studentId \n" +
+                "   AND e.assessment_id = :assessmentId \n" +
+                "   AND e.client_name = :clientName \n" +
+                "ORDER BY \n" +
```
jeffjohnson9046: Changed `ORDER BY e.exam_id` to `ORDER BY e.id`.  The `exam.exam_id` is `VARBINARY(16)`/`UUID`, thus might not necessarily be in the right sequence.  The `exam.id` column is auto-incrementing, so ordering by that should get the most recent exam based on the query criteria.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -310,12 +329,15 @@ public void shouldOpenPreviousExamIfSessionIdSame(){
         openExamRequest.setClientName("SBAC-PT");
         openExamRequest.setMaxAttempts(5);
 
-        Session currentSession = new Session();
-        currentSession.setType(2);
+        Session currentSession = new Session.Builder()
+                .withId(sessionId)
```
jeffjohnson9046: Had to make the session id for `currentSession` and `previousSession` the same.  Looking at the code (current and legacy), this can happen and is one of the conditions for successfully opening a previous exam.  I hope this is what that unit test is trying to express.

service/src/main/java/tds/exam/web/resources/ExamApprovalResource.java
```diff
@@ -0,0 +1,41 @@
...
+import tds.exam.ExamApproval;
+import tds.exam.web.endpoints.ExamController;
+
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
+
+/**
+ * A HATEOAS representation of an {@link tds.exam.ExamApproval}
+ */
+public class ExamApprovalResource extends ResourceSupport {
```
jeffjohnson9046: Added the `ExamApprovalResource` for consistency, even though we're likely moving away from returning data to the caller when the entity -> resource/"view" mapping is one to one (as it is in this case).  Will clean up later.

client/src/main/java/tds/exam/ExamApprovalStatus.java
```diff
@@ -0,0 +1,15 @@
+package tds.exam;
+
+/**
+ * Enumerate the possible status values for exam approval.
+ * <p>
+ *     This enumeration maps to the {@code OpportunityApprovalStatus} contained within the
+ *     {@code tds.student.services.data.ApprovalInfo} class of the legacy Student application.
+ * </p>
+ */
+public enum ExamApprovalStatus {
```
millerts: What if we added a property `code` to the enum.  Then in the approval status creation you wouldn't need to declare Strings to do the comparison.  Also, later we could pass the enums around rather than Strings if we ever need to query by status

```java
public enum ExamApprovalStatus {
    WAITING("waiting"),
    APPROVED("approved");
    
    private final String code;
    
    private ExamApprovalStatus(String code) { 
        this.code = code;
    }

    public String getCode() {
        return code;
    }
}
```

service/src/main/java/tds/exam/configuration/ExamServiceProperties.java
```diff
@@ -59,7 +58,23 @@ public String getAssessmentUrl() {
     }
 
     public void setAssessmentUrl(String assessmentUrl) {
-        if (studentUrl == null) throw new IllegalArgumentException("asssessmentUrl cannot be null");
```
millerts: Oops.  Thanks for fixing my mistake

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -7,21 +7,18 @@
...
 import java.util.UUID;
 
 import tds.common.Response;
 import tds.common.ValidationError;
 import tds.common.data.legacy.LegacyComparer;
-import tds.exam.Exam;
-import tds.exam.ExamStatusCode;
-import tds.exam.OpenExamRequest;
+import tds.config.TimeLimitConfiguration;
+import tds.exam.*;
```
millerts: I prefer us to just list them out instead of the * in non test code.  It better conveys the imports being used when looking at the code.  We should get our IntelliJ settings in sync on it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -199,4 +211,74 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+        // RULE:  Session id for the approval request must match the session id of the exam.
+        if (!exam.getSessionId().equals(examApprovalRequest.getSessionId())) {
+            return Optional.of(new ValidationError(ValidationErrorCode.EXAM_APPROVAL_SESSION_ID_MISMATCH, "The session keys do not match; please consult your test administrator"));
+        }
+
+        Session session = sessionService.findSessionById(examApprovalRequest.getSessionId())
+                .orElseThrow(() -> new IllegalArgumentException("Could not find session for id " + examApprovalRequest.getSessionId()));
+
+        // RULE:  Unless the environment is set to "simulation" or "development", the exam's session must be open.
```
millerts: let's move this comment down to line 255 above the check.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -199,4 +211,74 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+        ExternalSessionConfiguration externalSessionConfig =
+                sessionService.findExternalSessionConfigurationByClientName(examApprovalRequest.getClientName())
+                        .orElseThrow(() -> new IllegalStateException("External Session Configuration could not be found for client name " + examApprovalRequest.getClientName()));
```
millerts: Just calling this out that this will throw a 500 error.  I think that is appropriate.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -199,4 +211,74 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+        boolean checkSession = (!externalSessionConfig.getEnvironment().toLowerCase().equals(SIMULATION_ENVIRONMENT)
+                        && !externalSessionConfig.getEnvironment().toLowerCase().equals(DEVELOPMENT_ENVIRONMENT));
```
millerts: I think the `SIMULATION_ENVIRONMENT` and `DEVELOPMENT_ENVIRONMENT` should be constants in ExternalSessionConfiguration.  I have to believe this will not be the last time we'll need these values in the application.

service/src/main/resources/db/migration/V1475278170__exam_add_browser_key_column.sql
```diff
@@ -0,0 +1,36 @@
...
+Since this is an early migration there should be no data in the exam table except for development purposes.  Due to
+compatibility issues between database engines the easiest way to do this right now is:
+
+1. Drop the exam table
+2. Recreate it
+
+*/
+DROP TABLE exam;
+
+CREATE TABLE IF NOT EXISTS exam(
```
millerts: Why are we dropping the table?  I dropped the table in the previous migration because the primary key and auto incrementing would have required a bunch of extra steps.  SHouldn't this simply be an alter statement?

And sadly the previous migration doesn't even need to drop anymore now that we're using MySQL for test and prod.

jeffjohnson9046: There's no technical reason.  I did this so that the column I added would appear in the right spot, rather than being tacked on at the end.  On a subsequent migration (where I add the `status_change_reason`), I punted on this idea and added the new column at the end.  Ideally, as the `exam` schema stabilizes, we'll go through and organize the tables so the columns are in a logical order (e.g. I think the `status_change_reason` column should come after `status` and I think all the `*id` columns should be near one another).

millerts: Actually, I mentioned this in the parking lot today, and @jtreuting told us there is a placement option in the alter statement in MySQL.  From the [docs](http://dev.mysql.com/doc/refman/5.7/en/alter-table.html)

`To add a column at a specific position within a table row, use FIRST or AFTER col_name. The default is to add the column last. You can also use FIRST and AFTER in CHANGE or MODIFY operations to reorder columns within a table.`

jtreuting: @millerts stealing my thunder!

`ALTER table table_name
    Add column column_name57 integer AFTER column_name56`

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -418,4 +444,504 @@ public void shouldOpenPreviousExamIfPreviousExamIsInactiveStage() {
...
+        Response<ExamApproval> result = examService.getApproval(examApprovalRequest);
+
+        assertThat(result.getErrors()).isNotPresent();
+        assertThat(result.getData()).isPresent();
+        assertThat(result.getData().get().getExamId()).isEqualTo(examId);
+        assertThat(result.getData().get().getExamApprovalStatus()).isEqualTo(ExamApprovalStatus.APPROVED);
+    }
+
+    @Test
+    public void shouldReturnExamApprovalBecauseEnvironmentIsDevelopment() {
```
jtreuting: These tests are pulling back different exams in various approval states it seems?  So should this be named something like `shouldReturnExamApprovalInWaitingWhenEnvironmentIsDevelopment`?  The reason I ask is it looks similar to the previous test but with a different status check?

client/src/main/java/tds/exam/ExamApprovalStatus.java
```diff
@@ -8,8 +8,38 @@
...
+     *
+     * @param status The status value from the {@link ExamStatusCode}
+     * @return The appropriate {@link ExamApprovalStatus} for the {@link Exam}'s status; otherwise
+     * {@code ExamApprovalStatus.WAITING}.
+     */
+    public static ExamApprovalStatus fromExamStatus(String status) {
+        if (status == null) throw new IllegalArgumentException("status cannot be null");
+
+        // Can't use Java 8 optional or streams here; must be Java 7 compliant.
+        for (ExamApprovalStatus approvalStatus : ExamApprovalStatus.values()) {
```
jtreuting: Instead of looping through this, I think you can use `ExamApprovalStatus.valueOf(status)`

https://www.mkyong.com/java/java-convert-string-to-enum-object/

PR #11, opened 2016-10-08
Feature/get initial ability

burzum619:
## Notes
This commit includes changes to the Get Initial Ability portion of the ExamService.

* Created ConfigService to retrieve ClientTestProperties for an assessment from the configs microservice.
* Added **scores** and **history** table, and added three new columns to **exam**.
* Added methods to ExamService and ExamQueryRepository to retrieve ability-related data from exam database

**NOTE** : The ExamService.getInitialAbility() method is based on the legacy code in StudentDLL._GetInitialAbility_SP method (TDS_TestDeliverySystemDataAccess).


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -109,6 +117,54 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+    @Override
+    public Optional<Float> getInitialAbility(Exam exam, ClientTestProperty property) {
+        Optional<Float> ability = Optional.empty();
+        Double slope = property.getAbilitySlope();
+        Double intercept = property.getAbilityIntercept();
+        List<Ability> testAbilities = examQueryRepository.findAbilities(exam.getId(), exam.getClientName(),
+                property.getSubjectName(), exam.getStudentId());
+
+        // Attempt to retrieve the most recent ability for the current subject and assessment
+        Optional<Ability> initialAbility = getMostRecentTestAbility(testAbilities, exam.getAssessmentId(), false);
```
jtreuting: When reviewing the code it's hard to tell what the last flag is for.  Not sure if there is a Java best practice for these situations (or if it's an issue because of IDE support for easily seeing what it is), would love to get folks thoughts on this?  @millerts 

in C# I would use Named Parameters like `getMostRecentTestAbility(testAbilities, exam.getAssessmentId(), matchesAssessment: false);`, but Java doesn't have anything like this, so the best thing I can think of is something like this getMostRecentTestAbility(testAbilities, exam.getAssessmentId(), /* matchesAssessment */ false);`which may or may not be a good idea.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -124,6 +180,36 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+     * @param matchesAssessment Specifies whether to search for matches or non-matches of the assessment key
+     * @return
+     */
+    private Optional<Ability> getMostRecentTestAbility(List<Ability> abilityList, String assessmentId, boolean matchesAssessment) {
+        Ability mostRecentAbility = null;
+
+        for (Ability ability : abilityList) {
+            if (matchesAssessment) {
+                if (!assessmentId.equals(ability.getAssessmentId())) {
+                    if (mostRecentAbility == null || mostRecentAbility.getDateScored().isBefore(ability.getDateScored())) {
```
jtreuting: Can combine these 2 if statements into one.  Same with below.

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -53,4 +55,19 @@ public void shouldNotReturnLatestExamWithNonNullDeletedDate() {
...
+                "otherclient", "ELA", 9999L);
+        assertThat(noAbilities).isEmpty();
+    }
+
+    @Test
+    public void shouldReturnSingleAbility() {
+        List<Ability> oneAbility = examQueryRepository.findAbilities(UUID.fromString("af880054-d1d2-4c24-805c-1f0dfdb45989"),
+                "clientName", "ELA", 9999L);
+        assertThat(oneAbility).hasSize(1);
+
```
jtreuting: Validate that the properties of the ability are what is expected.

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+                        "X'12380054d1d24c24805c0dfdb45a0daa', '03', 'SSID1', null, null)";
+        jdbcTemplate.update(SQL1);
+        jdbcTemplate.update(SQL2);
+        jdbcTemplate.update(SQL3);
+        jdbcTemplate.update(SQL4);
+    }
+
+    @After
+    public void cleanUp() {
+        jdbcTemplate.update("DELETE FROM history");
```
jtreuting: If we used @Transactional wouldn't it roll back the inserts for us?  Or does that not work when doing the inserts in the @Before method.

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Test
+    public void shouldRetrieveLargestAbilityValue() {
+        Optional<Float> maxAbility = historyQueryRepository.findAbilityFromHistoryForSubjectAndStudent("CLIENT_TEST", "ELA", STUDENT_ID1);
+        assertThat(maxAbility.get()).isEqualTo(MAX_ABILITY_VAL_FOR_STUDENT1);
+    }
+
+    @Test
+    public void shouldReturnNullForNullAbilities() {
+        Optional<Float> maxAbility = historyQueryRepository.findAbilityFromHistoryForSubjectAndStudent("CLIENT_TEST", "ELA", STUDENT_ID2);
+        assertThat(maxAbility.isPresent()).isFalse();
```
jtreuting: I believe we've been using `assertThat(maxAbility).isNotPresent()` in these cases.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
+        differentAssessmentAbility.setAssessmentId(assessmentId);
+        differentAssessmentAbility.setAttempts(1);
+        differentAssessmentAbility.setDateScored(Instant.now());
+
+        List<Ability> abilities = new ArrayList<>();
+        abilities.add(sameAssessmentAbility);
+        abilities.add(differentAssessmentAbility);
+
+        abilities.add(sameAssessmentAbility);
+        abilities.add(differentAssessmentAbility);
```
jtreuting: Is adding these again on purpose?

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
+        sameAssessmentAbility.setAssessmentId(assessmentId);
+        sameAssessmentAbility.setAttempts(1);
+        sameAssessmentAbility.setDateScored(Instant.now());
+
+        Ability differentAssessmentAbility = new Ability();
+        differentAssessmentAbility.setExamId(UUID.randomUUID());
+        differentAssessmentAbility.setScore(50F);
+        differentAssessmentAbility.setAssessmentId(assessmentId);
+        differentAssessmentAbility.setAttempts(1);
+        differentAssessmentAbility.setDateScored(Instant.now());
```
jtreuting: `sameAssessmentAbility` and `differentAssessmentAbility` are not added to the `abilities` array below, I'm guessing they can be deleted.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
```
jtreuting: Same.  These 2 Ability aren't used.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
+
+        List<Ability> abilities = new ArrayList<>();
+        when(repository.findAbilities(thisExamId, clientName, "ELA", studentId)).thenReturn(abilities);
+        when(historyRepository.findAbilityFromHistoryForSubjectAndStudent(clientName, "ELA", studentId))
+                .thenReturn(Optional.empty());
+        when(assessmentService.findSetOfAdminSubjectByKey(thisExam.getAssessmentId())).thenReturn(Optional.empty());
+        Optional<Float> maybeAbilityReturned = examService.getInitialAbility(thisExam, clientTestProperty);
+        assertThat(maybeAbilityReturned.isPresent()).isFalse();
```
jtreuting: `assertThat(maybeAbilityReturned).isNotPresent()`

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
+        sameAssessmentAbility.setAssessmentId(assessmentId);
+        sameAssessmentAbility.setAttempts(1);
+        sameAssessmentAbility.setDateScored(Instant.now());
+
+        Ability differentAssessmentAbility = new Ability();
+        differentAssessmentAbility.setExamId(UUID.randomUUID());
+        differentAssessmentAbility.setScore(50F);
+        differentAssessmentAbility.setAssessmentId(assessmentId);
+        differentAssessmentAbility.setAttempts(1);
+        differentAssessmentAbility.setDateScored(Instant.now());
```
jtreuting: not used

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -394,6 +402,355 @@ public void shouldOpenPreviousExamIfSessionEndTimeIsBeforeNow() {
...
```
jtreuting: not used

service/src/main/java/tds/exam/models/Ability.java
```diff
@@ -0,0 +1,74 @@
...
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Model representing an exam's ability.
+ */
+public class Ability {
+    /**
+     * The id of the exam
+     */
```
millerts: Please put the JavaDocs on the getters and setters.  Field JavaDocs won't show up if we ever create JavaDoc html files.

service/src/main/java/tds/exam/models/Ability.java
```diff
@@ -0,0 +1,74 @@
+package tds.exam.models;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Model representing an exam's ability.
+ */
+public class Ability {
```
millerts: We should make this immutable.  I think five arguments is the top limit we decided for constructors so a constructor with all arguments being passed in should work.  Also, the setters should be removed.

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Created by emunoz on 10/6/16.
+ */
+@Repository
+public class HistoryQueryRepositoryImpl implements HistoryQueryRepository {
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Autowired
+    public HistoryQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDataSource) {
+        this.jdbcTemplate = new NamedParameterJdbcTemplate(queryDataSource);
```
millerts: We should discuss configuring (in Spring @Config annotated class) a standard QueryNamedParameterJdbcTemplate and a CommandNamedParameterJdbcTemplate and have that passed in rather than passing the datasource everywhere.  All repositories simply do this new instantiation.

service/src/main/java/tds/exam/services/ConfigService.java
```diff
@@ -0,0 +1,13 @@
...
+import tds.config.ClientTestProperty;
+
+import java.util.Optional;
+
+/**
+ * A service that handles configuration interaction.
+ */
+public interface ConfigService {
+
+    Optional<ClientTestProperty> findClientTestPropertyByClientAndAssessment(final String clientName, final String assessmentId);
```
millerts: JavaDocs

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -124,6 +180,36 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+     *
+     * @param abilityList the list of {@link Ability}s to iterate through
+     * @param assessmentId  The test key
+     * @param matchesAssessment Specifies whether to search for matches or non-matches of the assessment key
+     * @return
+     */
+    private Optional<Ability> getMostRecentTestAbility(List<Ability> abilityList, String assessmentId, boolean matchesAssessment) {
+        Ability mostRecentAbility = null;
+
+        for (Ability ability : abilityList) {
```
millerts: Can we change the conditional to this:

```java
isAbilityBeforeRecentDateScore = mostRecentAbility == null || mostRecentAbility.getDateScored().isBefore(ability.getDateScored());
if (matchesAssessment
    && !assessmentId.equals(ability.getAssessmentId())
    && isAbilityBeforeRecentDateScore) {
    	mostRecentAbility = ability;
    }
} else if (assessmentId.equals(ability.getAssessmentId()) && isAbilityBeforeRecentDateScore) {
    mostRecentAbility = ability;
}
```

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -124,6 +180,36 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
 
+    /**
+     * Gets the most recent {@link Ability} based on the dateScored value
+     *
+     * @param abilityList the list of {@link Ability}s to iterate through
+     * @param assessmentId  The test key
+     * @param matchesAssessment Specifies whether to search for matches or non-matches of the assessment key
+     * @return
+     */
+    private Optional<Ability> getMostRecentTestAbility(List<Ability> abilityList, String assessmentId, boolean matchesAssessment) {
```
millerts: I assume we're porting over this logic, but looking at the query we don't order the abilities.  That doesn't quite jive with the for loop we have below since we iterate over and basically take the last one we find that matches.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -124,6 +180,36 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         return null;
     }
 
+    /**
+     * Gets the most recent {@link Ability} based on the dateScored value
+     *
+     * @param abilityList the list of {@link Ability}s to iterate through
+     * @param assessmentId  The test key
+     * @param matchesAssessment Specifies whether to search for matches or non-matches of the assessment key
+     * @return
```
millerts: Thanks for doing the JavaDocs, but I personally feel they're only necessary on public methods.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Another question, when matchesAssessment is set to true the conditional checks for assessmentId != assessmentId?  I'm having trouble following the reasoning behind this logic.  Maybe the naming can be improved?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -109,6 +117,54 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         return new Response<>(exam);
     }
 
+    /**
+     * @inheritDoc
+     */
+    @Override
+    public Optional<Float> getInitialAbility(Exam exam, ClientTestProperty property) {
+        Optional<Float> ability = Optional.empty();
+        Double slope = property.getAbilitySlope();
```
millerts: Surprising these are Doubles when we simply convert them to float.

service/src/main/java/tds/exam/configuration/ExamServiceProperties.java
```diff
@@ -9,6 +9,27 @@
...
+
+    /**
+     * Get the URL for the config microservice.  Will always end with "/" unless blank
+     *
+     * @return config microservice URL
+     */
+    public String getConfigUrl() {
+        return configUrl.endsWith("/")
+                ? configUrl
+                : configUrl + "/";
```
millerts: I ported this pattern over from the POC.  I agree it should be in the setter, and I actually would advocate that we should strip the "/" character if the String ends with one.  I think now that I've seen how it is used in practice it would read clearer.  For example, right now we do ("%s%s/%s", url, studentId, client).  Removing the "/" would make it ("%s/%s/%s", url, studentId,client) which I think reads a lot better.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Sure would be nice if Java had that naming style.  After taking another look at the code I'd rather separate the `getMostRecentTestAbility` into two methods.  One can be `getMostRecentTestAbilityMatchingAssessment()` and the other could be  `getMostRecentTestAbilityNotMatchingAssessment()`.  The logic within the method doesn't need to live together as far as I can tell.

This [article](http://martinfowler.com/bliki/FlagArgument.html) by Martin Fowler aligns with my thinking on the subject.  I prefer to have two methods than pass in a flag.  If the branching cannot be removed I usually provide two public methods that simply hide the fact the underlying logic is using a flag to branch.

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration test for testing the {@link HistoryQueryRepositoryImpl}
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
+@SqlConfig(dataSource = "queryDataSource")
+public class HistoryQueryRepositoryImplIntegrationTest {
```
millerts: Rather than write this repository integration tests I suggest writing one against the service only.

service/src/main/java/tds/exam/configuration/ExamServiceProperties.java
jtreuting: I agree with removing the last / which makes the string format calls cleaner

service/src/main/resources/db/migration/V1475261281__exam_create_table_scores_and_history.sql
```diff
@@ -0,0 +1,42 @@
...
+
+  Desc: Creates the scores database table
+
+***********************************************************************************************************************/
+USE exam;
+
+CREATE TABLE IF NOT EXISTS scores(
+  fk_scores_examid_exam VARBINARY(16) NOT NULL,
+  measure_label VARCHAR(100) NOT NULL,
+  value VARCHAR(100) DEFAULT NULL,
```
burzum619: Yes, it is. the legacy table "testopportunityscores" has the "value" column as a VARCHAR(100), but in StudentDLL._GetInitialAbility_SP(), this value is converted into a Float when it is inserted into the temporary table that is created. I think we're probably safe to change this to a FLOAT in the schema, but that would be a deviation from the legacy schema. I see no reason why it would be a VARCHAR realistically.

millerts: I think it should be float in our schema.  Also, can we call this exam_scores if it is for the overall score of the exam?

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTest.java
burzum619: I considered that. However, the service method that calls this is the ExamServiceImpl.getInitialAbilty() method, which also requires an external microservice (ConfigService). Unit tests were created for getInitialAbility to cover each branch of logic in the service method, while integration tests were created for this repository.

client/src/main/java/tds/exam/Exam.java
```diff
@@ -142,6 +158,8 @@ public String getClientName() {
         return clientName;
     }
 
+    public String getSubject() { return subject; }
```
burzum619: Done. Thanks to IntelliJ for being "helpful"

service/src/main/java/tds/exam/configuration/ExamServiceProperties.java
burzum619: Updated ExamServiceProperties (and associated tests) to trim the trailing "/" on urls.

service/src/main/java/tds/exam/models/Ability.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * The date when the exam was scored
+     */
+    private Instant dateScored;
+
+    /**
+     * The ability score value
+     */
+    private Float score;
```
burzum619: Done

service/src/main/resources/db/migration/V1475261281__exam_create_table_scores_and_history.sql
burzum619: Table name updated to exam_scores and "value" is now a DOUBLE.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -110,6 +113,51 @@ public ExamQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDat
...
+
+        final String SQL =
+            "SELECT\n" +
+                "exam.exam_id,\n" +
+                "exam.assessment_id,\n" +
+                "exam.attempts,\n" +
+                "exam.date_scored,\n" +
+                "scores.value AS score\n" +
+            "FROM\n" +
+                "exam, scores\n" +
```
burzum619: Done - this query was ported directly from legacy which had the JOIN in the WHERE.

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.sql.DataSource;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Class for retrieving data from the exam history table.
+ *
+ * Created by emunoz on 10/6/16.
```
burzum619: Done

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("clientName", clientName);
+        parameters.put("subject", subject);
+        parameters.put("studentId", studentId);
+
+        final String SQL =
+                "SELECT\n" +
+                        "MAX(initial_ability)\n" +
+                        "FROM\n" +
+                        "exam.history\n" +
```
burzum619: Done

service/src/main/java/tds/exam/services/ConfigService.java
burzum619: Done. Bad habit of mine!

service/src/main/java/tds/exam/services/impl/ConfigServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    public ConfigServiceImpl(RestTemplate restTemplate, ExamServiceProperties examServiceProperties) {
+        this.restTemplate = restTemplate;
+        this.examServiceProperties = examServiceProperties;
+    }
+
+    @Override
+    public Optional<ClientTestProperty> findClientTestPropertyByClientAndAssessment(final String clientName, final String assessmentId) {
+        UriComponentsBuilder builder =
+            UriComponentsBuilder
+                .fromHttpUrl(String.format("%s/%s/%s", examServiceProperties.getConfigUrl(), clientName, assessmentId));
```
burzum619: Took the latter approach - no changes necessary here since the setter now trims off the trailing "/" from the URI

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Broke the private method into two separate methods as Todd suggested.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Refactored this method into two, more simple methods. Date comparison no longer necessary as query that retrieves the list of abilities is now being ordered by date_scored descending.

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
burzum619: Done

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTest.java
burzum619: Done

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
burzum619: Removed

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
burzum619: Removed

service/src/main/java/tds/exam/models/Ability.java
burzum619: Removed setters, created Default constructor (for jdbc template mapping) and the 5 arg constructor.

service/src/main/java/tds/exam/models/Ability.java
burzum619: Done

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
burzum619: Completely agree. We can discuss this tomorrow morning during parking lot, but I'll leave as is for now.

service/src/main/java/tds/exam/services/ConfigService.java
burzum619: Done

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: matchesAssessment was a typo on my part (in legacy app, this flag was simply named "inverse" which was also confusing). Broke this method out into two separate private methods, so no longer an issue.

burzum619:
Comments/suggestions have been addressed with the most recent commit/push. Please feel free to re-review.

service/src/main/java/tds/exam/repositories/HistoryQueryRepository.java
```diff
@@ -0,0 +1,18 @@
...
+ * Repository for querying the history table.
+ */
+public interface HistoryQueryRepository {
+    /**
+     * Retrieves the maximum ability value from the history table for the specified subject, client, and student.
+     *
+     * @param clientName    Name of the SBAC client
+     * @param subject       Subject name (typically ELA or MATH)
+     * @param studentId     Id of the student to obtain the ability for
+     * @return
```
millerts: Can you fill out this return?  Mostly because I can't figure out what the double represents by the method name since we have an `Ability` object and this isn't returning it even though the method is `findAbility`.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -110,6 +113,51 @@ public ExamQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDat
...
+                "exam.exam_id <> :examId AND\n" +
+                "exam.exam_id = scores.fk_scores_examid_exam AND\n" +
+                "scores.use_for_ability = 1 AND\n" +
+                "scores.value IS NOT NULL";
+
+
+        return jdbcTemplate.query(SQL, parameters, new AbilityRowMapper());
+    }
+
+    private class AbilityRowMapper implements RowMapper<Ability> {
```
millerts: It would be worth trying.  Not sure how it deals with immutability

jtreuting: Yeah, not well :)  But in my defense, it wasn't immutable when I suggested it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -124,6 +180,45 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+    /**
+     * Gets the most recent {@link Ability} based on the dateScored value for the same assessment.
+     *
+     * @param abilityList the list of {@link Ability}s to iterate through
+     * @param assessmentId  The test key
+     * @return
+     */
+    private Optional<Ability> getMostRecentTestAbilityForSameAssessment(List<Ability> abilityList, String assessmentId) {
+        for (Ability ability : abilityList) {
+            if (assessmentId.equals(ability.getAssessmentId())) {
```
jtreuting: Just noticing that this might be a nice place for Java 8 streams.  I think it would go something like this:

`abilityList.stream().filter(x -> assessmentId.equals(x.getAssessmentId())).findFirst().get();`

PR #13, opened 2016-10-17
Enabling flyway

burzum619:
Enabling flyway migrations for the exam service.

### DIFF
--- a/service/src/main/docker/docker-startup.sh
+++ b/service/src/main/docker/docker-startup.sh
@@ -19,6 +19,10 @@ java \
     -Dexam-service.student-url=http://student/ \
     -Dexam-service.assessment-url=http://assessment/ \
     -Dexam-service.config-url=http://config/ \
+    -Dflyway.enabled=true \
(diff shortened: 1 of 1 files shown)


service/src/main/docker/docker-startup.sh
```diff
@@ -19,6 +19,10 @@ java \
     -Dexam-service.student-url=http://student/ \
     -Dexam-service.assessment-url=http://assessment/ \
     -Dexam-service.config-url=http://config/ \
+    -Dflyway.enabled=true \
```
millerts: Do we want to have a separate shell script called docker-startup-development.sh that will contain these values?

burzum619: Or I can change this to ${FLYWAY_ENABLED} and pass this in from TC.

millerts: Sounds great. 👍

millerts:
Approve after that one small tweak.

PR #15, opened 2016-10-20
Feature/verify access

jeffjohnson9046:
This pull request addresses TDS-349: implement a refactored version of `ValidateTesteeAccess_SP`.  `ValidateTesteeAccess_SP` conducts similar steps to the legacy `checkApproval` method, so the rules that were previously leveraged in `checkApproval` have been moved to its own method.

* Added `verifyAccess` method, which is called by `getApproval()` (previously called `verifyExamApprovalRules`).
  * The `verifyAccess` method does not currently have an HTTP endpoint; the assumption here is that this method will be called internally by other methods within the exam service
  * If other micro service methods (e.g. session or student microservices) need to call `verifyAccess`, an HTTP endpoint can easily be added
* Renamed classes that were specific to `getApproval` from `Approval` to `Access`

**NOTE:** This pull request is dependent upon the [`feature/verify-access`](https://github.com/SmarterApp/TDS_SessionService/pull/13) pull request against **TDS_Session**.

### DIFF
--- a/service/src/main/java/tds/exam/services/ExamService.java
+++ b/service/src/main/java/tds/exam/services/ExamService.java
@@ -53,8 +53,8 @@ public interface ExamService {
      * <p>
      *     The rules are:
      *     <ul>
-     *         <li>The browser key of the approval request must match the browser key of the {@link Exam}.</li>
-     *         <li>The session id of the approval request must match the session id of the {@link Exam}.</li>
+     *         <li>The browser key of the access request must match the browser key of the {@link Exam}.</li>
+     *         <li>The session id of the access request must match the session id of the {@link Exam}.</li>
      *         <li>The {@link Session} must be open (unless the environment is set to "simulation" or "development")</li>
      *         <li>The TA Check-In time window cannot be passed</li>
      *     </ul>
--- a/service/src/main/java/tds/exam/services/ExamService.java
+++ b/service/src/main/java/tds/exam/services/ExamService.java
@@ -8,7 +8,7 @@ import tds.common.ValidationError;
 import tds.config.ClientTestProperty;
 import tds.exam.Exam;
 import tds.exam.ExamApproval;
-import tds.exam.AccessRequest;
+import tds.exam.ApprovalRequest;
 import tds.exam.OpenExamRequest;
 import tds.session.Session;
 
@@ -34,10 +34,10 @@ public interface ExamService {
      *     the Student's request to start his/her exam.
      * </p>
      *
-     * @param accessRequest The {@link AccessRequest} representing the request to open the specified exam.
+     * @param approvalRequest The {@link ApprovalRequest} representing the request to open the specified exam.
      * @return {@link ExamApproval} describing whether the exam is approved to be opened.
      */
-    Response<ExamApproval> getApproval(AccessRequest accessRequest);
+    Response<ExamApproval> getApproval(ApprovalRequest approvalRequest);
 
     /**
      * Retrieves the initial ability value for an {@link Exam}.
@@ -49,23 +49,23 @@ public interface ExamService {
     Optional<Double> getInitialAbility(Exam exam, ClientTestProperty clientTestProperty);
 
     /**
-     * Verify all the rules for granting access to an {@link Exam} are satisfied.
+     * Verify all the rules for granting approval to an {@link Exam} are satisfied.
      * <p>
      *     The rules are:
      *     <ul>
-     *         <li>The browser key of the access request must match the browser key of the {@link Exam}.</li>
-     *         <li>The session id of the access request must match the session id of the {@link Exam}.</li>
+     *         <li>The browser key of the approval request must match the browser key of the {@link Exam}.</li>
+     *         <li>The session id of the approval request must match the session id of the {@link Exam}.</li>
      *         <li>The {@link Session} must be open (unless the environment is set to "simulation" or "development")</li>
      *         <li>The TA Check-In time window cannot be passed</li>
      *     </ul>
      *     <strong>NOTE:</strong>  If the {@link Session} has no Proctor (because the {@link Session} is a guest session
-     *     or is otherwise proctor-less), access is granted as long as the {@link Session} is open.
+     *     or is otherwise proctor-less), approval is granted as long as the {@link Session} is open.
      * </p>
      *
-     * @param accessRequest The {@link AccessRequest} being evaluated
-     * @param exam The {@link Exam} for which access is being requested
-     * @return An empty optional if the access rules are satisfied; otherwise an optional containing a
+     * @param approvalRequest The {@link ApprovalRequest} being evaluated
+     * @param exam The {@link Exam} for which approval is being requested
+     * @return An empty optional if the approval rules are satisfied; otherwise an optional containing a
      * {@link ValidationError} describing the rule that was not satisfied
      */
-    Optional<ValidationError> verifyAccess(AccessRequest accessRequest, Exam exam);
+    Optional<ValidationError> verifyAccess(ApprovalRequest approvalRequest, Exam exam);
 }
--- a/service/src/main/java/tds/exam/services/ExamService.java
+++ b/service/src/main/java/tds/exam/services/ExamService.java
@@ -4,11 +4,13 @@ import java.util.Optional;
 import java.util.UUID;
 
 import tds.common.Response;
+import tds.common.ValidationError;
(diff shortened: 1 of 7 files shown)


service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -32,10 +34,10 @@
      *     the Student's request to start his/her exam.
      * </p>
      *
-     * @param examApprovalRequest The {@link ExamApprovalRequest} representing the request to open the specified exam.
+     * @param accessRequest The {@link AccessRequest} representing the request to open the specified exam.
      * @return {@link ExamApproval} describing whether the exam is approved to be opened.
      */
-    Response<ExamApproval> getApproval(ExamApprovalRequest examApprovalRequest);
+    Response<ExamApproval> getApproval(AccessRequest accessRequest);
```
jtreuting: Should the name of this method be changed?  I might just be stuck on the object being passed in being an AccessRequest.  Isn't this getApproval, really verifying access?

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -58,8 +54,8 @@ public ExamController(ExamService examService) {
 
     @RequestMapping(value = "/{id}/get-approval", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity<ExamApprovalResource> getApproval(@PathVariable final UUID examId, @RequestParam final UUID sessionId, @RequestParam final UUID browserId, final String clientName) {
```
jtreuting: This is really exposing the `verifyAccess` method to the other micro-services I would argue.  So it seems like maybe we should rename it from `getApproval`.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -857,7 +857,7 @@ public void shouldReturnExamApprovalWithApprovedStatusBecauseEnvironmentIsSimula
     }
 
     @Test
-    public void shouldReturnExamApprovalWithDeniedStatusBecauseSessionIsProctorless() {
```
jtreuting: Was this just named incorrectly originally?  Do we need a test for proctorless?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -185,6 +185,60 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         return ability;
     }
 
+    @Override
+    public Optional<ValidationError> verifyAccess(AccessRequest accessRequest, Exam exam) {
+        final String SIMULATION_ENVIRONMENT = "simulation";
```
millerts: I think these constant values should live on the `ExternalSessionConfiguration` object since that is where the values originate.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -185,6 +185,60 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+            .orElseThrow(() -> new IllegalArgumentException("Could not find session for id " + accessRequest.getSessionId()));
+
+        // RULE:  the exam's session must be open.
+        if (!session.isOpen()) {
+            return Optional.of(new ValidationError(ValidationErrorCode.EXAM_APPROVAL_SESSION_CLOSED, "The session is not available for testing, please check with your test administrator."));
+        }
+
+        // RULE:  If the session has no proctor, there is nothing to approve.  This is either a guest session or an
+        // otherwise proctor-less session.
+        if (session.getProctorId() == null) {
```
millerts: What if we add an isProctorless() method to the session object?

service/src/main/java/tds/exam/web/endpoints/ExamController.java
millerts: And rename `AccessRequest` to `ApprovalRequest`?

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
jeffjohnson9046: This actually _is_ testing the proctorless session, but the name was confusing.  Turns out I didn't do any better this time.  I've changed the mock exam's status to Approved because the name of the test implied (to me, anyway) that an exam is denied because the session is proctorless, which really isn't the case.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
jtreuting: I feel like the approval and the fetching of accommodations are distinct things, but they aren't in the legacy system.  This is a hard one.  We want to limit the footprint of cutting into the legacy Student app, so I'm tempted to say we leave them together and then the existing name works.

Thoughts?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jtreuting: Along those lines, is it worth having a `isSimulation()` and `isDevelopment()` methods on it, or 'isSimualtionEnvironment()` might be better.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
jeffjohnson9046: I agree - getting approval and getting accommodations are entirely separate things.  I think for now, let's emulate the legacy system - this `getApproval` method will:
* Make sure all the approval rules are satisfied
* Get the appropriate accommodations (when TDS-368 and TDS-371 are completed)

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: Done:  Refer to [this pull request](https://github.com/SmarterApp/TDS_SessionService/pull/13) against the *TDS_SessionService* for details.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: Done.  Added `isProctorless()` to the `Session` in [this pull request](https://github.com/SmarterApp/TDS_SessionService/pull/13) against the *TDS_SessionService*.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -126,15 +126,15 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
     }
 
     @Override
-    public Response<ExamApproval> getApproval(ExamApprovalRequest examApprovalRequest) {
-        Exam exam = examQueryRepository.getExamById(examApprovalRequest.getExamId())
-                .orElseThrow(() -> new IllegalArgumentException("Exam could not be found for id " + examApprovalRequest.getExamId()));
+    public Response<ExamApproval> getApproval(ApprovalRequest approvalRequest) {
+        Exam exam = examQueryRepository.getExamById(approvalRequest.getExamId())
+                .orElseThrow(() -> new IllegalArgumentException(String.format("Exam could not be found for id %s", approvalRequest.getExamId())));
```
millerts: Should this be an `IllegalArgumentException` or a `NotFoundException`?

millerts: Same question throughout.  If we can't find one of the pieces of information waht is the appropriate http response code.

jeffjohnson9046: Right now, the `IllegalArgumentException` and `IllegalStateException` (in the case of getting `ExternalSessionConfiguration` are consistent with what's thrown in `openExam()` under similar conditions.  I think one of two things should happen (in no particular order):

1.  The exception thrown should be changed to a `NotFoundException`, which is consistent with what we're doing elsewhere (e.g. when a `find` method returns `Optional.empty()` in a controller.
2.  The exception thrown should remain the same, but the message should change.  Instead of saying "could not find exam foo", the message should say "attempted to get approval for exam foo, which does not exist."  Or something like that.

Right now, my instincts lean a bit towards #2: the caller's passing in bad arguments, trying to get approval for something that doesn't exist (for whatever reason).  I feel like that is a bad request and should return a 400.  However, I could also see a case for returning a 404 for pretty much the same reason (trying to find something that doesn't exist).

PR #19, opened 2016-10-31
Feature/find accommodation

jeffjohnson9046:
* add `findAccommodations()` method and supporting classes  
* Caller can specify an array of accommodation types to return
* Approved and denied accommodations are returned (accommodations can be denied by a Proctor), so the caller can choose how to handle denial of accommodations
  * the `Accommodation` class has an `isApproved()` method to determine if it is approved


service/src/test/java/tds/exam/builder/AccommodationBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+ * Build an {@link tds.exam.Accommodation} populated with test data
+ */
+public class AccommodationBuilder {
+    public static class SampleData {
+        public static final UUID DEFAULT_EXAM_ID = UUID.fromString("6b824c7d-0215-4229-ba95-99f1dae5ef04");
+        public static final String DEFAULT_SEGMENT_ID = "segment-1";
+        public static final String DEFAULT_ACCOMMODATION_TYPE = "language";
+        public static final String DEFAULT_ACCOMMODATION_CODE = "ENU";
+
+        public static final String SECOND_SEGMENT_ID = "segment-2";
```
millerts: This data looks to be specific for the tests and not for the builder itself.  So I think this should be in the tests that need it not here.

service/src/test/java/tds/exam/repositories/impl/AccommodationQueryRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,218 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
+@Transactional
+public class AccommodationQueryRepositoryIntegrationTests {
+    @Autowired
+    @Qualifier("commandDataSource")
+    private DataSource commandDataSource;
```
millerts: This will also need to be tweaked once you inject the NamedParameterJdbcTemplate rather than the datasource.

service/src/test/java/tds/exam/repositories/impl/AccommodationQueryRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,218 @@
...
+    @Test
+    public void shouldGetOneAccommodationForExamAndSegmentAndSingleAccommodationType() {
+        List<Accommodation> result = accommodationQueryRepository.findAccommodations(
+            AccommodationBuilder.SampleData.DEFAULT_EXAM_ID,
+            AccommodationBuilder.SampleData.DEFAULT_SEGMENT_ID,
+            new String[] { AccommodationBuilder.SampleData.DEFAULT_ACCOMMODATION_TYPE});
+
+        assertThat(result).isNotNull();
+        assertThat(result).hasSize(1);
+        assertThat(result.get(0).getId()).isGreaterThan(0);
```
millerts: I know it isn't a big deal but can we have something like `Accommodation accommodation = result.get(0);` then use that in the assertions.  It is easier for me to read and in addition we aren't doing the look up in the list each time.  Not a huge deal in tests but in general I wouldn't do `result.get(i)` multiple times in non test code so I wouldn't do it in tests..

service/src/test/resources/application.properties
```diff
@@ -1,17 +1,15 @@
 #flyway.enabled=
-flyway.url=
-flyway.user=
-flyway.password=
+flyway.url=jdbc:mysql://localhost:3306/exam
```
millerts: Don't think you meant to commit this file

service/src/test/java/tds/exam/web/endpoints/ExamControllerTest.java
```diff
@@ -225,4 +232,133 @@ public void shouldGetAValidationErrorWhenBrowserIdsDoNotMatch() {
...
+        ResponseEntity<List<Accommodation>> response = controller.getAccommodations(AccommodationBuilder.SampleData.DEFAULT_EXAM_ID,
+            AccommodationBuilder.SampleData.DEFAULT_SEGMENT_ID,
+            new String[] {
+                AccommodationBuilder.SampleData.DEFAULT_ACCOMMODATION_TYPE,
+                AccommodationBuilder.SampleData.ACCOMMODATION_TYPE_CLOSED_CAPTIONING });
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+        assertThat(response.getBody()).hasSize(2);
+
+        Accommodation firstResult = response.getBody().stream()
```
millerts: I like Java 8 streams but is it really necessary here?  You build the list so you know that the first one is for the default and the second is for closed caption.

service/src/test/java/tds/exam/services/impl/AccommodationServiceImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+            AccommodationBuilder.SampleData.DEFAULT_SEGMENT_ID,
+            new String[] {
+                AccommodationBuilder.SampleData.DEFAULT_ACCOMMODATION_TYPE,
+                AccommodationBuilder.SampleData.ACCOMMODATION_TYPE_CLOSED_CAPTIONING });
+
+        assertThat(results).isNotNull();
+        assertThat(results).hasSize(2);
+
+        Accommodation firstResult = results.stream()
+            .filter(x -> x.getType().equals(AccommodationBuilder.SampleData.DEFAULT_ACCOMMODATION_TYPE))
```
millerts: Same comment concerning the use of streams.  You build the list so you know the order of the accommodations so a stream.filter seems unnecessary.

service/src/test/java/tds/exam/repositories/impl/AccommodationQueryRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Test
+    public void shouldGetAnEmptyListForAnExamIdThatDoesNotExist() {
+        List<Accommodation> result = accommodationQueryRepository.findAccommodations(
+            UUID.randomUUID(),
+            AccommodationBuilder.SampleData.DEFAULT_SEGMENT_ID,
+            new String[] { AccommodationBuilder.SampleData.DEFAULT_ACCOMMODATION_TYPE});
+
+        assertThat(result).isNotNull();
```
millerts: Not a huge deal but the not null check isn't really necessary.  The assertion checking size will complain and fail the test if null telling you it is null.

service/src/test/java/tds/exam/repositories/impl/HistoryQueryRepositoryImplIntegrationTests.java
```diff
@@ -8,11 +8,11 @@
 import org.springframework.boot.test.context.SpringBootTest;
```
millerts: Not a huge deal but unsure why this class was modified as part of this work.  All that was done was line breaks and reordering of imports.

client/src/main/java/tds/exam/Accommodation.java
```diff
@@ -0,0 +1,168 @@
...
+ */
+public class Accommodation {
+    private long id;
+    private UUID examId;
+    private String segmentId;
+    private String type;
+    private String code;
+    private String description;
+    private Instant deniedAt;
+    private Instant createdAt;
```
jtreuting: I am used to seeing `deniedOn` and `createdOn` for dates.  Do we have a consensus across the new code?

service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
```diff
@@ -4,6 +4,7 @@
 import java.util.Optional;
 import java.util.UUID;
 
+import tds.exam.Accommodation;
```
jtreuting: Doesn't look like this is used

service/src/main/java/tds/exam/repositories/impl/AccommodationQueryRepositoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public List<Accommodation> findAccommodations(UUID examId, String segmentId, String[] accommodationTypes) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examId))
+            .addValue("segmentId", segmentId)
+            .addValue("accommodationTypes", Arrays.asList(accommodationTypes));
+
+        final String SQL =
+            "SELECT \n" +
```
jtreuting: Are you adding the newlines for debugging purposes?  For some reason those jumped out at me as something unnecessary, but it's not a big deal of course.

service/src/main/java/tds/exam/repositories/impl/AccommodationQueryRepositoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Autowired
+    public AccommodationQueryRepositoryImpl(@Qualifier("queryDataSource") DataSource queryDataSource) {
+        this.jdbcTemplate = new NamedParameterJdbcTemplate(queryDataSource);
+    }
+
+    @Override
+    public List<Accommodation> findAccommodations(UUID examId, String segmentId, String[] accommodationTypes) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examId))
```
jtreuting: Should `null` or and empty array for `accommodationTypes` represent getting all accommodations regardless of type?

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -7,14 +7,13 @@
...
+import java.util.List;
 import java.util.UUID;
 
 import tds.common.Response;
 import tds.common.web.exceptions.NotFoundException;
-import tds.exam.Exam;
-import tds.exam.ExamApproval;
-import tds.exam.ApprovalRequest;
-import tds.exam.OpenExamRequest;
+import tds.exam.*;
```
jtreuting: Let's use IntelliJ to only include the packages that are used.

service/src/main/java/tds/exam/repositories/impl/AccommodationQueryRepositoryImpl.java
jeffjohnson9046: I do this so the query prints out "pretty" in logging/console output.  It's been helpful once or twice.

client/src/main/java/tds/exam/Accommodation.java
millerts: I don't know if we've ever discussed it.  I created the first table that used this pattern and I'm use to the `At` instead of `On`.  I don't have a strong opinion outside of it being consistent, which I believe it is for new tables.  We're stuck with the current lack of naming convention in the existing tables.

service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
millerts: There is an `Optimize Imports` checkbox on the IntelliJ commit dialog.  I always leave it checked and will take care of unused imports and if you have your system set up remove the `*`

service/src/main/java/tds/exam/repositories/impl/AccommodationQueryRepositoryImpl.java
millerts: I personally would prefer null or empty array to return exactly what the SQL would return which I believe would be nothing.

If we have a use case for all accommodations I'd rather a separate method on the interface that doesn't take in accommodations type.  I personally don't like nulls to be used for logic behavior if it can be avoided.

jeffjohnson9046: I thought about adding a method that would get all accommodations for an exam, but I'm not sure there's a use case for it.  Thus I erred on the side of not creating that method on the assumption that if we need it later we can code it up.

client/src/main/java/tds/exam/Accommodation.java
jeffjohnson9046: I'm with @millerts on this one; I don't have a strong opinion about naming them `at` vs `on`; we've been using `created_at` in most of our conversations, so that's what I went with. As long as we're consistent, I don't really care which we choose.

jtreuting: Yep consistent is the important part.  If we've been using at then let's stick with it.

service/src/main/java/tds/exam/repositories/impl/AccommodationQueryRepositoryImpl.java
jtreuting: sounds good

PR #20, opened 2016-11-01
Feature/initialize segments - ExamSegments

burzum619:
This feature adds the exam_segment and exam_segment_event tables to the exam schema, and adds repositories for reading and writing to these tables. These tables correspond to the legacy table "testopportunitysegment" in the session schema.

- I made my best effort to identify the definitions/usages of each property in ExamSegments, but I'm sure more will be learned throughout development of InitializeSegments.
- Note: I considered removing the pool count related fields in the exam_segment table and simply using the size of the item id lists (itemPool.size()), but after analyzing our development database, it seems these values are not always necessarily equal. Therefor, I do not feel comfortable removing this column/property.
- All "updatable" fields are contained within the exam_segment_event table, as discussed in previous meetings.

ExamSegmentRepository.save - StudentDll line 4751 - "INSERT INTO testopportunitysegment..."
ExamSegmentRepository.find(by exam id) - StudentDLL 4536, 10223
ExamSegmentRepository.find(by examid and segmentpos) - StudentDLL 4848, 8644, ItemSelectionDLL 278


millerts:
Can you link to the legacy code that needs these calls?  It is hard to have context for these code additions without seeing how they'll be used.

PR #24, opened 2016-11-16
TDS-338: Update to use ExceptionAdvice from TDS_Common

jeffjohnson9046:
Updated to use the `ExceptionAdvice` and `RestTemplateConfiguration` from **TDS_Common** instead of its own internal implementations.


service/src/test/java/tds/exam/web/endpoints/ExamControllerIntegrationTests.java
```diff
@@ -22,6 +25,7 @@
 
 @RunWith(SpringRunner.class)
 @WebMvcTest(ExamController.class)
+@Import({ExceptionAdvice.class, RestTemplateConfiguration.class})
```
millerts: Do you need the RestTemplateConfiguration?   I believe you can remove it and everything will work.

PR #25, opened 2016-11-16
Use Java 7 for clients and modify code to handle other support servic…

millerts:
Two major changes:
* build client with Java 7
* modify existing code to use the new joda time support client service objects

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -16,12 +16,22 @@
         <version>0.0.1-SNAPSHOT</version>
     </parent>
 
+    <properties>
+        <java.version>1.7</java.version>
+        <assertj.version>2.5.0</assertj.version>
+        <hibernate.version>5.2.4.Final</hibernate.version>
+    </properties>
+
     <dependencies>
         <!-- Validation -->
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
             <version>${hibernate.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
         </dependency>
 
         <!-- Testing -->
@@ -38,4 +48,37 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>${maven.compiler.plugin.version}</version>
+                <configuration>
+                    <source>${java.version}</source>
+                    <target>${java.version}</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-toolchains-plugin</artifactId>
+                <version>1.1</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>toolchain</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <toolchains>
+                        <jdk>
+                            <version>1.7</version>
+                        </jdk>
+                    </toolchains>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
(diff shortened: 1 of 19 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -226,7 +226,8 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
             timeLimitConfigurationService.findTimeLimitConfiguration(approvalRequest.getClientName(), exam.getAssessmentId())
                 .orElseThrow(() -> new IllegalArgumentException(String.format("Could not find time limit configuration for client name %s and assessment id %s", approvalRequest.getClientName(), exam.getAssessmentId())));
 
-        if (Instant.now().isAfter(session.getDateVisited().plus(timeLimitConfig.getTaCheckinTimeMinutes(), ChronoUnit.MINUTES))) {
+        Instant sessionDateVisited = Instant.ofEpochMilli(session.getDateVisited().getMillis());
+        if (Instant.now().isAfter(sessionDateVisited.plus(timeLimitConfig.getTaCheckinTimeMinutes(), ChronoUnit.MINUTES))) {
```
jtreuting: Just triple checking, but Instant.now() is UTC right?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -319,13 +320,13 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
          */
         boolean daysSinceLastChange = false;
         if (previousExam.getDateChanged() != null) {
-            daysSinceLastChange = DAYS.between(previousExam.getDateChanged(), Instant.now()) >= 1;
+            daysSinceLastChange = DAYS.between(Instant.ofEpochMilli(previousExam.getDateChanged().getMillis()), Instant.now()) >= 1;
```
jtreuting: Seems like we will need to do this conversion a lot (meaning Instant.ofEpochMilli(someDate.getMillis()) because of the java 7 in the client.  Might want a helper method in common to simplify it a bit.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -319,13 +320,13 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
         }
 
         if (daysSinceLastChange ||
             LegacyComparer.isEqual(previousSession.getId(), currentSession.getId()) ||
             LegacyComparer.isEqual("closed", previousSession.getStatus()) ||
-            LegacyComparer.greaterThan(Instant.now(), previousSession.getDateEnd())) {
+            (previousSession.getDateEnd() != null && LegacyComparer.greaterThan(Instant.now(), Instant.ofEpochMilli(previousSession.getDateEnd().getMillis())))) {
```
jtreuting: since getDateEnd is not null, we should be able to get rid of the LegacyComparer and do a more normal greater than check.

client/pom.xml
```diff
@@ -16,13 +16,23 @@
         <version>0.0.1-SNAPSHOT</version>
     </parent>
 
+    <properties>
+        <java.version>1.7</java.version>
+        <assertj.version>2.5.0</assertj.version>
+        <hibernate.version>5.2.4.Final</hibernate.version>
```
jtreuting: what are we using hibernate for?

millerts: I'm playing around with the validation annotations.  They are in OpenExamRequest.  We are not leveraging the hibernate ORM but just the validation annotations.  They reject posts if the validations fail.  For example, assessmentKey cannot be null.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Actually, I'm going to add the common conversion method which will handle null by just returning null.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Done and added to common

PR #26, opened 2016-11-21
Feature/exam events

millerts:
This separates the event properties that change into a separate table allowing us to keep immutability.  The columns in the tables have also been renamed to follow our new convention.  FK constraints are also enforced.

### DIFF
--- a/service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
+++ b/service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
@@ -3,7 +3,6 @@ package tds.exam.repositories.impl;
 import org.joda.time.Instant;
 import org.joda.time.Minutes;
 import org.junit.Before;
-import org.junit.Ignore;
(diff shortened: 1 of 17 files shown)


service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -105,46 +108,20 @@ public void shouldReturnEmptyListOfAbilities() {
 
     @Test
     public void shouldReturnSingleAbility() {
-        List<Ability> oneAbility = examQueryRepository.findAbilities(UUID.fromString("af880054-d1d2-4c24-805c-1f0dfdb45989"),
```
millerts: This was tricky for me to spot.  If you notice the UUID used before my change is different from the the `currentExamId` by a single character (last character is '9' whereas the currentExamId ends with a '0').  It was very hard for me to see that difference so instead I'm creating a random UUID.

In the future we should avoid these very subtle differences if possible because it can be hard for those who haven't seen the code previously figure out what is supposed to happen.

PR #28, opened 2016-11-25
Feature/open previous exam without accommodations

millerts:
This is the first iteration of opening a previous exam.  It simply creates the new exam but there will be a follow up pull request containing the accommodation section of work.


client/src/main/java/tds/exam/ExamStatusStage.java
```diff
@@ -0,0 +1,35 @@
+package tds.exam;
+
+/**
+ * Exam status also has an associated stage.
+ */
+public enum ExamStatusStage {
+    inuse("inuse"),
```
burzum619: Can we use ALL_CAPS convention for enums representing constant values?

client/src/main/java/tds/exam/ExamStatusStage.java
```diff
@@ -0,0 +1,35 @@
+package tds.exam;
+
+/**
+ * Exam status also has an associated stage.
+ */
+public enum ExamStatusStage {
+    INUSE("INUSE"),
```
millerts: @burzum619 Changed to all caps as requested

PR #29, opened 2016-11-28
[Initialize Segments] ComputeSegmentPool + ItemPool Services

burzum619:
This pull request covers code related to the "ComputeSegmentPool" portion of startExam/initializeExamSegments. The general purpose of this code is to select which items will be available for use in the exam (based on exam accommodations and configuration-based constraints on segments), as well as the calculation of the segment length (the number of items that need to be selected for the segment from the pool). This portion of the code deals exclusively with adaptive algorithm based segments/assessments.

- ItemPoolService.getItemPool() is the equivalent of the code located in StudentDLL._AA_ItempoolString_FN() [line 2552]
- SegmentPoolService.computeSegmentPool() represents the logic in StudentDll._ComputeSegmentPool_SP() [line 2842]

- In getItemPool(), the enormous multi-join query has been replaced by in-memory lambda expressions used for filtering eligible items. All relevant data (item constraints, properties, item metadata, etc) aside from the ExamAccommodations are now included as part of the Assessment object.
- computeSegmentPool() does not include "sessionKey" as a parameter, and ignores a portion of code that is only access when sessionKey is not null. This is because the sessionKey is only used in the segment calculation when in simulation mode. 
- This PR does **not** include code related to simulation mode. The simulation implementation will be completed at a later date.


service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+public class ItemPoolServiceImpl implements ItemPoolService {
+    private ExamAccommodationService examAccommodationService;
+
+    @Autowired
+    public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
+        this.examAccommodationService = examAccommodationService;
+    }
+
+    @Override
+    public Set<Item> getItemPool(final UUID examId, final List<ItemConstraint> itemConstraints, final List<Item> items) {
```
burzum619: Did my best to write this method to capture the join logic in _AA_ItempoolString_FNOptimized [StudentDLL 2631], but I imagine there has to be a way to optimize this method and perhaps remove some of the intermediary steps. Certainly open to suggestions here (and of course everywhere else).

service/src/main/java/tds/exam/models/SegmentPoolInfo.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * A class that contains information about a computed segment pool.
+ */
+public class SegmentPoolInfo {
+    private int length;
+    private int poolCount;
+    private Set<String> itemPoolIds;
+
+    public SegmentPoolInfo(Builder builder) {
```
millerts: I don't think a builder is needed for this class.  I think our standard argument count for a builder is 5+.  If you decide to keep the builder the constructor should be private.

service/src/main/java/tds/exam/repositories/ExamAccommodationQueryRepository.java
```diff
@@ -18,4 +18,12 @@
...
      */
     List<ExamAccommodation> findAccommodations(UUID examId, String segmentKey, String[] accommodationTypes);
+
+    /**
+     * Retrieves a list of {@link tds.exam.ExamAccommodation}s for the specified exam.
+     *
+     * @param examId    The id of the {@link tds.exam.Exam} that owns the {@link tds.exam.ExamAccommodation}s
+     * @return  A list of {@link tds.exam.ExamAccommodation}s that correspond to the specified examId
+     */
+    List<ExamAccommodation> findAllAccommodations(UUID examId);
```
millerts: Noting here for others that we've discussed that we both wrote this method.  I'll probably merge first and can help with the conflicts if necessary.

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+import tds.assessment.Item;
+import tds.assessment.ItemConstraint;
+import tds.assessment.ItemProperty;
+import tds.exam.ExamAccommodation;
+import tds.exam.services.ExamAccommodationService;
+import tds.exam.services.ItemPoolService;
+
+@Service
+public class ItemPoolServiceImpl implements ItemPoolService {
+    private ExamAccommodationService examAccommodationService;
```
millerts: Can you make this final?

service/src/test/java/tds/exam/services/impl/AssessmentServiceImplTest.java
```diff
@@ -48,7 +48,7 @@ public void shouldFindSetOfAdminSubjectsByKey() {
         assessment.setKey("key");
         assessment.setAssessmentId("assessmentId");
         assessment.setSegments(segments);
-        assessment.setSelectionAlgorithm(Algorithm.FIXED_FORM.VIRTUAL);
+        assessment.setSelectionAlgorithm(Algorithm.VIRTUAL);
```
millerts: What was the thinking behind this change?  It doesn't alter the unit test results.  If `FIXED_FORM` isn't a valid algorithm we should remove the constant.

service/src/test/java/tds/exam/services/impl/SegmentPoolServiceImplTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SegmentPoolServiceImplTest {
+    private SegmentPoolServiceImpl segmentPoolService;
+    private ItemPoolService mockItemPoolService;
+
+
```
millerts: Extra line break.

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Autowired
+    public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
+        this.examAccommodationService = examAccommodationService;
+    }
+
+    @Override
+    public Set<Item> getItemPool(final UUID examId, final List<ItemConstraint> itemConstraints, final List<Item> items) {
+        /*
+            This method is meant to replace StudentDLL._AA_ItempoolString_FNOptimized() [1643]
```
jtreuting: What is [1643], I'd assume line number but I see that as 2604

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+        */
+        //TODO: remove intermediary steps/optimize
+        List<ExamAccommodation> allAccommodations = examAccommodationService.findAllAccommodations(examId);
+        // Gather the full list of item props
+        List<ItemProperty> allItemProperties = items.stream()
+                .flatMap(item -> item.getItemProperties().stream())
+                .collect(Collectors.toList());
+
+        // First, get all the constraints for our exam accommodations marked as "inclusive"
+        Set<ExamAccommodation> includedAccommodations = allAccommodations.stream()
```
millerts: I think we're iterating over the accommodations more than necessary.  It seems to me if you did it old school you could end up with:

```
Set<ExamAcc> includedAcc = new Set;
Set<ExamAcc> excludedAcc = new Set;

forEach(allAccommodations) {
   if (accommodation.isInclusive...) { add to included acc set}
   if (accommodation.isExclusive..) {add to excluded acc set}
}
```
Then you could do your other stuff.  Still working through that part.

service/src/test/java/tds/exam/services/impl/AssessmentServiceImplTest.java
burzum619: This change snuck in and should not have been included as part of the PR, to be honest. The change has no real impact in terms of the unit test results of functionality, I was just correcting a typo.

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
burzum619: Jeff, you're right, its 2604. I must have been looking at the de-compiled version of StudentDLL.class when I wrote that comment.

service/src/main/java/tds/exam/services/impl/SegmentPoolServiceImpl.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public SegmentPoolInfo computeSegmentPool(final UUID examId, final Segment segment,
+                                              final List<ItemConstraint> itemConstraints) {
+        // Get the list of eligible items based on constraints and exam accommodations
+        Set<Item> itemPool = itemPoolService.getItemPool(examId, itemConstraints, segment.getItems());
+        /* getItemPool selects the items that are eligible for the segment pool we are constructing.
+           In legacy code, we can skip a lot of the temp-table initialization logic because of this */
+        Set<Strand> strands = segment.getStrands();
+        List<SegmentBluePrint> segmentBluePrints = new ArrayList<>();
```
millerts: Can this be a `Set`?

PR #30, opened 2016-11-29
Feature/add exam accommodation event table

millerts:
It seemed there were enough moving pieces around accommodations that it would be good to have this change merged separate from the feature work around opening an exam.

### DIFF
--- a/client/src/main/java/tds/exam/ExamAccommodation.java
+++ b/client/src/main/java/tds/exam/ExamAccommodation.java
@@ -17,6 +17,7 @@ public class ExamAccommodation {
     private String description;
     private Instant deniedAt;
     private Instant createdAt;
+    private Instant deletedAt;
 
     public static class Builder {
         private long id;
@@ -27,6 +28,7 @@ public class ExamAccommodation {
         private String description;
         private Instant deniedAt;
         private Instant createdAt;
+        private Instant deletedAt;
 
         public Builder withId(long id) {
             this.id = id;
@@ -87,6 +89,11 @@ public class ExamAccommodation {
             return this;
         }
 
+        public Builder withDeletedAt(Instant deletedAt) {
+            this.deletedAt = deletedAt;
+            return this;
+        }
+
         public ExamAccommodation build() {
             return new ExamAccommodation(this);
         }
@@ -101,6 +108,7 @@ public class ExamAccommodation {
         description = builder.description;
         deniedAt = builder.deniedAt;
         createdAt = builder.createdAt;
+        deletedAt = builder.deletedAt;
     }
 
     /**
@@ -166,5 +174,18 @@ public class ExamAccommodation {
      */
     public boolean isApproved() {
         return deniedAt == null;
+    }
+
+    /**
+     * Delete date
+     *
+     * @return the delete date
+     */
+    public Instant getDeletedAt() {
+        return deletedAt;
+    }
+
+    public void setId(long id) {
+        this.id = id;
     }
 }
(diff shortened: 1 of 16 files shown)


client/src/main/java/tds/exam/ExamAccommodation.java
```diff
@@ -167,4 +175,17 @@ public Instant getCreatedAt() {
...
+    /**
+     * Delete date
+     *
+     * @return the delete date
+     */
+    public Instant getDeletedAt() {
+        return deletedAt;
+    }
+
+    public void setId(long id) {
```
millerts: Note here the id.  I couldn't think of a better way to do this since the id is auto incrementing.  Used in the command repository.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
```diff
@@ -37,16 +37,28 @@ public ExamAccommodationQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") Name
...
-            "   exam_accommodations \n" +
+            "   exam_accommodation ea \n" +
+            "JOIN ( \n" +
+            "   SELECT \n" +
+            "       exam_accommodation_id, \n" +
+            "       MAX(id) AS id \n" +
+            "   FROM \n" +
+            "       exam_accommodation_event \n" +
+            "   WHERE deleted_at is null \n" +
+            "   GROUP BY exam_accommodation_id \n" +
```
jtreuting: I think I would need to test this out with real data, but it seems to me that this subquery might not be bringing back what is needed, but I might be reading it wrong.

If an accommodation is deleted, then the latest record would have deleted_at set to `null`, but this query would ignore that record and get the previous one where deleted_at is `not null`.

service/src/main/java/tds/exam/repositories/ExamAccommodationCommandRepository.java
```diff
@@ -13,5 +13,12 @@
      *
      * @param accommodations list of {@link tds.exam.ExamAccommodation} to insert
      */
-    void insertAccommodations(List<ExamAccommodation> accommodations);
+    void save(List<ExamAccommodation> accommodations);
```
millerts: This is the convention that we have been using.  I personally don't care as long as it is consistent throughout.  `insert()`?  `update()` only updates certain parts as well so there is nuance regardless of the names we choose.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
millerts: You are correct.  Added a unit test to verify the failure and corrected the query.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
```diff
@@ -91,14 +91,14 @@ public ExamAccommodationQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") Name
...
-                "   WHERE deleted_at is null \n" +
                 "   GROUP BY exam_accommodation_id \n" +
                 ") last_event \n" +
                 "  ON ea.id = last_event.exam_accommodation_id \n" +
                 "JOIN exam_accommodation_event eae \n" +
                 "  ON last_event.id = eae.id \n" +
                 "WHERE \n" +
-                "   exam_id = :examId;";
+                "   exam_id = :examId" +
+                "   AND deleted_at IS NULL;";
```
jtreuting: Hope i'm not picking nits here but i also like the aliases here when there are multiple tables so you can easily tell what table they are coming from

millerts: Fixed

PR #31, opened 2016-11-30
Fieldtest eligibility check

burzum619:
This method will replace the legacy StudentDLL.FT_IsEligible_FN() function (line **4398**)

The purpose of this code is to check to see if a segment (and its parent assessment) contains items eligible for a field test and the exam falls within the eligible field test time window.

Please note that the legacy code references data (field test start/end times) from client_testproperties and client_segmentproperties for assessments and segments, respectively. Per our discussion on 11/29, this data is being moved into the Assessment and Segment classes.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
@@ -4,12 +4,10 @@ import org.joda.time.Instant;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
-import java.util.Date;
 import java.util.Optional;
 
 import tds.assessment.Assessment;
 import tds.assessment.Segment;
-import tds.common.data.legacy.LegacyComparer;
 import tds.config.ClientSegmentProperty;
 import tds.config.ClientTestProperty;
 import tds.exam.Exam;
--- a/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
@@ -26,7 +26,7 @@ public class FieldTestServiceImpl implements FieldTestService {
                 item.getItemProperties().stream()
                     .filter(prop ->
                         prop.getName().equalsIgnoreCase(Accommodation.ACCOMMODATION_TYPE_LANGUAGE) &&
-                        prop.getValue().equals(languageCode))
+                        prop.getValue().equalsIgnoreCase(languageCode))
                     .findFirst().isPresent())
             .count();
 
--- a/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
@@ -3,7 +3,10 @@ package tds.exam.services.impl;
 import org.joda.time.Instant;
 import org.springframework.stereotype.Service;
 
+import java.util.Optional;
+
 import tds.assessment.Assessment;
+import tds.assessment.Item;
 import tds.assessment.Segment;
 import tds.config.Accommodation;
 import tds.exam.Exam;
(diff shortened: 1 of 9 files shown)


service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+            if (ExternalSessionConfiguration.SIMULATION_ENVIRONMENT.equalsIgnoreCase(exam.getEnvironment())) {
+                isEligible = true;
+            } else {
+                /* Line [4473] : client_testproperties already included in Assessment */
+                boolean assessmentEligible = isWithinFieldTestWindow(assessment.getFieldTestStartDate(), assessment.getFieldTestEndDate());
+                /* parentKey == testKey when the assessment is non-segmented */
+                if (!assessment.isSegmented() || !assessmentEligible) {
+                    return assessmentEligible;
+                }
+                /* Line [4491]  : client_segmentproperties already included in Segments */
```
millerts: Similar comment concerning the line number and the accompanying comment about `... already included`.  It doesn't really explain much for someone without knowledge of the Student DLL.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+import tds.assessment.Assessment;
+import tds.assessment.Segment;
+import tds.config.Accommodation;
+import tds.exam.Exam;
+import tds.exam.services.FieldTestService;
+import tds.session.ExternalSessionConfiguration;
+
+@Service
+public class FieldTestServiceImpl implements FieldTestService {
```
millerts: I'm wondering if this really needs to be a full fledged service or can just be a simple POJO?  It currently takes in all its dependencies.  Are you thinking there will be more methods added to the service in the future?

service/src/test/java/tds/exam/services/impl/FieldTestServiceImplTest.java
```diff
@@ -0,0 +1,496 @@
...
+        items.add(item2);
+
+        Segment seg1 = new Segment(segmentKey);
+        seg1.setFieldTestMinItems(1);
+        seg1.setItems(items);
+
+        List<Segment> segments = new ArrayList<>();
+        segments.add(seg1);
+
+        Assessment assessment = new Assessment();
```
millerts: We should consider having a builder for an Assessment that will build out a Segmented or a NonSegmented assessment.  Once that fully populated assessment is built out our tests could simply tweak what is needed to hit different logic paths.

At the very least I think this particular test class can have a private method that builds most of the assessment since it looks like this large block is almost identical in every test in this class.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: I do believe this service will have more methods added to it in the near future.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segmentKey, String languageCode) {
+        boolean isEligible = false;
+        Segment currentSegment = assessment.getSegment(segmentKey);
+
+        // Get the count items that are compatible with the selected language
+        /* StudentDLL [4430] */
+        int fieldTestItemCount = (int) currentSegment.getItems().stream()
```
burzum619: Not sure I follow - I believe "I" is the tblsetofadminitem (not adminsubjects), which contains the individual items that are part of the assessment. I don't think assessments have an "isFieldTest" property. Technically an assessment can contain a combination of FT and non-FT items.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        return isEligible;
+    }
+
+    /*
+        This helper method is a null-tolerant Instant/date comparison for the test window
+     */
+    private boolean isWithinFieldTestWindow(Instant startTime, Instant endTime) {
+        boolean inWindow;
+        // null field test start/end times are considered "always open" windows
+        if (startTime != null) {
```
burzum619: I'll do that. Thanks for the suggestion

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        /* StudentDLL [4430] */
+        int fieldTestItemCount = (int) currentSegment.getItems().stream()
+            .filter(item ->
+                item.isFieldTest() &&
+                item.getItemProperties().stream()
+                    .filter(prop ->
+                        prop.getName().equalsIgnoreCase(Accommodation.ACCOMMODATION_TYPE_LANGUAGE) &&
+                        prop.getValue().equalsIgnoreCase(languageCode))
+                    .findFirst().isPresent())
+            .count();
```
burzum619: Sure, I like that. I'm also going to move the "currentSegment.getFieldTestMinItems() > 0" out of the if because if the condition above is false, theres really no point in iterating over all the items to find a FT item.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+        /* This is line 4437 and 4453 combined into one conditional */
+        if (fieldTestItemCount > 0 && currentSegment.getFieldTestMinItems() > 0) {
+            if (ExternalSessionConfiguration.SIMULATION_ENVIRONMENT.equalsIgnoreCase(exam.getEnvironment())) {
+                isEligible = true;
+            } else {
+                /* Line [4473] : client_testproperties already included in Assessment */
```
burzum619: I'll add more information  to the comment

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
millerts: Can we just flip the annotation to @Component until it has a need for transaction management?

burzum619: Yep, sounds good to me.

PR #32, opened 2016-12-01
Add clientName parameter to AssessmentService.findAssessmentByKey()

jeffjohnson9046:
Add `clientName` parameter to `AssessmentService.findAssessmentKey()` method call.

### DIFF
--- a/service/src/main/java/tds/exam/services/AssessmentService.java
+++ b/service/src/main/java/tds/exam/services/AssessmentService.java
@@ -10,8 +10,10 @@ import tds.assessment.Assessment;
 public interface AssessmentService {
     /**
      * Finds the {@link tds.assessment.Assessment}
+     *
+     * @param clientName The name of the client (e.g. SBAC or SBAC_PT)
      * @param key unique key for the assessment
      * @return {@link tds.assessment.Assessment the assessment}
      */
-    Optional<Assessment> findAssessmentByKey(String key);
+    Optional<Assessment> findAssessmentByKey(String clientName, String key);
 }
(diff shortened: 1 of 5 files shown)


service/src/main/java/tds/exam/services/AssessmentService.java
```diff
@@ -10,8 +10,10 @@
 public interface AssessmentService {
     /**
      * Finds the {@link tds.assessment.Assessment}
+     *
+     * @param clientName The name of the client (e.g. SBAC or SBAC_PT)
      * @param key unique key for the assessment
      * @return {@link tds.assessment.Assessment the assessment}
      */
-    Optional<Assessment> findAssessmentByKey(String key);
+    Optional<Assessment> findAssessmentByKey(String clientName, String key);
```
jeffjohnson9046: Throughout most of our code, the signature for `findAssessmentByKey` is `findAssessmentByKey(clientName, assessmentKey`.  However, in the TDS_AssessmentService, the `AssessmentQueryRepository` has the signature set to `findAssessmentByKey(assessmentKey, clientName)` (the parameters are reversed).  At some point, we should probably come back and clean that up, but I think that's a separate pull request.

millerts: I vote that you change it now.

jtreuting: I know I'm totally late to the game here, but do we need the `ByKey` part?  Seems like that is the implied behavior of an `findAssessment` call.

millerts: No.  I'll make the change in a PR I'll have up in a few

PR #35, opened 2016-12-09
InitializeSegments implementation

burzum619:
The main method of focus for this pull request is ExamSegmentService.initializeExamSegments() which replaced StudentDLL._InitializeTestSegments_SP() in the legacy application.

In summary, initializeExamSegments() does the following:

         1. Loops over each Segment in the selected assessment, creating an exam-specific representation of the segment
         2. ExamSegment is populated with various pieces of data that are dependent on the selection algorithm
            a. fixed form segments contain form data, and the number of items is fixed based on the language
            b. adaptive segments contain a segment pool containing all possible items, as well as the number of items
               that need to be selected from the segment pool
         3. Inserts a record into exam_segments for each segment initialized

- For the time  being, FieldTestService.selectItemGroups() is stubbed out until feedback is received from Smarter Balanced and the method is implemented. All other code that is not simulation mode-specific called by InitializeSegments is complete.
- Form cohort related code is currently being ignored while we determine whether or not form cohorts are actually utilized by Smarter Balanced.
- Contains some minor changes to support Segment refactor

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
@@ -5,7 +5,11 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
+import org.springframework.jdbc.core.namedparam.SqlParameterSourceUtils;
 import org.springframework.stereotype.Repository;
+
+import java.util.ArrayList;
+import java.util.List;
 
 import tds.common.data.mapping.ResultSetMapperUtility;
 import tds.common.data.mysql.UuidAdapter;
@@ -29,25 +33,28 @@ public class ExamSegmentCommandRepositoryImpl implements ExamSegmentCommandRepos
      * @inheritDoc
      */
     @Override
-    public void insert(final ExamSegment segment) {
-        final SqlParameterSource parameters =
-                new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(segment.getExamId()))
-                        .addValue("assessmentSegmentKey", segment.getAssessmentSegmentKey())
-                        .addValue("assessmentSegmentId", segment.getAssessmentSegmentId())
-                        .addValue("segmentPosition", segment.getSegmentPosition())
-                        .addValue("formKey", segment.getFormKey())
-                        .addValue("formId", segment.getFormId())
-                        .addValue("algorithm", segment.getAlgorithm())
-                        .addValue("examItemCount", segment.getExamItemCount())
-                        .addValue("fieldTestItemCount", segment.getFieldTestItemCount())
-                        .addValue("fieldTestItems", String.join(",", segment.getFieldTestItems()))
-                        .addValue("formCohort", segment.getFormCohort())
-                        .addValue("poolCount", segment.getPoolCount())
-                        .addValue("isSatisfied", segment.isSatisfied())
-                        .addValue("isPermeable", segment.isPermeable())
-                        .addValue("restorePermeableOn", segment.getRestorePermeableCondition())
-                        .addValue("dateExited", ResultSetMapperUtility.mapInstantToTimestamp(segment.getDateExited()))
-                        .addValue("itemPool", String.join(",", segment.getItemPool()));
+    public void insert(final List<ExamSegment> segments) {
+        final List<SqlParameterSource> parameterSources = new ArrayList<>();
+        segments.forEach(segment -> {
+                SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(segment.getExamId()))
+                    .addValue("segmentKey", segment.getSegmentKey())
+                    .addValue("segmentId", segment.getSegmentId())
+                    .addValue("segmentPosition", segment.getSegmentPosition())
+                    .addValue("formKey", segment.getFormKey())
+                    .addValue("formId", segment.getFormId())
+                    .addValue("algorithm", segment.getAlgorithm().getType())
+                    .addValue("examItemCount", segment.getExamItemCount())
+                    .addValue("fieldTestItemCount", segment.getFieldTestItemCount())
+                    .addValue("formCohort", segment.getFormCohort())
+                    .addValue("poolCount", segment.getPoolCount())
+                    .addValue("isSatisfied", segment.isSatisfied())
+                    .addValue("isPermeable", segment.isPermeable())
+                    .addValue("restorePermeableOn", segment.getRestorePermeableCondition())
+                    .addValue("dateExited", ResultSetMapperUtility.mapInstantToTimestamp(segment.getDateExited()))
+                    .addValue("itemPool", String.join(",", segment.getItemPool()));
+
+                parameterSources.add(parameters);
+            });
 
         final String segmentQuery =
                 "INSERT INTO exam_segment (\n" +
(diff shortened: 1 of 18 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
```diff
@@ -119,4 +120,42 @@ public void update(final ExamSegment segment) {
 
         jdbcTemplate.update(segmentEventQuery, parameters);
     }
+
+    @Override
+    public void update(final List<ExamSegment> segments) {
```
jtreuting: Seems like we should combine the `update(List)` and `update(obj)` so that the logic is not duplicated.  How about `update(obj)` putting the item into a list and calling `update(List)`

service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
```diff
@@ -60,36 +67,30 @@ public void insert(final ExamSegment segment) {
...
                 "   :fieldTestItemCount, \n" +
-                "   :fieldTestItems, \n" +
                 "   :formCohort, \n" +
                 "   :poolCount \n" +
                 ")";
 
-        jdbcTemplate.update(segmentQuery, parameters);
-        update(segment);
+        jdbcTemplate.batchUpdate(segmentQuery, parameterSources.toArray(new SqlParameterSource[parameterSources.size()]));
+        update(segments);
```
jtreuting: Don't we need these 2 statements to run in a transaction so that it's all or nothing?

burzum619: Agreed, but do we want transactions at both the service level (not yet implemented) and the repo level?

service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
burzum619: Yep, I like that idea.

millerts: I did a similar thing in my branch but I used a varargs.  You can see it [here](https://github.com/SmarterApp/TDS_ExamService/blob/feature/previousExamAccommodations/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java#L59)

service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
jtreuting: Yeah service level makes sense to me for sure.  This was just the first time I specifically saw a place where it was needed because of 2 separate inserts.  Good to hear we have a JIRA for it, so that's good.

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                selectedForm = ((FixedFormSegment)segment).getForm(languageCode);
+                poolCount = selectedForm.getLength();
+            } else { // Algorithm is adaptive2
+                segmentPoolInfo = segmentPoolService.computeSegmentPool(exam.getId(),
+                    (AdaptiveSegment) segment, assessment.getItemConstraints(), languageCode);
+                itemPool = segmentPoolInfo.getItemPool();
+                poolCount = segmentPoolInfo.getPoolCount(); // poolCount does not always == itemPool.size
+
+                /* In legacy, opitemcnt = segment's max items. See lines [4624], [4630], [4672] */
+                if (fieldTestService.isFieldTestEligible(exam, assessment, segment.getKey(), languageCode)
```
millerts: Which line in StudentDLL contains this logic check?  I can't follow based on the comment right above the conditional.

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                    && segmentPoolInfo.getLength() == segment.getMaxItems()) {
+                    fieldTestItemCount = fieldTestService.selectItemGroups(exam, assessment, segment.getKey(), languageCode);
+                }
+
+                if (fieldTestItemCount + segmentPoolInfo.getLength() == 0) {
```
millerts: Could change this to `isSatisfied = fieldTestItemCount + segmentPoolInfo.getLength() == 0;`

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                    isSatisfied = true;
+                }
+            }
+
+            // Keep track during the segment loop if items exist(do not need to check again if itemsInPool = true in prior segment)
```
millerts: Where is itemsInPool?  Reading this comment it makes me think there should be a boolean variable of the same name.

service/src/test/java/tds/exam/services/impl/ExamSegmentServiceImplTest.java
```diff
@@ -0,0 +1,492 @@
...
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
+@Transactional
+public class ExamSegmentServiceImplTest {
```
millerts: Do you need to use any of the annotations?  It looks like this is a unit test not an integration test with you passing in all the dependencies.  Am i missing the use of Spring somewhere?  If this was done to get the `@Captor` to work you should instead replace the annotations above with `@RunWith(MockitoJUnitRunner.class)`.  That'll trigger the captor annotation.  If you wanted to take it a step further you could also get rid of the `mock(Class)` in the setup and instead leverage the `@Mock` annotation.  Your call on that last one.

service/src/test/java/tds/exam/services/impl/ExamSegmentServiceImplTest.java
```diff
@@ -0,0 +1,492 @@
...
+    @Test(expected = IllegalStateException.class)
+    public void shouldThrowExceptionForNoItemsFound() {
+        Exam exam = new ExamBuilder().build();
+        ExamAccommodation language = new ExamAccommodation.Builder()
+            .withCode("ENU")
+            .build();
+        AdaptiveSegment segment = (AdaptiveSegment) new SegmentBuilder()
+            .withSelectionAlgorithm(Algorithm.ADAPTIVE_2)
+            .withMaxItems(5)
+            .build();
```
millerts: Instead of a build method maybe add a `buildFixedForm()` and `buildAdaptive()` to the builder so you don't have to cast?

service/src/main/java/tds/exam/services/FieldTestService.java
```diff
@@ -18,4 +18,6 @@
      * @return  true if the the exam segment is eligible for a field test, false otherwise
      */
     boolean isFieldTestEligible(Exam exam, Assessment assessment, String segmentKey, String languageCode);
+
+    int selectItemGroups(Exam exam, Assessment assessment, String segmentKey, String languageCode);
```
burzum619: Didn't add javadocs to this method yet because the requirements are still unclear and it has not been implemented yet.

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
burzum619: Sorry, the variable name was changed but the comment wasn't updated. Fixed.

service/src/test/java/tds/exam/services/impl/ExamSegmentServiceImplTest.java
burzum619: Good call, thanks.

service/src/main/java/tds/exam/services/FieldTestService.java
millerts: Why is it being added then?  Seems like we shouldn't add it until we implement it.

service/src/test/java/tds/exam/services/impl/ExamSegmentServiceImplTest.java
millerts: After our in person review this is no longer necessary.

burzum619:
Closing PR. Will create new PR with changes to form selection + integration w/ recent assessment service refactoring.

PR #36, opened 2016-12-19
Feature/previous exam accommodations

millerts:
This is the code that initializes the exam accommodations when opening a previous exam.


client/src/main/java/tds/exam/ExamAccommodation.java
```diff
@@ -12,12 +12,17 @@
...
     private String type;
     private String code;
+    private String value;
     private String description;
+    private boolean selectable;
+    private boolean allowChange;
     private Instant deniedAt;
     private Instant createdAt;
     private Instant deletedAt;
+    private boolean multipleToolTypes;
```
jtreuting: Is there a benefit to storing this as a boolean, which is specific to the use case as I remember, instead of `int numToolTypes` with a helper getter like `boolean isMultipleToolTypes() { return numToolTypes > 1 }`?

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -67,7 +68,8 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
             "   ee.abnormal_starts, \n" +
             "   ee.waiting_for_segment_approval, \n" +
             "   ee.current_segment_position, \n" +
-            "   e.created_at, \n" +
+            "   ee.custom_accommodations, \n" +
+            "   ee.language, \n" +
```
jtreuting: Same question as above, is this language or languageCode?

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -70,4 +84,153 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
...
+         */
+        List<ExamAccommodation> examAccommodations = findAllAccommodations(exam.getId());
+        if (examAccommodations.isEmpty()) {
+            initializeExamAccommodations(exam);
+        } else {
+            initializePreviousAccommodations(exam, assessment, segmentPosition, restoreRts, guestAccommodations, examAccommodations);
+        }
+
+        //Fetch the updated exam accommodations
+        examAccommodations = findAllAccommodations(exam.getId());
```
jtreuting: Is it possible to return the accommodations from the initialize methods called right before this to save a DB trip?

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -70,4 +84,153 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
...
+
+        //StudentDLL lines 6967 - 6875
```
jtreuting: On develop, I'm seeing the line numbers as either 6857 - 6868 or maybe 6865 - 6868

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -70,4 +84,153 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
...
+            .toArray(ExamAccommodation[]::new);
+
+        examAccommodationCommandRepository.update(examAccommodationsToDenyApproval);
+    }
+
+    private static String getOtherAccommodationValue(String formattedValue) {
+        return formattedValue.substring("TDS_Other#".length());
+    }
+
+    private List<String> splitAccommodationCodes(String accommodationFamily, String guestAccommodations) {
```
jtreuting: Let's run thru this when we do our review meeting tomorrow.  The `_SplitAccomCodes_FN` method is hard to follow.

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -70,4 +84,153 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
...
+         This replaces the functionality of the following bits of code
+         - StudentDLL 6834 - 6843
+         - StudentDLL _InitOpportunityAccommodations_SP
+         - CommonDLL _UpdateOpportunityAccommodations_SP
+         */
+        List<ExamAccommodation> examAccommodations = findAllAccommodations(exam.getId());
+        if (examAccommodations.isEmpty()) {
+            initializeExamAccommodations(exam);
+        } else {
+            initializePreviousAccommodations(exam, assessment, segmentPosition, restoreRts, guestAccommodations, examAccommodations);
```
jtreuting: What is `segmenetPosition` here?  I see in `initializePreviousAccommodations` it matches on segment (which is 0 for the entire assessment and X for a specific segment, as I recall).  I think I'm confused what would be passed in here.  Isn't this initializing accommodations for an exam based on a previous exam?  So would it need to do it for all segments?

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -70,4 +84,153 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
...
+                    .withAllowChange(false)
+                    .withSelectable(false)
+                    .withSegmentPosition(segmentPosition)
+                    .build();
+
+                break;
+            }
+        }
+
+        final Optional<ExamAccommodation> maybeOtherExamAccommodation = Optional.ofNullable(otherExamAccommodation);
```
jtreuting: It feels like the use of Optional here is not needed since it is within the internals of the method and not used to convey meaning as the return type of the service.  Using `otherExamAccommodation` below seems just as clean to me:  `.filter(accommodation -> otherExamAccommodation == null || accommodation.getType().equals(otherExamAccommodation.getType()))`

client/src/main/java/tds/exam/Exam.java
```diff
@@ -64,7 +66,8 @@
         private int abnormalStarts;
         private boolean waitingForSegmentApproval;
         private int currentSegmentPosition;
-
+        private String language;
```
millerts: Yes, this should probably be language code.  I used language because that is what legacy stores.  I can make the change.

burzum619: I believe it is "languageCode" technically, which I've added in my PR for initializeSegments

client/src/main/java/tds/exam/ExamAccommodation.java
millerts: I'll change it to store the int.  I think this check is only done once so a convenience method isn't needed.

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
millerts: It is 0 in the legacy code. I didn't sweep through the legacy code for all the calls to the method this is replacing.  I can do so.

client/src/main/java/tds/exam/Exam.java
millerts: In the legacy code it is the language not the code.  I left it as is because I wanted to avoid all the churn that was happening with language and assessment.  Now that that work is complete I will make the necessary change to use language code.

PR #37, opened 2016-12-19
Feature/initialize segments 2.0

burzum619:
This is an update to initializeExamSegments, which code reviewed in a previous pull request. This PR includes updates to support the segment refactor (in particular, removing segment subclassing) and the implementation of a round robin form selector for fixed form assessments with multiple forms.

Specific changes  that were not a part of the previous pull request include:
- Removed ExamAccommodationService in ExamSegmentService - the service was used to retrieve the languageCode for the Exam, but that code is now included in the Exam object itself. The language_code column was also added to exam_event.
- Added FormSelector/RoundRobinFormSelectorImpl and unit tests. 
- Added form cohort support
- Returning the number of total items (field test items + regular items) for all exam segments instead of void
- Removed "hasItems" flag in initializeExamSegments(), instead check if total items in exam > 0;


service/src/main/java/tds/exam/services/impl/RoundRobinFormSelectorImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import tds.assessment.Form;
+import tds.assessment.Segment;
+import tds.exam.services.FormSelector;
+
+@Component
+public class RoundRobinFormSelectorImpl implements FormSelector {
+    // Keeps track of the index of the next form to assign for a segment
+    private Cache<String, FormIndex> formIndexMap;
```
burzum619: Guava Cache is (fortunately for us) thread safe by nature.

service/src/main/java/tds/exam/services/impl/RoundRobinFormSelectorImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import tds.assessment.Form;
+import tds.assessment.Segment;
+import tds.exam.services.FormSelector;
+
+@Component
+public class RoundRobinFormSelectorImpl implements FormSelector {
+    // Keeps track of the index of the next form to assign for a segment
+    private Cache<String, FormIndex> formIndexCache;
```
burzum619: Guava Cache is (fortunately for us) thread safe by nature. Decided to roll with a Cache with a very long expiration time here as opposed to a ConcurrentHashMap to prevent the (probably rare) case of the formIndexCache growing indefinitely and eating up memory over time. Thanks to @millerts for the thought

service/src/main/java/tds/exam/repositories/impl/ExamSegmentCommandRepositoryImpl.java
```diff
@@ -29,25 +34,28 @@ public ExamSegmentCommandRepositoryImpl(final @Qualifier("commandJdbcTemplate")
...
-                        .addValue("formCohort", segment.getFormCohort())
-                        .addValue("poolCount", segment.getPoolCount())
-                        .addValue("isSatisfied", segment.isSatisfied())
-                        .addValue("isPermeable", segment.isPermeable())
-                        .addValue("restorePermeableOn", segment.getRestorePermeableCondition())
-                        .addValue("dateExited", ResultSetMapperUtility.mapInstantToTimestamp(segment.getDateExited()))
-                        .addValue("itemPool", String.join(",", segment.getItemPool()));
+    public void insert(final List<ExamSegment> segments) {
+        final List<SqlParameterSource> parameterSources = new ArrayList<>();
+        segments.forEach(segment -> {
```
jeffjohnson9046: You might be able to use `map` here instead of `forEach`, maybe something like this (which is pseudo-code, but should be fairly close):
```
final List<SqlParameterSource> parameterSources = segments.stream()
    .map(s -> { // build a new SqlParameterSource with values here... })
```

... that way seems a bit easier to read to me, and it's more purely functional in the sense that it doesn't have any side effects.  the `forEach` call modifies `parameterSources` on each iteration, whereas using `map` won't - the `map` call will build the `parameterSources` list.

Incidentally, I also think it's a wee bit more readable because it says what you're doing:  mapping a collection of `Segment`s to a collection of `MapSqlParameterSource`s.

All that said, this'll work just fine the way it is (obviously).

service/src/test/resources/application.properties
```diff
@@ -1,15 +1,17 @@
-#flyway.enabled=
-flyway.url=
-flyway.user=
-flyway.password=
+flyway.enabled=true
```
jeffjohnson9046: I don't think this file was meant to be checked in.

service/src/main/java/tds/exam/services/impl/RoundRobinFormSelectorImpl.java
```diff
@@ -0,0 +1,71 @@
...
+            .build();
+    }
+
+    @Override
+    public Optional<Form> selectForm(Segment segment, String languageCode) {
+        List<Form> forms = segment.getForms(languageCode);
+        int index = 0;
+        Optional<Form> selectedForm = Optional.empty();
+
+        if (!forms.isEmpty()) {
```
millerts: This isn't a huge deal in the grand scheme of things but I prefer early returns rather than this type of if block.  Doing that would make this:

```{java}
if(forms.isEmpty) {
  return Optional.empty();
}

//Get size once
int formSize = forms.size();
if(formSize > 1) {
  index = selectNextIndex(segment.getKey(), formSize);
}

return Optional.of(forms.get(index));
```

I think early returns make code easier to read.

service/src/main/resources/db/migration/V1481914584__exam_add_columns.sql
```diff
@@ -0,0 +1,18 @@
...
+
+***********************************************************************************************************************/
+
+USE exam;
+
+ALTER TABLE exam_event
+  ADD COLUMN expire_from DATETIME(3) DEFAULT NULL;
+
+ALTER TABLE exam_event
+  ADD COLUMN max_items INT(11) DEFAULT 0;
```
millerts: suggest making this not null

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -111,6 +115,7 @@ public void setUp() {
             mockHistoryRepository,
             mockSessionService,
             mockStudentService,
+            mockExamSegmentService,
```
millerts: Is this here because of some future work?  Surprised to see a dependency added without a subsequent test change.

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+        Optional<Form> maybeRetForm3B = formSelector.selectForm(segment, languageCode);
+        assertThat(maybeRetForm3B).isPresent();
+        Form retForm3B = maybeRetForm3B.get();
+        assertThat(retForm3B.getKey()).isEqualTo(enuForm3.getKey());
+
+        Optional<Form> maybeRetForm4B = formSelector.selectForm(segment, languageCode);
+        assertThat(maybeRetForm4B).isPresent();
+        Form retForm4B = maybeRetForm4B.get();
+        assertThat(retForm4B.getKey()).isEqualTo(enuForm4.getKey());
+
```
millerts: Not critical but mind removing the extra whitespace?

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+        final String languageCode = "ENU";
+
+        Form enuForm1 = new Form.Builder("formkey1")
+            .withLanguage(languageCode)
+            .build();
+        Form enuForm2 = new Form.Builder("formkey2")
+            .withLanguage(languageCode)
+            .build();
+        Form enuForm3 = new Form.Builder("formkey3")
+            .withLanguage(languageCode)
```
millerts: I know this isn't part of your changes but I would like to have `withLanguage` be `withLanguageCode` if that is the intended value. Especially since both are present in the data.

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+            .build();
+        Form esnForm = new Form.Builder("formkey5")
+            .withLanguage("ESN")
+            .build();
+
+        Segment segment = new Segment("segmentKey", Algorithm.FIXED_FORM);
+        segment.setForms(Arrays.asList(enuForm1, enuForm2, enuForm3, enuForm4, esnForm));
+
+        Optional<Form> maybeRetForm1A = formSelector.selectForm(segment, languageCode);
```
millerts: Can we somehow name this in conjunction to what we expect?  It is hard for me to follow the expected behavior by the vague names.  For example, why can't this be `maybeReturnEnuForm1` since that is what we expect here, correct?

millerts: Same comment for the names throughout.

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -81,11 +86,22 @@ public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
                     .map(accommodation -> itemProperty.getItemId()))
                 .collect(Collectors.toSet());
 
-        Set<Item> itemPool = items.stream()
+        Set<Item> itemPool;
+
+        if (isFieldTest != null) {
```
millerts: I think this conditional can be collapsed into this:

```{java}
itemPool = items.stream()
  .filter(item ->
    (itemFieldTest == null || item.isFieldTest == isFieldTest)
    itemPoolIds.contain(item.getId()) && 
    !excludedItemIds.contains(item.getId()))
  .collect(Collectors.toSet());
```

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -25,7 +25,12 @@ public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
     }
 
     @Override
-    public Set<Item> getItemPool(final UUID examId, final List<ItemConstraint> itemConstraints, final List<Item> items) {
+    public Set<Item> getItemPool(UUID examId, List<ItemConstraint> itemConstraints, List<Item> items) {
+        return getItemPool(examId, itemConstraints, items, null);
+    }
+
+    @Override
+    public Set<Item> getItemPool(final UUID examId, final List<ItemConstraint> itemConstraints, final List<Item> items, Boolean isFieldTest) {
```
millerts: Why the Boolean object instead of the primitive?

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+
+            /* NOTE: Skipping [4660-4678]. This segment of code just increments the position and re-iterates to the next
+                segment. This is done in case the minimumSegmentPosition is not found in the temp table they create.
+                That can never happen with our data structure, as it is always 1-based segment positioning. */
+            if (Algorithm.FIXED_FORM == segment.getSelectionAlgorithm()) {
+                /* If no form cohort is defined, this is likely the first form being selected */
+                if (formCohort == null) {
+                    Optional<Form> maybeSelectedForm = formSelector.selectForm(segment, exam.getLanguageCode());
+
+                    if (!maybeSelectedForm.isPresent()) {
```
millerts: I'd like to see some unit tests around these Exceptional states.  I don't see any that have the `@Test(expected = IllegalStateException.class)`

millerts: All of them within this class if possible.

service/src/main/java/tds/exam/services/impl/RoundRobinFormSelectorImpl.java
burzum619: At my old employer, they were ironically pretty strict about single returns at the end of the method wherever possible, so this habit sort of stuck with me. I'll take your suggestion though.

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
burzum619: The lettering at the end (1A, 3B) is to emphasize the ordering, but I can certainly add the language code prefix to the property name if you feel that makes it easier to follow.

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
burzum619: It is nullable

PR #39, opened 2016-12-21
Start new Exam

burzum619:
This pull request contains code for starting/initializing a new exam.

In legacy performance code, the entry point for startExam() is TestOpportunityServiceImpl.startTestOpportunity() [line 119] and the original code is found in StudentDLL.T_StartTestOpportunity_SP() [5242]

Starting an existing exam will be implemented in a separate pull request.


service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -62,6 +63,25 @@ public ExamController(ExamService examService) {
         return new ResponseEntity<>(exam, headers, HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{examId}/start", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<Response<ExamConfiguration>> startExam(@PathVariable final UUID examId) {
+        Response<ExamConfiguration> examConfiguration = examService.startExam(examId);
+
+        if ((examConfiguration.getData().isPresent() && examConfiguration.getData().get().getFailureMessage() != null)
```
burzum619: If we decide to stick the failure data (previously in the ReturnStatus within TestConfig) then this conditional will be simplified.

client/src/main/java/tds/exam/ExamStatusStage.java
```diff
@@ -4,6 +4,7 @@
  * Exam status also has an associated stage.
  */
 public enum ExamStatusStage {
+    INPROGRESS("inprogress"),
```
millerts: Change them.  This was an oversight on my part.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         return Optional.empty();
     }
 
+    @Override
+    public Response<ExamConfiguration> startExam(final UUID examId) {
+        ExamConfiguration examConfig = null;
+
+        try {
+            Exam exam = examQueryRepository.getExamById(examId)
+                .orElseThrow(() -> new IllegalArgumentException(String.format("No exam found for id %s", examId)));
```
millerts: I feel this should throw a NotFound exception.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+            /* TestOpportunityServiceImpl [155] No need to go any further, so moving before service calls */
+            if (!exam.getStatus().getStatus().equalsIgnoreCase(ExamStatusCode.STATUS_APPROVED)) {
+                throw new IllegalStateException(String.format("Cannot start exam %s: Exam was not approved.", examId));
```
millerts: I'm not sure an `IllegalStateException` is the proper thing to throw.  Currently, `IllegalStateException` is (or should be) reserved for when the database is in a bad state usually from either manual loading of data or the loader scripts. 

I actually don't think this should throw at all and simply return a validation error.  In my opinion this isn't exceptional.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+                return new Response<ExamConfiguration>(maybeAccessViolation.get());
+            }
+            /* TestOpportunityServiceImpl [147] */
+            Assessment assessment = assessmentService.findAssessment(exam.getClientName(), exam.getAssessmentKey())
+                .orElseThrow(() -> new IllegalArgumentException(String.format("No assessment found for assessment key '%s'.", exam.getAssessmentKey())));
+
+            TimeLimitConfiguration timeLimitConfiguration =
+                timeLimitConfigurationService.findTimeLimitConfiguration(exam.getClientName(), assessment.getAssessmentId())
+                    .orElseThrow(() ->
+                        new IllegalArgumentException(String.format("No time limit configurations found for clientName '%s' and assessment id '%s'.",
```
millerts: Going along with my earlier comment isn't the only way the `TimeLimitConfiguration` isn't present is if something was configured improperly (either loader or manual loading)?  If so, I think this would be a good use of `IllegalStateException`.

That or we should create our own RuntimeException that `InvalidConfigurationException`.  That may be better than using `IllegalStateException` based on the [description](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalStateException.html).

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+            if (exam.getDateStarted() == null) { // Start a new exam
+                // Initialize the segments in the exam and get the testlength.
+                testLength = initializeExam(exam, assessment);
+            /* StudentDLL [5367] and TestOppServiceImpl [167] */
+                examConfig = ExamConfigurationHelper.getNew(examId, assessment, timeLimitConfiguration, testLength);
+            } else { //Restart the most recent exam
+                //TODO: Start existing opportunity [TDS-350]
+            }
+        } catch (IllegalArgumentException e) {
```
millerts: This can be collapsed to `catch (IllegalArgumentException | IllegalStateException e)`. 

That being said I'm questioning why this is catching Exceptions that are being thrown throughout this method.  We're basically using Exception for logic branching which is not ideal.  We should instead use early returns and create the appropriate response when errors happen instead of throwing.

client/src/main/java/tds/exam/ExamConfiguration.java
```diff
@@ -0,0 +1,201 @@
+package tds.exam;
+
+import java.util.UUID;
+
+/**
+ * Class containing exam configuration data
+ */
+public class ExamConfiguration {
+    private UUID examId;
```
millerts: I think this should hold an Exam not just the id.  Some of the data here can be fetched from the exam instead.  Especially since the exam itself is modified.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+                .withExamId(examId)
+                .withStatus(ExamStatusCode.STATUS_FAILED)
+                .withFailureMessage(e.getMessage())
+                .build();
+        }
+
+        return new Response<>(examConfig);
+    }
+
+    private int initializeExam(final Exam exam, final Assessment assessment) {
```
millerts: Along with my other comment I think this should return the new Exam or even just return the ExamConfig since that is what happens immediately with the return value.

service/src/main/java/tds/exam/utils/ExamConfigurationHelper.java
```diff
@@ -0,0 +1,36 @@
...
+ * A utility class for creating {@link tds.exam.ExamConfiguration} objects.
+ */
+public class ExamConfigurationHelper {
+    private static final int CONTENT_LOAD_TIMEOUT = 120;
+
+    /*
+        This method mimics the legacy TestConfigHelper.getNew().
+        - scoreByTds has been removed as it is unused by the application.
+     */
+    public static ExamConfiguration getNew(UUID examId, Assessment assessment, TimeLimitConfiguration timeLimitConfiguration,
```
millerts: Do we need a separate class for this method?  I think it should either be in Exam as a private static since that seems to be the only place it is used or it can be a public static method on ExamConfiguration. 

I also think the name should be more descriptive.  Something like `initializeDefaultExamConfiguration`.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
millerts: As discussed offline I think if we stick with the current way we're returning errors in the other endpoints in the new system this can be simplified to  `if(examConfiguration.getErrors().isPresent)`?

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -62,6 +63,25 @@ public ExamController(ExamService examService) {
         return new ResponseEntity<>(exam, headers, HttpStatus.OK);
     }
 
+    @RequestMapping(value = "/{examId}/start", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: I think this should be a `PUT` because we're updating the exam.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -62,6 +63,25 @@ public ExamController(ExamService examService) {
...
+    ResponseEntity<Response<ExamConfiguration>> startExam(@PathVariable final UUID examId) {
+        Response<ExamConfiguration> examConfiguration = examService.startExam(examId);
+
+        if ((examConfiguration.getData().isPresent() && examConfiguration.getData().get().getFailureMessage() != null)
+            || !examConfiguration.getData().isPresent() || examConfiguration.getErrors().isPresent()) {
+            return new ResponseEntity<>(examConfiguration, HttpStatus.UNPROCESSABLE_ENTITY);
+        }
+
+        Link link = linkTo(
```
millerts: I'm not sure this should be included with my comment above about this method being a PUT.

service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -46,6 +47,14 @@
     Response<ExamApproval> getApproval(ApprovalRequest approvalRequest);
 
     /**
+     * Starts a new or existing exam.
+     *
+     * @param examId The exam to start
+     * @return The exams configuration
```
millerts: Can we do something like `@return {@link tds.common.Response} containing  {@link tds.exam.ExamConfiguration} or containing validation errors`

millerts: If the others don't do that I'll fix it in my PR

client/src/main/java/tds/exam/ExamConfiguration.java
millerts: The legacy mapper that we're going to write will just handle that.

service/src/main/java/tds/exam/utils/ExamConfigurationHelper.java
jtreuting: I like the public static on ExamConfiguration

millerts:
I noticed that we didn't port over the isMsb method in the test config.  I'm not 100% sure how we'll integrate that but it currently is stored on the `testopportunity`.

service/src/main/java/tds/exam/utils/ExamConfigurationHelper.java
burzum619: I'm on board with renaming the method name, but it cannot be refactored into ExamConfiguration (nor Exam) because it references TimeLimitConfiguration and Assessment classes, which are in TDS_AssessmentService client jars (and not a dependency of Exam client jars)

burzum619: I wouldn't be opposed to making this a private static method inside of ExamServiceImpl.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -273,6 +276,86 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+                        new IllegalArgumentException(String.format("No time limit configurations found for clientName '%s' and assessment id '%s'.",
+                            exam.getClientName(), assessment.getAssessmentId())));
+
+            /* StudentDLL [5344] Skipping getInitialAbility() call here - the ability is retrieved in legacy but never set on TestConfig */
+            int testLength;
+
+            if (exam.getDateStarted() == null) { // Start a new exam
+                // Initialize the segments in the exam and get the testlength.
+                testLength = initializeExam(exam, assessment);
+            /* StudentDLL [5367] and TestOppServiceImpl [167] */
```
burzum619: Fixed. Not sure why intellij insists on formatting block (non-javadoc) comments this way.

service/src/main/java/tds/exam/utils/ExamConfigurationHelper.java
jeffjohnson9046: Of the two options, I think I prefer having it be a private static method on the `ExamServiceImpl`.

PR #40, opened 2016-12-24
Feature/pause exam

jeffjohnson9046:
Added `pauseExam` endpoint

**NOTE:** This pull request depends on a change in the **TDS_Common** repo.  That change is described in this [pull request](https://github.com/SmarterApp/TDS_Common/pull/10).


service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -79,47 +78,47 @@ public void shouldCreateAndRetrieveSingleExamSegment() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(permeable)
-                .withIsSatisfied(satisfied)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withRestorePermeableCondition(condition)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .build();
+            .withSegmentId(segmentId1)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -79,47 +78,47 @@ public void shouldCreateAndRetrieveSingleExamSegment() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(permeable)
-                .withIsSatisfied(satisfied)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .withRestorePermeableCondition(condition)
-                .build();
+            .withSegmentId(segmentId2)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -159,51 +158,51 @@ public void shouldCreateAndRetrieveLatestExamSegment() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(false)
-                .withIsSatisfied(false)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withRestorePermeableCondition(condition)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .build();
+            .withSegmentId(segmentId)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -159,51 +158,51 @@ public void shouldCreateAndRetrieveLatestExamSegment() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(newSatisfied)
-                .withIsSatisfied(newPermeable)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .withRestorePermeableCondition("new condition")
-                .build();
+            .withSegmentId(segmentId)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(permeable)
-                .withIsSatisfied(satisfied)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withRestorePermeableCondition(condition)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .build();
+            .withSegmentId(segmentId1)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(permeable)
-                .withIsSatisfied(satisfied)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .withRestorePermeableCondition(condition)
-                .build();
+            .withSegmentId(segmentId2)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
+            .withSegmentKey(segmentKey2)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(permeable)
-                .withIsSatisfied(satisfied)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .withRestorePermeableCondition(condition)
-                .build();
+            .withSegmentId(segmentId2)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(true)
-                .withIsSatisfied(true)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withRestorePermeableCondition(condition)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .build();
+            .withSegmentId(segmentId1)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
```diff
@@ -250,108 +249,108 @@ public void shouldCreateAndRetrieveAllLatestSegments() {
...
-                .withFormCohort(cohort)
-                .withIsPermeable(true)
-                .withIsSatisfied(false)
-                .withPoolCount(poolCount)
-                .withItemPool(itemPool)
-                .withRestorePermeableCondition(condition)
-                .withFormKey(formKey)
-                .withFormId(formId)
-                .build();
+            .withSegmentId(segmentId2)
```
jeffjohnson9046: Only change here is formatting - looks like I have 4 space indentation vs 8 space indentation that was on the file previously.

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
```diff
@@ -18,11 +14,9 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 
-@RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest
+
```
jeffjohnson9046: Removed the `@Spring` annotations.  Running this unit test would spin up the Spring Boot container and try to run the flyway database migrations.  Since the test is intended to exercise the `RoundRobinFormSelector`, I created an instance of it explicitly rather than rely on auto-wiring.

Alternately, if we want to make sure the `RoundRobinFormSelector` is created properly as a `@Component` by Spring, then we should make this an integration test.

millerts:
Please remember to add reviewers to pull requests.  I've added the team to this one.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -85,5 +88,21 @@ public ExamController(ExamService examService) {
...
         return ResponseEntity.ok(examApproval);
     }
+
+    @RequestMapping(value = "/{examId}/pause", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<NoContentResponse> pauseExam(@PathVariable final UUID examId) {
+        final Optional<ValidationError> maybeStatusTransitionFailure = examService.pauseExam(examId);
+
+        if (maybeStatusTransitionFailure.isPresent()) {
+            NoContentResponse response = new NoContentResponse(maybeStatusTransitionFailure.get());
+            return new ResponseEntity<>(response, HttpStatus.UNPROCESSABLE_ENTITY);
```
millerts: What does the NoContentResponse buy us?  Rather than simply passing back a list of errors?

service/src/test/java/tds/exam/repositories/impl/ExamSegmentRepositoryImplIntegrationTests.java
millerts: Sounds good.  Hopefully this will not happen much moving forward now that we're all using the same IDE settings.

service/src/test/java/tds/exam/services/impl/RoundRobinFormSelectorImplTest.java
```diff
@@ -18,11 +14,9 @@
...
 public class RoundRobinFormSelectorImplTest {
-    @Autowired
-    private FormSelector formSelector;
+    private FormSelector formSelector = new RoundRobinFormSelectorImpl();
```
millerts: Can you put this in a `@Before` block?

service/src/test/java/tds/exam/web/endpoints/ExamControllerIntegrationTests.java
```diff
@@ -63,4 +65,18 @@ public void shouldReturnNotFoundIfExamCannotBeFound() throws Exception {
 
         verify(mockExamService).findExam(examId);
     }
+
+    @Test
+    public void shouldPauseAnExam() throws Exception {
```
millerts: Can we add a test that has the failure condition as well?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -225,6 +226,38 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
     }
 
     @Override
+    public Optional<ValidationError> pauseExam(UUID examId) {
+        Exam exam = examQueryRepository.getExamById(examId)
+            .orElseThrow(() -> new IllegalArgumentException(String.format("Exam could not be found for id %s", examId)));
```
millerts: Should this be an `IllegalArgument` or `NotFound`?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -225,6 +226,38 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+        // From CommondDLL._IsValidStatusTransition_FN(): a collection of all the statuses that can transition to
+        // "paused".  That is, each of these status values has a nested switch statement that contains the "paused"
+        // status.
+        final List<String> statusesThatCanTransitionToPaused = Arrays.asList(ExamStatusCode.STATUS_PAUSED,
```
millerts: 1. Order doesn't seem to matter here so I'd make this a `Set`
1. It seems like this should either be configured in a static block or be created in the constructor.  Each time this method is called you're creating a new List.
1. I'll be curious if this doesn't become a `Map<String, Set<String>>` once we have more statuses than just pause.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -225,6 +226,38 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
     }
 
     @Override
+    public Optional<ValidationError> pauseExam(UUID examId) {
```
millerts: Should there be a `status_change_reason` somewhere in this code?   If not, why do we have that column in the DB? 

I don't see that column in the `session.testopportunity` table so I'm not sure where it came from.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Instead of creating it in the constructor or a static block you could also create it in a `@Configuration` class and have it injected.  I defer to you if that is better.

http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-autowired-annotation-qualifiers

service/src/main/java/tds/exam/web/endpoints/ExamController.java
jeffjohnson9046: It allows us to return an HTTP response with no body, which is what I wanted to do for the `pauseExam` method.  When it works, I want the `pauseExam` to return a `204 - No Content` and an HTTP response with no body.  When it fails, I want the `pauseExam` to return a `422 - Unprocessable Entity` with an error message stating "Bad transition from X to Y", and I wanted the JSON in the response body to be consistent with our other responses.  There's no `data` to return per se, but I wanted the `errors` collection to be consistent w/other responses we return.

If there's another way we want to handle no content on success and errors on failure, I'm certainly open to other ideas.

millerts: We discussed this a bit offline.  @jeffjohnson9046 and I decided for now to leave the NoContentResponse object instead of returning a List of errors.  That way the json looks like this:

```{json}
{
  errors: [{code: "something", message: "message"}]
}
```

instead of:

```{json}
[{code: "something", message: "message"}]
```

We will discuss in more detail with the group at a later date.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: `status_change_reason` maps to the `session.testopportunity.comment` column.  I thought `status_change_reason` was much more descriptive.  However, I'm pretty sure our Confluence "data dictionary"/database mapping page doesn't say that, so I'll make sure it's updated.

There is no need to provide a `status_change_reason` when pausing an exam, so it's omitted in this code.  The only time a `status_change_reason` is provided (that I've found so far) is when a Proctor changes the status of a student's exam (e.g. the Proctor denies approval of the exam).

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: 1.  Agreed - a `Set` will work here.
2.  Sure thing - I can make that change.
3.  This originally was a `Map<String, List<String>>`.  Since I narrowed the scope of this method to only set the exam's status to `paused`, I removed it and created the `List` shown in this PR.  The large `switch` statement that evaluates status transitions can (and should) easily be turned into a `Map`.

PR #42, opened 2016-12-28
Feature/pause exams in session

jeffjohnson9046:
Add endpoint to pause all `Exam`s that belong to a specified `Session`.  In the event there are no `Exam`s to pause, the endpoint will return successfully.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 
 import tds.common.data.mapping.ResultSetMapperUtility;
@@ -178,6 +179,68 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
         }
 
         return examOptional;
+    }
+
+    @Override
+    public List<Exam> findAllExamsInSessionWithStatus(UUID sessionId, Set<String> statusSet) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("sessionId", UuidAdapter.getBytesFromUUID(sessionId))
+            .addValue("statusSet", statusSet);
+
+        final String SQL =
+            "SELECT \n" +
+                "e.id, \n" +
+                "e.session_id, \n" +
+                "ee.browser_id, \n" +
+                "e.assessment_id, \n" +
+                "e.student_id, \n" +
+                "e.client_name, \n" +
+                "e.environment, \n" +
+                "e.subject, \n" +
+                "e.login_ssid, \n" +
+                "e.student_name, \n" +
+                "e.date_joined, \n" +
+                "e.assessment_key,\n" +
+                "e.assessment_window_id, \n" +
+                "e.assessment_algorithm, \n" +
+                "e.segmented, \n" +
+                "ee.attempts, \n" +
+                "ee.status, \n" +
+                "ee.status_change_date, \n" +
+                "ee.max_items, \n" +
+                "ee.expire_from, \n" +
+                "ee.language_code, \n" +
+                "ee.status_change_reason, \n" +
+                "ee.date_deleted, \n" +
+                "ee.date_changed, \n" +
+                "ee.date_completed, \n" +
+                "ee.date_started,    \n" +
+                "ee.date_scored, \n" +
+                "ee.abnormal_starts, \n" +
+                "ee.waiting_for_segment_approval, \n" +
+                "ee.current_segment_position, \n" +
+                "e.created_at, \n" +
+                "esc.description, \n" +
+                "esc.status, \n" +
+                "esc.stage\n" +
+                "FROM exam.exam e \n" +
+                "JOIN ( \n" +
+                "   SELECT \n" +
+                "       exam_id, \n" +
+                "       MAX(id) AS id \n" +
+                "   FROM \n" +
+                "       exam.exam_event \n" +
+                "   GROUP BY exam_id \n" +
+                ") last_event \n" +
+                "  ON e.id = last_event.exam_id \n" +
+                "JOIN exam.exam_event ee \n" +
+                "  ON last_event.exam_id = ee.exam_id AND \n" +
+                "     last_event.id = ee.id\n" +
+                "JOIN exam.exam_status_codes esc \n" +
+                "  ON esc.status = ee.status \n" +
+                "WHERE e.session_id = :sessionId \n" +
+                "AND ee.status IN (:statusSet)";
+
+        return jdbcTemplate.query(SQL, parameters, new ExamRowMapper());
     }
 
     @Override
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -31,6 +31,40 @@ import static tds.common.data.mapping.ResultSetMapperUtility.mapTimestampToJodaI
 @Repository
 public class ExamQueryRepositoryImpl implements ExamQueryRepository {
     private final NamedParameterJdbcTemplate jdbcTemplate;
+    private final String EXAM_QUERY_COLUMN_LIST = "e.id, \n" +
+        "e.session_id, \n" +
+        "ee.browser_id, \n" +
+        "e.assessment_id, \n" +
+        "e.student_id, \n" +
+        "e.client_name, \n" +
+        "e.environment,\n" +
+        "e.subject, \n" +
+        "e.login_ssid,\n" +
+        "e.student_name,\n" +
+        "e.date_joined, \n" +
+        "e.assessment_key, \n" +
+        "e.assessment_window_id, \n" +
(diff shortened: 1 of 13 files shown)


service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -102,15 +143,15 @@ public void shouldNotReturnLatestExamWithNonNullDeletedDate() {
     @Test
     public void shouldReturnEmptyListOfAbilities() {
         List<Ability> noAbilities = examQueryRepository.findAbilities(UUID.fromString("12345678-d1d2-4c24-805c-0dfdb45a0999"),
-                "otherclient", "ELA", 9999L);
+            "otherclient", "ELA", 9999L);
```
jeffjohnson9046: Only a formatting change.

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -102,15 +143,15 @@ public void shouldNotReturnLatestExamWithNonNullDeletedDate() {
...
         assertThat(noAbilities).isEmpty();
     }
 
     @Test
     public void shouldReturnSingleAbility() {
         UUID examId = UUID.randomUUID();
         List<Ability> oneAbility = examQueryRepository.findAbilities(examId,
-                "clientName", "ELA", 9999L);
+            "clientName", "ELA", 9999L);
```
jeffjohnson9046: Only a formatting change.

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -129,8 +188,8 @@ private void insertExamScoresData() {
 
         final String SQL =
             "INSERT INTO" +
-            "   exam_scores (exam_id, measure_label, value, measure_of, use_for_ability) " +
-            "VALUES(:examId, :measureLabel, :value, :measureOf, :useForAbility)";
+                "   exam_scores (exam_id, measure_label, value, measure_of, use_for_ability) " +
```
jeffjohnson9046: Only a formatting change.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -263,6 +264,26 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
     }
 
     @Override
+    public void pauseAllExamsInSession(UUID sessionId) {
+        List<Exam> examsInSession = examQueryRepository.findAllExamsInSessionWithStatus(sessionId,
+            statusesThatCanTransitionToPaused);
+
+        if (examsInSession.size() == 0) {
```
jtreuting: `isEmpty()` feels cleaner and in some cases is more performant

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -263,6 +264,26 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+
+        ExamStatusCode pausedStatus = new ExamStatusCode(ExamStatusCode.STATUS_PAUSED, ExamStatusStage.INACTIVE);
+        List<Exam> pausedExams = examsInSession.stream()
+            .map(e -> new Exam.Builder().fromExam(e)
+                .withStatus(pausedStatus, org.joda.time.Instant.now())
+                .withStatusChangeReason("paused by session")
+                .build())
+            .collect(Collectors.toList());
+
+        examCommandRepository.update(pausedExams);
```
jtreuting: I know I've seen this pattern of querying, changing some properties on the items, and then saving.  I just had a thought.  It would be more performant to have a command repository method like `updateExamsInSession()` and then it could do a single SQL statement like `INSERT INTO exam_event (....) SELECT col1, col2, ... FROM ...` etc, etc.

Thoughts?

jeffjohnson9046: The more think about this suggestion, the more I like it.  The only thing I can think of is that _maybe_ the table or part of the table might be locked up while the `INSERT... SELECT` runs, which could cause some contention.  However, I'm not an expert on how MySQL does it's locking/transaction management, so this could very well be a non-issue.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -181,6 +182,68 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
+                "   FROM \n" +
+                "       exam.exam_event \n" +
+                "   GROUP BY exam_id \n" +
+                ") last_event \n" +
+                "  ON e.id = last_event.exam_id \n" +
+                "JOIN exam.exam_event ee \n" +
+                "  ON last_event.exam_id = ee.exam_id AND \n" +
+                "     last_event.id = ee.id\n" +
+                "JOIN exam.exam_status_codes esc \n" +
+                "  ON esc.status = ee.status \n" +
```
millerts: This part of the query is similar in multiple methods.  I struggle with this a bit myself; it easier to follow when looking a the query, but we have to modify `n` queries any time we add a column.

service/src/test/java/tds/exam/repositories/impl/ExamCommandRepositoryImplIntegrationTests.java
```diff
@@ -108,4 +110,66 @@ public void shouldUpdateAnExam() {
...
+
+        exams.forEach(e -> examCommandRepository.insert(e));
+
+        List<Exam> examsWithChanges = new ArrayList<>();
+        examsWithChanges.add(new Exam.Builder().fromExam(mockFirstExam)
+            .withStatus(new ExamStatusCode(ExamStatusCode.STATUS_APPROVED, ExamStatusStage.IN_USE), Instant.now().plus(50000))
+            .withStatusChangeReason("unit test")
+            .withAttempts(500)
+            .build());
+        examsWithChanges.add(new Exam.Builder().fromExam(mockSecondExam)
```
millerts: Is there a benefit to updating 3 vs 2?  If not, I suggest doing the minimum

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -120,6 +161,24 @@ public void shouldReturnSingleAbility() {
         assertThat(myAbility.getDateScored()).isLessThan(java.time.Instant.now());
     }
 
+    @Test
+    public void shouldGetAllExamsInASession() {
+        List<Exam> exams = examQueryRepository.findAllExamsInSessionWithStatus(mockSessionId, statusesThatCanTransitionToPaused);
+
+        assertThat(exams).hasSize(3);
+        assertThat(exams.stream().filter(exam -> exam.getStatus().getStatus().equals(ExamStatusCode.STATUS_PENDING)).findAny()).isPresent();
```
millerts: Suggest removing the assertions 169-172 and replacing it with `assertThat(exams).containsOnly(examsInSession))`.  I would either make the examsInSession a instance level variable in the test or have each of these methods call a private method returning the set.

service/src/test/java/tds/exam/web/endpoints/ExamControllerIntegrationTests.java
```diff
@@ -97,4 +100,28 @@ public void shouldReturnAnErrorWhenAttemptingToPauseAnExamInAnInvalidTransitionS
...
+
+        http.perform(put(new URI(String.format("/exam/pause/%s", sessionId)))
+            .contentType(MediaType.APPLICATION_JSON))
+            .andExpect(status().isNoContent());
+
+        verify(mockExamService).pauseAllExamsInSession(sessionId);
+    }
+
+    @Test
+    public void shouldReturn500WhenAnExceptionIsThrownWhilePausingExamsInASession() throws Exception {
```
millerts: This should be removed.  This is testing that the Session application leverages the `ExceptionAdvice` and doesn't rely on any logic specific to pausing a session.

service/src/test/java/tds/exam/web/endpoints/ExamControllerTest.java
```diff
@@ -219,4 +222,24 @@ public void shouldNotPauseAnExam() {
...
+    public void shouldPauseAllExamsInASession() {
+        UUID sessionId = UUID.randomUUID();
+        doNothing().when(mockExamService).pauseAllExamsInSession(sessionId);
+
+        controller.pauseExamsInSession(sessionId);
+
+        verify(mockExamService).pauseAllExamsInSession(sessionId);
+    }
+
+    @Test(expected = SQLException.class)
```
millerts: This test should also be removed.  It only verifies that your mock is called and that it throws.  `shouldPauseAllExamsInASession` is sufficient coverage.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
jeffjohnson9046: I've been on the fence about this too.  I'll make the change so we can try it out.  If we don't like it, we can always change it back.

PR #43, opened 2016-12-29
Feature/start exam - Restarting or resuming an existing exam

burzum619:
This PR includes code surrounding the resuming and restarting of an exam in "/startExam". This code also includes a bit of repository code for interacting with the new exam_page, exam_item, and exam_item_response tables which are closely related and will be utilized more when /getPageContent and /updateResponses is re-implemented.

The main entrypoint of the code is in ExamServiceImpl.startExam() - in the conditonal at line **398**. This corresponds to TestOpportunityServiceImpl line **168**.

Some notes:
- "gracePeriodRestarts" has been renamed to "resumptions" 
- "restarts" has been renamed to "restartsAndResumptions"
- Neither of the above should be confused with "attempts" ("opportunities" in legacy), a restart/resume of the same attempt number.


service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import tds.exam.repositories.ExamPageQueryRepository;
+import tds.exam.repositories.ExamResponseCommandRepository;
+import tds.exam.repositories.ExamResponseQueryRepository;
+import tds.exam.services.ExamItemService;
+
+/**
+ * A service for handling interactions with exam items, pages, and responses.
+ */
+@Service
+public class ExamItemServiceImpl implements ExamItemService {
```
burzum619: This service is called "ExamItemService" although it interacts with exam pages, items, and responses (which are all closely coupled), and not just items specifically. I'm open to a better name, but do feel it makes sense to couple these repositories, at least at the service level.

jtreuting:
I am proud that resumptions ended up being a word.

service/src/test/java/tds/exam/builder/ExamBuilder.java
```diff
@@ -229,4 +234,21 @@ public ExamBuilder withLanguage(String language) {
         this.language = language;
         return this;
     }
+
```
millerts: extra line break

service/src/test/java/tds/exam/builder/ExamBuilder.java
```diff
@@ -229,4 +234,21 @@ public ExamBuilder withLanguage(String language) {
...
+    public ExamBuilder withRestartsAndResumptions(int restartsAndResumptions) {
+        this.restartsAndResumptions = restartsAndResumptions;
+        return this;
+    }
+
+    public ExamBuilder withMaxItems(int maxItems) {
+        this.maxItems = maxItems;
+        return this;
+    }
+
```
millerts: extra line break

service/src/test/java/tds/exam/repositories/impl/ExamPageRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,68 @@
...
+            .withExamId(exam.getId())
+            .withPagePosition(1)
+            .withItemGroupKey("GroupKey1")
+            .build();
+        ExamPage examPage2 = new ExamPage.Builder()
+            .withExamId(exam.getId())
+            .withPagePosition(2)
+            .withItemGroupKey("GroupKey2")
+            .build();
+
```
millerts: Can we add an `assertThat(examPageQueryRepository.findAll(exam.getId()))).isEmpty` before you insert?

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -150,6 +151,91 @@ public void shouldReturnEmptyListOfAbilities() {
...
+        assertThat(maybeLastTimeStudentResponded.get()).isEqualTo(datePageCreated);
+    }
+
+    private void insertTestDataForResponses(Instant datePageCreated, Instant dateLastResponseSubmitted, Instant dateEarlierResponseSubmitted, Exam exam) {
+        MapSqlParameterSource testParams = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(exam.getId()))
+            .addValue("datePageCreated", new Timestamp(datePageCreated.getMillis()))
+            .addValue("dateLastResponseSubmitted", new Timestamp(dateLastResponseSubmitted.getMillis()))
+            .addValue("dateEarlierResponseSubmitted", new Timestamp(dateEarlierResponseSubmitted.getMillis()));
+
+        final String insertPageSQL =
```
millerts: Instead of running straight SQL what if we used the repositories?  My thinking is that would make future refactors or table migrations easier to manage.

millerts: If you agree I think we should follow that pattern throughout.

service/src/test/java/tds/exam/services/impl/ExamItemServiceImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+        ExamPage examPage2 = new ExamPage.Builder()
+            .withId(79)
+            .build();
+        List<ExamPage> examPages = new ArrayList<>();
+        examPages.add(examPage1);
+        examPages.add(examPage2);
+
+        when(mockExamPageQueryRepository.findAll(examId)).thenReturn(examPages);
+        List<ExamPage> retExamPages = examItemService.findAllPages(examId);
+        assertThat(retExamPages).hasSize(2);
```
millerts: Can we change this to `containsExactly(examPage1, examPage2)`?  That won't only check size but ensure the list contains exactly what we expect.

millerts: If it is possible we should always do that instead of just checking for size.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -0,0 +1,74 @@
...
+                "   FROM \n" +
+                "       exam_page_event \n" +
+                "   GROUP BY exam_page_id \n" +
+                ") last_event \n" +
+                "   ON P.id = last_event.exam_page_id \n" +
+                "JOIN exam_page_event PE \n" +
+                "   ON last_event.id = PE.id \n" +
+                "WHERE \n" +
+                "   P.exam_id = :examId AND\n" +
+                "   PE.deleted_at IS NULL";
```
millerts: I think we should order this by page_position

service/src/main/resources/db/migration/V1483051338__exam_create_item_response_page_tables.sql
```diff
@@ -0,0 +1,65 @@
...
+***********************************************************************************************************************/
+
+USE exam;
+
+
+DROP TABLE IF EXISTS exam_page;
+
+CREATE TABLE exam_page (
+  id BIGINT NOT NULL AUTO_INCREMENT,
+  page_position INT(11) NOT NULL,
```
millerts: I think we should just call this `position`.  It being in exam_page implies that it is the position of the page.  I don't feel strongly about this suggestion.  It would just be consistent with `position` that we use in item.

service/src/test/java/tds/exam/builder/ExamBuilder.java
```diff
@@ -78,6 +80,9 @@ public Exam build() {
             .withWaitingForSegmentApproval(waitingForSegmentApproval)
             .withCurrentSegmentPosition(currentSegmentPosition)
             .withCustomAccommodation(customAccommodations)
+            .withResumptions(resumptions)
+            .withRestartsAndResumptions(restartsAndResumptions)
+
```
millerts: extra line break

client/src/main/java/tds/exam/ExamConfiguration.java
```diff
@@ -28,7 +28,7 @@ private ExamConfiguration(Builder builder) {
         this.examRestartWindowMinutes = builder.examRestartWindowMinutes;
         this.requestInterfaceTimeoutMinutes = builder.requestInterfaceTimeoutMinutes;
         this.prefetch = builder.prefetch;
-        this.attempt = builder.attempt;
+        this.restartsAndResumptions = builder.restartsAndResumptions;
```
burzum619: Yep, I believe you're right.

service/src/main/java/tds/exam/models/ExamPage.java
```diff
@@ -0,0 +1,131 @@
+package tds.exam.models;
+
+import org.joda.time.Instant;
+
+import java.util.UUID;
+
+/**
+ * Represents the page of an exam
+ */
+public class ExamPage {
```
burzum619: I think if we wanted to take that kind of approach, an ExamPage could have a collection of ExamItems and each of those ExamItems would have an Optional<ExamResponse>

service/src/main/java/tds/exam/models/ExamResponse.java
```diff
@@ -0,0 +1,81 @@
+package tds.exam.models;
+
+import org.joda.time.Instant;
+
+/**
+ * Represents a response to an {@link tds.exam.models.ExamItem} of an {@link tds.exam.Exam}.
+ */
+public class ExamResponse {
```
burzum619: I'd prefer ExamItemResponse since the class models a row in the exam_item_response table anyway.

service/src/main/java/tds/exam/models/ExamPage.java
millerts: I guess it depends on how things are used.  I don't see much value in querying an response without querying the item.  I don't think we need to do that for everything, but I think we can do it when appropriate.

I made this comment before I saw how everything is used in your flow.

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
burzum619: I think we should discuss this in the parking lot tomorrow. I considered this, but the problem with that is that we're letting the database generate the autoincremented ids and we'd need to add "find" methods to retrieve the newly inserted test records. The alternative is to have our Inserts and "updates" return the object that was inserted/updated with the generated id set by the repository. 

I'm actually okay with either solution, and I'm fine with adding a few simple "getter" methods even if they aren't being used by the code I'm currently working on. It might be useful in the near future, but I know we're also making an effort to shy away from test-specific production code.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -152,6 +154,70 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
+
+        Optional<Instant> maybeLastPausedTime;
+        try {
+            Timestamp lastPausedTime = jdbcTemplate.queryForObject(SQL, parameters, Timestamp.class);
+            maybeLastPausedTime = Optional.of(new Instant(lastPausedTime.getTime()));
+        } catch (EmptyResultDataAccessException e) {
+            maybeLastPausedTime = Optional.empty();
+        }
+
+        return maybeLastPausedTime;
```
jtreuting: This isn't really the last paused time is it?  Seems like it could be the last paused or the last response.  I'd suggest changing the name to `maybeLastStudentActivity`

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -152,6 +154,70 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
     }
 
     @Override
+    public Optional<Instant> getLastStudentActivityInstant(final UUID id) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(id));
+
+        final String SQL =
```
jtreuting: I think it might be useful to have a quick comment that mentioned what each of the 3 queries in the UNION is checking for.

service/src/main/java/tds/exam/repositories/impl/ExamResponseQueryRepositoryImpl.java
```diff
@@ -0,0 +1,51 @@
...
+                "   exam_item I\n" +
+                "JOIN \n" +
+                "   exam_item_response IR \n" +
+                "ON \n" +
+                "   IR.exam_item_id = I.id\n" +
+                "JOIN \n" +
+                "   exam_page P \n" +
+                "ON \n" +
+                "   P.id = I.exam_page_id\n" +
+                "JOIN\n" +
```
jtreuting: space before the \n makes it easier to read (where is the icon for picking gnits?)

service/src/main/java/tds/exam/services/ExamItemService.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+
+import tds.exam.models.ExamPage;
+import tds.exam.models.ExamResponse;
+
+/**
+ * Created by emunoz on 12/28/16.
```
jtreuting: Change to description

service/src/main/java/tds/exam/services/ExamItemService.java
```diff
@@ -0,0 +1,44 @@
...
+     * @param examPage
+     */
+    void insertPages(List<ExamPage> examPage);
+
+    /**
+     * Marks all {@link tds.exam.models.ExamPage}s as "deleted" for the exam.
+     *
+     * @param examId the id of the {@link tds.exam.Exam}
+     */
+    void deletePages(UUID examId);
```
jtreuting: maybe this should be `deleteAllPages(UUID examId)`, just to be very clear of the intent and to be consistent with `findAllPages`?

service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    private final ExamPageCommandRepository examPageCommandRepository;
+    private final ExamPageQueryRepository examPageQueryRepository;
+    private final ExamResponseQueryRepository examResponseQueryRepository;
+
+    @Autowired
+    public ExamItemServiceImpl(ExamPageQueryRepository examPageQueryRepository,
+                               ExamPageCommandRepository examPageCommandRepository,
+                               ExamResponseQueryRepository examResponseQueryRepository,
+                               ExamResponseCommandRepository examResponseCommandRepository)
```
jtreuting: `examResponseCommandRepository` doesn't appear to be used here

service/src/main/java/tds/exam/repositories/impl/ExamPageCommandRepositoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+            KeyHolder keyHolder = new GeneratedKeyHolder();
+            jdbcTemplate.update(examPageSQL, parameterSources, keyHolder);
+            examPage.setId(keyHolder.getKey().longValue());
+
+            update(examPage);
+        });
+    }
+
+    @Override
+    public void delete(final UUID examId) {
```
jtreuting: Seems like maybe this should be `deleteAllPages` or something like that.  I tend to view `delete(UUID)` as deleting the item itself that the UUID represents.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -383,15 +389,53 @@ public void pauseAllExamsInSession(UUID sessionId) {
...
         if (exam.getDateStarted() == null) { // Start a new exam
             // Initialize the segments in the exam and get the testlength.
             Exam initializedExam = initializeExam(exam, assessment);
             /* StudentDLL [5367] and TestOppServiceImpl [167] */
             examConfig = initializeDefaultExamConfiguration(initializedExam, assessment, timeLimitConfiguration);
-        } else { //Restart the most recent exam
-            //TODO: Start existing opportunity [TDS-350]
+        } else { // Restart or resume the most recent exam
+            int resumptions = exam.getResumptions();
+            int restartsAndResumptions = exam.getRestartAndResumptions() + 1; // Increment the restartAndResumption count
```
jtreuting: @millerts Can you use the `+=` when declaring the variable?  And if you can, wouldn't `int restartsAndResumptions += exam.getRestartAndResumptions();` equal `int restartsAndResumptions = exam.getRestartAndResumptions();`?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -383,15 +389,53 @@ public void pauseAllExamsInSession(UUID sessionId) {
...
+            /* TestOpportunityServiceImpl [199] / StudentDlLL [5424]
+             * Session type is always 1 (online) in TDS, so skip first conditional on [200]
+             * No need to update restart count on line [203] because restart count can be derived from event table, so skip [202-203] */
+
+            if (isResumeable) { // Resume exam
+                /* TestOpportunityServiceImpl [215] - Only need to get latest position when resuming, else its 1 */
+                startPosition = examItemService.getExamPosition(exam.getId());
+                /* This increment is done in TestOpportunityServiceImpl [179] */
+                resumptions++;
+            } else if (!isResumeable && assessment.getDeleteUnansweredItems()) { // Restart exam
```
jtreuting: Session type is always 1 right?  If that is the case then I don't think the current application logic would ever get to the part for deleting unanswered items.  The first condition should always be met and since the next 2 are `else if`s and not just `if`s, they should never get hit.

```
                // Conditions starting on StudentDLL.T_StartTestOpportunity_SP @ line 5424
                if (LegacyComparer.isEqual(testSession.getSessionType(), 1)) {
                    testeeResponseDao.updateRestartCount(opportunityInstance.getKey(), restartCount, false);
                } else if (isTimeDiffLessThanDelay) {
                    testeeResponseDao.updateRestartCount(opportunityInstance.getKey(), restartCount, true);
                } else if (LegacyComparer.isEqual(clientTestProperty.getDeleteUnansweredItems(), true)) {
                    removeUnanswered(testOpportunity);
                }
```

If I'm wrong then the `!isResumable` part of your `else if` is not needed since it will only get there if it is not resumable.

jtreuting: maybe `shouldDeleteUnansweredItems()`

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
millerts: I figure we will need some way to query this data.  I'm not sure if it needs to be one to one but I believe we will need to fetch pages, items in the pages, and responses.

I don't think we need to do that right now, but when we do add the fetching logic (getPageContent & updateResponses) we should revisit these tests and remove the inserts.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: What if we just get rid of session type all together?  I think it would clarify things quite a bit.  If we need to keep it maybe we can have an `isOnline()` is whatever other type is supported?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: You cannot.  Not sure what I was thinking with my comment.  Please ignore.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Another note, the `!isResumeable` can be removed because if it is true it'll never reach this if/esle block due to line 416.

millerts: I also agree with @jtreuting concerning never deleting items.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -383,15 +389,53 @@ public void pauseAllExamsInSession(UUID sessionId) {
...
+                .fromExam(exam)
+                .withStatus(new ExamStatusCode(ExamStatusCode.STATUS_STARTED, ExamStatusStage.IN_PROGRESS), now)
+                .withDateChanged(now)
+                .withResumptions(resumptions)
+                .withRestartsAndResumptions(restartsAndResumptions)
+                .withDateStarted(now)
+                .build();
+
+            examCommandRepository.update(restartedExam);
+            /* Skip restart increment on [209] because we are already incrementing earlier in this method */
```
millerts: I don't think this line is correct.  Line 209 in `TestOpportunityServiceImpl` is `removeUnanswered(testOpportunity);`.  No incrementing.  I think you mean line 213.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -383,15 +389,53 @@ public void pauseAllExamsInSession(UUID sessionId) {
...
+            /* [212] No need to call updateUnifnishedResponsePages since we no longer need to keep count of "opportunityrestart" */
```
millerts: Isn't this line 215?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -430,13 +474,30 @@ private static ExamConfiguration initializeDefaultExamConfiguration(Exam exam, A
             .withPrefetch(assessment.getPrefetch())
             .withValidateCompleteness(assessment.isValidateCompleteness())
             .withRequestInterfaceTimeout(timeLimitConfiguration.getRequestInterfaceTimeoutMinutes())
-            .withAttempt(0)
+            .withRestartsAndResumptions(0)
             .withStartPosition(1)
```
millerts: Is `startPosition` the same as `resumeItemPosition` in the legacy code?  If so, why is it always 1?  There is a call to a DB function to get the `resumeItemPosition` for restarts.  It may be easier to explain in our in person overview, but I'd like it called out in a comment somewhere if we don't need to replicate that functionality.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jtreuting: Yeah the line numbers felt off to me as well, just in general.

service/src/main/java/tds/exam/repositories/impl/ExamPageCommandRepositoryImpl.java
jeffjohnson9046: I could get behind `deleteAll` as a name; I think because this method is a member of `ExamPageCommandRepository` the page part is implied.  But I agree - there should be some indication that many records could be affected.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -0,0 +1,120 @@
...
+
+        public ExamItem build() {
+            return new ExamItem(this);
+        }
+    }
+
+    /**
+     * @return The id of the exam item
+     */
+    public long getId() {
```
millerts: I think we should leverage [@JacksonIgnore](http://fasterxml.github.io/jackson-annotations/javadoc/2.5/com/fasterxml/jackson/annotation/JsonIgnore.html) for these ID's we don't want to expose to the clients.  I should've done the same thing in the exam accommodations.

service/src/main/java/tds/exam/services/ExamItemService.java
burzum619: Woops! nice catch

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: This block of code is only hit when the exam is initialized for the first time (neither a resume nor restart), so the startPosition is always 1 - no reason to hit the database. Same with Restarts and attempts. I can add  comments to make this a bit more clear.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Nope. [212] is where updateUnfinishedResponsePages() is being called (at least in the code I'm looking at). That's the portion I skipped.

PR #44, opened 2017-01-04
Remove session type because it is no longer needed.

millerts:


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -652,8 +650,6 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
 
         examAccommodationService.initializeAccommodationsOnPreviousExam(previousExam, assessment, 0, restoreAccommodations, guestAccommodations);
 
-        //TODO - Custom accommodations?
```
millerts: This is a left over TODO.  There is some legacy code that sets custom accommodations bit flag depending if the accommodations are custom and not the default.  However, I didn't see it being used so left it off.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -664,16 +660,12 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
         }
 
         //Port of Student.DLL lines 5531-5551
-        //If either session type is null or if they don't match an error is returned
         Optional<Session> maybePreviousSession = sessionService.findSessionById(previousExam.getSessionId());
         if (!maybePreviousSession.isPresent()) {
-            return Optional.of(new ValidationError(ValidationErrorCode.SESSION_TYPE_MISMATCH, "current session type and previous session type don't match"));
+            return Optional.of(new ValidationError(ValidationErrorCode.PREVIOUS_SESSION_NOT_FOUND, "Exam's previous session could not be found"));
```
millerts: Not sure if this should be a validation error or throw an IllegalStateException.  It should really never happen.

The earlier error and text were incorrect

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -193,47 +193,6 @@ public void shouldReturnValidationErrorWhenSessionIsNoOpen() {
         assertThat(response.getErrors()[0].getCode()).isEqualTo(ValidationErrorCode.SESSION_NOT_OPEN);
     }
 
-    @Test
```
millerts: This unit test was only for session type mismatch error.  Since that can no longer happen the test is no longer needed.

client/src/main/java/tds/exam/error/ValidationErrorCode.java
```diff
@@ -5,8 +5,8 @@
     public static final String MAX_OPPORTUNITY_EXCEEDED = "maxOpportunityPassed";
     public static final String NOT_ENOUGH_DAYS_PASSED = "notEnoughDaysPassed";
     public static final String SIMULATION_ENVIRONMENT_REQUIRED = "simulationEnvironmentRequired";
-    public static final String SESSION_TYPE_MISMATCH = "sessionTypeMismatch";
     public static final String CURRENT_EXAM_OPEN = "examAlreadyOpen";
+    public static final String PREVIOUS_SESSION_NOT_FOUND = "previousSessionNotFound";
```
millerts: Look [below](https://github.com/SmarterApp/TDS_ExamService/pull/44/files#diff-48bacc108f86a2823be9083d6fd1c3d7R665) to see how this is used.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -517,15 +515,15 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
         List<AssessmentWindow> assessmentWindows = configService.findAssessmentWindows(
             openExamRequest.getClientName(),
             assessment.getAssessmentId(),
-            session.getType(),
+
```
jtreuting: extra line break

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jtreuting: There is code that sets a "is custom accommodations" type flag though, right?  I remember seeing it in a code review and think it is stored in the DB.  I'm just assuming it is happening somewhere else and not here.

millerts: Yes, there is code that sets it in legacy.  However, I cannot see how it is used.  It is queried in Proctor but I don't see that it is used so I left it off.

millerts: Discussed offline.  It is used in Proctor for display purposes.  I will add that into the exam service in a follow up PR.

PR #45, opened 2017-01-05
Add missing status codes supported in the legacy system.

millerts:
This is a smallish change.  I've added the missing status codes that are used by the legacy system.  They are currently unused in our code, and if we find out that they are never used we can remove them later.  This is for the legacy conversion code I'm writing.

Also renamed `status` to `code` in the `ExamStatusCode` class.

### DIFF
--- a/client/src/main/java/tds/exam/ExamApproval.java
+++ b/client/src/main/java/tds/exam/ExamApproval.java
@@ -13,7 +13,7 @@ public class ExamApproval {
     public ExamApproval(UUID examId, ExamStatusCode examStatusCode, String statusChangeReason) {
         this.examId = examId;
         this.statusChangeReason = statusChangeReason;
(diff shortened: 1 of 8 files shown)


client/src/main/java/tds/exam/ExamApproval.java
```diff
@@ -13,7 +13,7 @@
     public ExamApproval(UUID examId, ExamStatusCode examStatusCode, String statusChangeReason) {
         this.examId = examId;
         this.statusChangeReason = statusChangeReason;
-        this.examApprovalStatus = ExamApprovalStatus.fromExamStatus(examStatusCode.getStatus());
+        this.examApprovalStatus = ExamApprovalStatus.fromExamStatus(examStatusCode.getCode());
```
millerts: It seems weird to me to have `exam.getStatus().getStatus()` to get the String value of the status.  If we don't like this maybe we should rename the getter on the exam?

PR #47, opened 2017-01-13
Feature/field test select items - Field Test Item Group Selection

burzum619:
This PR includes logic for selecting field test item group for a exam segment. 

The entrypoint for this code is the FieldTestServiceImpl.selectItemGroups() method, which is called by ExamSegmentServiceImpl.initializeExamSegments() [line 113 currently].

The selector implementation keeps a (thread safe and ordered) cached list for each segment, which is ordered by the number of group key usages (See EqualDistributionFieldTestItemGroupSelector.java). After updates to the counts are made, the list of group keys is resorted and the "least used" groups are placed at the front of the array list. This cache is flushed out daily, but ultimately should result in an approximately equal distribution of all field test item groups.

The purpose of the FieldTestServiceImpl.selectItemGroups() is to not only select the item groups that will be assigned to an exam, but also to calculate the position for the field test item groups in the exam (based on data in the itembank.tblsetofadminsubjects table). Much of the data that was being stored in the legacy ft_opportunityitem table has been scrapped, as it existed for debug and auditing purpose and was never read from again in the application.

Some minor updates to TDS_AssessmentService will accompany this PR.

### DIFF
--- a/service/src/main/java/tds/exam/models/FieldTestItemGroup.java
+++ b/service/src/main/java/tds/exam/models/FieldTestItemGroup.java
@@ -1,0 +1,297 @@
+package tds.exam.models;
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Represents a field test item group
+ */
+public class FieldTestItemGroup {
+    private long id;
+    private String segmentKey;
+    private UUID examId;
+    private int position;
+    private int numItems;
+    private String groupId;
+    private String groupKey;
+    private String blockId;
+    private String segmentId;
+    private UUID sessionId;
+    private String languageCode;
+    private Instant createdAt;
+    private Instant deletedAt;
+    private Integer positionAdministered;
+    private Instant administeredAt;
+
+    private FieldTestItemGroup() {}
+
+    public FieldTestItemGroup(Builder builder) {
+        this.segmentKey = builder.segmentKey;
+        this.examId = builder.examId;
+        this.position = builder.position;
+        this.numItems = builder.numItems;
+        this.groupId = builder.groupId;
+        this.groupKey = builder.groupKey;
+        this.blockId = builder.blockId;
+        this.segmentId = builder.segmentId;
+        this.sessionId = builder.sessionId;
+        this.languageCode = builder.languageCode;
+        this.administeredAt = builder.administeredAt;
+        this.createdAt = builder.createdAt;
+        this.positionAdministered = builder.positionAdministered;
+        this.deletedAt = builder.deletedAt;
+    }
+
+    /**
+     * @return The id of the {@link tds.exam.models.FieldTestItemGroup}
+     */
+    public long getId() {
+        return id;
+    }
+
+    public void setId(long id) {
+        this.id = id;
+    }
+
+    /**
+     * @return The group id of the {@link tds.exam.models.FieldTestItemGroup}
+     */
+    public String getGroupId() {
+        return groupId;
+    }
+
+    /**
+     * @return The key of the {@link tds.assessment.Segment} the {@link tds.exam.models.FieldTestItemGroup} is a part of
+     */
+    public String getSegmentKey() {
+        return segmentKey;
+    }
+
+    /**
+     * @return The id of the {@link tds.assessment.Segment} the {@link tds.exam.models.FieldTestItemGroup} is a part of
+     */
+    public String getSegmentId() {
+        return segmentId;
+    }
+
+    /**
+     * @return The id of the {@link tds.session.Session} the {@link tds.exam.models.FieldTestItemGroup} is a part of
+     */
+    public UUID getSessionId() {
+        return sessionId;
+    }
+
+    /**
+     * @return The id of the {@link tds.exam.Exam} this {@link tds.exam.models.FieldTestItemGroup} is a part of
+     */
+    public UUID getExamId() {
+        return examId;
+    }
+
+    /**
+     * @return The position of the {@link tds.exam.models.FieldTestItemGroup} in the {@link tds.exam.Exam}
+     */
+    public int getPosition() {
+        return position;
+    }
+
+    /**
+     * @return The number of field test {@link tds.assessment.Item}s in the {@link tds.exam.models.FieldTestItemGroup}
+     */
+    public int getNumItems() {
+        return numItems;
+    }
+
+    /**
+     * @return The group key of the {@link tds.exam.models.FieldTestItemGroup}. Usually in the form of "groupid_blockid".
+     */
+    public String getGroupKey() {
+        return groupKey;
+    }
+
+    /**
+     * @return The block id of the {@link tds.exam.models.FieldTestItemGroup}
+     */
+    public String getBlockId() {
+        return blockId;
+    }
+
+    /**
+     * @return The language code of the {@link tds.exam.Exam}
+     */
+    public String getLanguageCode() {
+        return languageCode;
+    }
+
+    /**
+     * @return The {@link java.time.Instant} the {@link tds.exam.models.FieldTestItemGroup} was assigned at
+     */
+    public Instant getCreatedAt() {
+        return createdAt;
+    }
+
+    /**
+     * @return The {@link java.time.Instant}
+     */
+    public Instant getDeletedAt() {
+        return deletedAt;
+    }
+
+    /**
+     * @return A flag that determines whether the {@link tds.exam.models.FieldTestItemGroup} is deleted
+     */
+    public boolean isDeleted() {
+        return deletedAt == null;
+    }
+
+    /**
+     * @return The position in the exam the {@link tds.exam.models.FieldTestItemGroup} was actually administered at
+     */
+    public Integer getPositionAdministered() {
+        return positionAdministered;
+    }
+
+    /**
+     * @return The {@link java.time.Instant} the item was administered at
+     */
+    public Instant getAdministeredAt() {
+        return administeredAt;
+    }
+
+    public static class Builder {
+        private long id;
+        private String segmentKey;
+        private UUID examId;
+        private int position;
+        private int numItems;
+        private String groupId;
+        private String groupKey;
+        private String blockId;
+        private String segmentId;
+        private UUID sessionId;
+        private String languageCode;
+        private Instant createdAt;
+        private Instant deletedAt;
+        private Integer positionAdministered;
+        private Instant administeredAt;
+
+        public Builder withId(int id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder withSegmentKey(String segmentKey) {
+            this.segmentKey = segmentKey;
+            return this;
+        }
+
+        public Builder withExamId(UUID examId) {
+            this.examId = examId;
+            return this;
+        }
+
+        public Builder withPosition(int position) {
+            this.position = position;
+            return this;
+        }
+
+        public Builder withNumItems(int numItems) {
+            this.numItems = numItems;
+            return this;
+        }
+
+        public Builder withGroupId(String groupId) {
+            this.groupId = groupId;
+            return this;
+        }
+
+        public Builder withGroupKey(String groupKey) {
+            this.groupKey = groupKey;
+            return this;
+        }
+
+        public Builder withBlockId(String blockId) {
+            this.blockId = blockId;
+            return this;
+        }
+
+        public Builder withSegmentId(String segmentId) {
+            this.segmentId = segmentId;
+            return this;
+        }
+
+        public Builder withSessionId(UUID sessionId) {
+            this.sessionId = sessionId;
+            return this;
+        }
+
+        public Builder withLanguageCode(String languageCode) {
+            this.languageCode = languageCode;
+            return this;
+        }
+
+        public Builder withCreatedAt(Instant createdAt) {
+            this.createdAt = createdAt;
+            return this;
+        }
+
+        public Builder withDeletedAt(Instant deletedAt) {
+            this.deletedAt = deletedAt;
+            return this;
+        }
+
+        public Builder withPositionAdministered(Integer positionAdministered) {
+            this.positionAdministered = positionAdministered;
+            return this;
+        }
+
+        public Builder withAdministeredAt(Instant administeredAt) {
+            this.administeredAt = administeredAt;
+            return this;
+        }
+
+        public Builder fromFieldTestItemGroup(final FieldTestItemGroup fieldTestItemGroup) {
+            this.segmentKey = fieldTestItemGroup.segmentKey;
+            this.examId = fieldTestItemGroup.examId;
+            this.position = fieldTestItemGroup.position;
+            this.numItems = fieldTestItemGroup.numItems;
+            this.groupId = fieldTestItemGroup.groupId;
+            this.groupKey = fieldTestItemGroup.groupKey;
+            this.blockId = fieldTestItemGroup.blockId;
+            this.segmentId = fieldTestItemGroup.segmentId;
+            this.sessionId = fieldTestItemGroup.sessionId;
+            this.languageCode = fieldTestItemGroup.languageCode;
+            this.administeredAt = fieldTestItemGroup.administeredAt;
+            this.createdAt = fieldTestItemGroup.createdAt;
+            this.positionAdministered = fieldTestItemGroup.positionAdministered;
+            this.deletedAt = fieldTestItemGroup.deletedAt;
+            return this;
+        }
+
+        public FieldTestItemGroup build() {
+            return new FieldTestItemGroup(this);
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        FieldTestItemGroup that = (FieldTestItemGroup) o;
+
+        if (!examId.equals(that.examId)) return false;
+        if (!languageCode.equals(that.languageCode)) return false;
+        if (!segmentKey.equals(that.segmentKey)) return false;
+        return groupKey.equals(that.groupKey);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = examId.hashCode();
+        result = 31 * result + languageCode.hashCode();
+        result = 31 * result + segmentKey.hashCode();
+        result = 31 * result + groupKey.hashCode();
+        return result;
+    }
+}
--- a/service/src/main/java/tds/exam/models/FieldTestItemGroup.java
+++ b/service/src/main/java/tds/exam/models/FieldTestItemGroup.java
@@ -11,7 +11,7 @@ public class FieldTestItemGroup {
     private String segmentKey;
     private UUID examId;
     private int position;
-    private int numItems;
+    private int itemCount;
     private String groupId;
     private String groupKey;
     private String blockId;
@@ -23,13 +23,14 @@ public class FieldTestItemGroup {
     private Integer positionAdministered;
     private Instant administeredAt;
 
+    //Exists for frameworks
     private FieldTestItemGroup() {}
 
     public FieldTestItemGroup(Builder builder) {
         this.segmentKey = builder.segmentKey;
         this.examId = builder.examId;
         this.position = builder.position;
-        this.numItems = builder.numItems;
+        this.itemCount = builder.itemCount;
         this.groupId = builder.groupId;
         this.groupKey = builder.groupKey;
         this.blockId = builder.blockId;
@@ -98,8 +99,8 @@ public class FieldTestItemGroup {
     /**
      * @return The number of field test {@link tds.assessment.Item}s in the {@link tds.exam.models.FieldTestItemGroup}
      */
-    public int getNumItems() {
-        return numItems;
+    public int getItemCount() {
+        return itemCount;
     }
 
     /**
@@ -141,7 +142,7 @@ public class FieldTestItemGroup {
      * @return A flag that determines whether the {@link tds.exam.models.FieldTestItemGroup} is deleted
      */
     public boolean isDeleted() {
-        return deletedAt == null;
+        return deletedAt != null;
     }
 
     /**
@@ -163,7 +164,7 @@ public class FieldTestItemGroup {
         private String segmentKey;
         private UUID examId;
         private int position;
-        private int numItems;
+        private int itemCount;
         private String groupId;
         private String groupKey;
         private String blockId;
@@ -195,8 +196,8 @@ public class FieldTestItemGroup {
             return this;
         }
 
-        public Builder withNumItems(int numItems) {
-            this.numItems = numItems;
+        public Builder withItemCount(int itemCount) {
+            this.itemCount = itemCount;
             return this;
         }
 
@@ -251,10 +252,11 @@ public class FieldTestItemGroup {
         }
 
         public Builder fromFieldTestItemGroup(final FieldTestItemGroup fieldTestItemGroup) {
+            this.id = fieldTestItemGroup.id;
             this.segmentKey = fieldTestItemGroup.segmentKey;
             this.examId = fieldTestItemGroup.examId;
             this.position = fieldTestItemGroup.position;
-            this.numItems = fieldTestItemGroup.numItems;
+            this.itemCount = fieldTestItemGroup.itemCount;
             this.groupId = fieldTestItemGroup.groupId;
             this.groupKey = fieldTestItemGroup.groupKey;
             this.blockId = fieldTestItemGroup.blockId;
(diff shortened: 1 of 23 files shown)


service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+import tds.exam.Exam;
+import tds.exam.models.FieldTestItemGroup;
+import tds.exam.services.FieldTestItemGroupSelector;
+import tds.exam.services.ItemPoolService;
+
+@Component
+public class EqualDistributionFieldTestItemGroupSelector implements FieldTestItemGroupSelector {
+    private final ItemPoolService itemPoolService;
+    // Keeps track of the field test item groups and their usages
+    private Cache<String, List<FieldTestItemGroupCounter>> fieldTestItemGroupCounterCache;
```
burzum619: The Key here is the segmentKey. I can make this more clear in code comments if necessary.

millerts: I think adding something to your comment on line 28 would be great

service/src/main/java/tds/exam/models/FieldTestItemGroup.java
```diff
@@ -0,0 +1,297 @@
...
+    private String blockId;
+    private String segmentId;
+    private UUID sessionId;
+    private String languageCode;
+    private Instant createdAt;
+    private Instant deletedAt;
+    private Integer positionAdministered;
+    private Instant administeredAt;
+
+    private FieldTestItemGroup() {}
```
millerts: I've been adding a comment to these empty constructors that says `//Exists for frameworks`

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+            .collect(Collectors.toSet());
+        List<FieldTestItemGroup> selectedFieldTestItemGroups = fieldTestItemGroupSelector.selectItemGroupsLeastUsed(exam, assignedGroupIds, assessment,
+            segmentKey, minItems);
+
+        /* [3240-3242] endPos variable is never used again, no need to increment it - only read from in debug mode */
+        /* [3244] no need to select an unused groupkey - we know our FieldTestGroupItems have unique groupkeys. */
+        /* [3244-3246] Since we have list of unused items returned by selectItemgroupsRoundRobin(), no need to check that groupkey exists */
+        List<FieldTestItemGroup> selectedItemGroups = new ArrayList<>();
+
+        /* This loop begins at [3246] */
```
millerts: I think this comment should be expanded explaining why you can just iterate over the selectedFieldTestItemGroups instead of the way the while loop behaves in the legacy code.  I admit the legacy code is a bit hard to follow as currently written.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+                "between the ending and starting field test positions. (%s - %s)", maxItems, endPosition, startPosition));
+        }
+
+        /* [3131-3137] Note tht if endPos - startPos < numIntervals, the integer division below results in 0, which
+        *  results in  a division-by-zero at line [3320] */
+        int numIntervals = maxItems;
+        int intervalSize = (endPosition - startPosition) / numIntervals;
+        int intervalIndex = startPosition; // keeps track of position in exam to administer ft item
+
+        /* [3212- 3224] Skip Cohort code - the loader script hardcodes "ratio" and "cohortIndex" as 1, targetcount = ftMaxItems */
```
jtreuting: not sure we can skip this.  on line 3293 it updates groupcount based on some SQL (that I haven't had a chance to run yet).  then on line 3300 it checks for when the `groupcount >0 and itemcount < targetcount` to see if anything meets that criteria and if it does, it doesn't do the rest of the logic in the loop.  so it seems like some way of controlling what items can be included.  maybe so that mutiple FT items from the same "cohort" are not included for one student.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+        // Used to keep track of all the group ids that have already been assigned.
+        Set<String> assignedGroupIds = previouslyAssignedFieldTestItemGroups.stream()
+            .map(fieldTestItemGroup -> fieldTestItemGroup.getGroupId())
+            .collect(Collectors.toSet());
+        List<FieldTestItemGroup> selectedFieldTestItemGroups = fieldTestItemGroupSelector.selectItemGroupsLeastUsed(exam, assignedGroupIds, assessment,
```
jtreuting: `selectLeastUsedItemGroups` reads better to me, don't feel too strongly about it

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
         This code covers legacy StudentDLL._FT_SelectItemgroups_SP [line 3033] and is called by _InitializeTestSegments_SP [4704]
      */
     @Override
-    public int selectItemGroups(Exam exam, Assessment assessment, String segmentKey, String language) {
-        //TODO: Implement. Adding stub to prevent compilation errors
-        return 0;
+    public int selectItemGroups(final Exam exam, final Assessment assessment, final String segmentKey) {
+        Segment currentSegment = assessment.getSegment(segmentKey);
```
burzum619: We also need the ItemConstraints (that are a property of Assessment) but I can pass in a List<ItemConstraint> and Segment instead.

service/src/main/java/tds/exam/models/FieldTestItemGroup.java
```diff
@@ -0,0 +1,297 @@
...
+     */
+    public Instant getDeletedAt() {
+        return deletedAt;
+    }
+
+    /**
+     * @return A flag that determines whether the {@link tds.exam.models.FieldTestItemGroup} is deleted
+     */
+    public boolean isDeleted() {
+        return deletedAt == null;
```
millerts: Should this be `!=null`?

service/src/main/java/tds/exam/repositories/impl/FieldTestItemGroupQueryRepositoryImpl.java
```diff
@@ -0,0 +1,86 @@
...
+        this.jdbcTemplate = queryJdbcTemplate;
+    }
+
+    @Override
+    public List<FieldTestItemGroup> find(UUID examId, String segmentKey) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examId))
+            .addValue("segmentKey", segmentKey);
+
+        final String SQL =
+            "SELECT \n" +
```
millerts: We should be pulling back the `F.id` field.  Otherwise how will one update the fieldTestItemGroup?

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+            .build();
+    }
+
+    @Override
+    public List<FieldTestItemGroup> selectItemGroupsLeastUsed(final Exam exam, final Set<String> assignedGroupIds,
+                                                              final Assessment assessment, final String segmentKey, final int numItems) {
+        int ftItemCount = 0;
+        Segment currentSegment = assessment.getSegment(segmentKey);
+        // Fetch every eligible item based on item constraints, item properties, and user accommodations
+        Set<Item> fieldTestItems = itemPoolService.getItemPool(exam.getId(), assessment.getItemConstraints(), currentSegment.getItems(exam.getLanguageCode()), true);
```
millerts: I know this isn't part of the review but what if instead of having the boolean flag we had two methods on the interface:

```
getItemPool()
getFieldTestItemPool()
```

The second implementation could call the `getItemPool()` method then filter it based on the field test setting.  It'd be easier to understand the intent.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: The query at line 3293 updates the group count - which appears to just be the number of items in the item group (including both FT and non-FT items) and for the current language (hence the join on tblitemprops) and segment. We know that

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
jtreuting: oops, i missed that one

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+            List<FieldTestItemGroupCounter> initGroupCounters = fieldTestItemGroupsMap.keySet().stream()
+                .map(groupKey -> new FieldTestItemGroupCounter(groupKey))
+                .collect(Collectors.toList());
+
+            fieldTestItemGroupCounterCache.put(segmentKey, Collections.synchronizedList(initGroupCounters));
+        }
+
+        List<FieldTestItemGroup> itemGroupsWithItemCounts = new ArrayList<>();
+        // Get the list (sorted by number of group key occurrences) and starting at the top, add as many as we need to the list
+        List<FieldTestItemGroupCounter> fieldTestItemGroupCounters = fieldTestItemGroupCounterCache.getIfPresent(segmentKey);
```
millerts: It seems like this should be before line 65.  Your conditional could then check if this is null and add it to the cache and assign the list to `fieldTestItemGroupCounters`.  To me it would reduce some of the complexity in reading this and you wouldn't have to access the cache for something you just added (line 70).

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+    private final ItemPoolService itemPoolService;
+    // Keeps track of the field test item groups and their usages
+    private Cache<String, List<FieldTestItemGroupCounter>> fieldTestItemGroupCounterCache;
+
+    @Autowired
+    public EqualDistributionFieldTestItemGroupSelector(ItemPoolService itemPoolService) {
+        this.itemPoolService = itemPoolService;
+
+        fieldTestItemGroupCounterCache = CacheBuilder.newBuilder()
+            .expireAfterAccess(1, TimeUnit.DAYS)
```
jtreuting: In order to get a more equal distribution I think we let this cache for much longer and maybe configurable as a property.  this size of the cache won't grow past a fixed size since it is keeping a count and not adding entries each time it is used.

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+        Segment currentSegment = assessment.getSegment(segmentKey);
+        // Fetch every eligible item based on item constraints, item properties, and user accommodations
+        Set<Item> fieldTestItems = itemPoolService.getItemPool(exam.getId(), assessment.getItemConstraints(), currentSegment.getItems(exam.getLanguageCode()), true);
+        /* In StudentDLL.FT_Prioritize2012_SP() [3544] - The legacy code creates temporary tables and groups data by groupid, blockid, groupkey.
+            We can just worry about grouping by groupkey since groupKey appears to be the same as "<group-id>_<block-id>".
+
+            The deletion at line [3186] simply removes item groups already found to be assigned to a student
+                - the filter below will take care of this.
+         */
+        Map<String, List<FieldTestItemGroup>> fieldTestItemGroupsMap = fieldTestItems.stream()
```
millerts: Can we name this something better than `...Map`?

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+
+        List<FieldTestItemGroup> itemGroupsWithItemCounts = new ArrayList<>();
+        // Get the list (sorted by number of group key occurrences) and starting at the top, add as many as we need to the list
+        List<FieldTestItemGroupCounter> fieldTestItemGroupCounters = fieldTestItemGroupCounterCache.getIfPresent(segmentKey);
+        for (FieldTestItemGroupCounter groupCounter : fieldTestItemGroupCounters) {
+            // Break out of this loop if we've selected the # of items we needed to select
+            if (ftItemCount >= numItems) {
+                break;
+            }
+
```
millerts: Can we add `String groupKey = groupCounter.getGroupKey();` here and just use it rather than doing the get everywhere? It would read easier to me.

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+            // Check that the groupKey is one of the ones this examinee is eligible for
+            if (fieldTestItemGroupsMap.containsKey(groupCounter.getGroupKey())) {
+                List<FieldTestItemGroup> items = fieldTestItemGroupsMap.get(groupCounter.getGroupKey());
```
millerts: Can we call this itemGroups?  items makes me think items in a group however that isn't what this.

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+                fieldTestItemGroupsMap.remove(groupCounter.getGroupKey());
+            }
+        }
+
+        // Check if there are any groups that may need to be added to cache (initialized)
+        if (!fieldTestItemGroupsMap.isEmpty()) {
+            cacheInitializedCounters(segmentKey, fieldTestItemGroupsMap);
+        }
+
+        Collections.sort(fieldTestItemGroupCounters);
```
millerts: Can we add a comment here explaining why we're sorting the collection?

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+
+        for (String groupKey : fieldTestItemGroupsMap.keySet()) {
+            if (!cachedGroupKeys.contains(groupKey)) {
+                groupCounters.add(new FieldTestItemGroupCounter(groupKey));
+            }
+        }
+    }
+
+    private class FieldTestItemGroupCounter implements Comparable<FieldTestItemGroupCounter> {
+        private String groupKey;
```
millerts: Can you add final here.

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
```diff
@@ -0,0 +1,147 @@
...
+    public EqualDistributionFieldTestItemGroupSelector(ItemPoolService itemPoolService) {
+        this.itemPoolService = itemPoolService;
+
+        fieldTestItemGroupCounterCache = CacheBuilder.newBuilder()
+            .expireAfterAccess(1, TimeUnit.DAYS)
+            .build();
+    }
+
+    @Override
+    public List<FieldTestItemGroup> selectItemGroupsLeastUsed(final Exam exam, final Set<String> assignedGroupIds,
```
millerts: I feel this should just pass in the Segment rather than the key.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
millerts: I'd rather just leave the Assessment than do that

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+              Skip [3276-3285] - debug code */
+            int numItems = fieldTestItemGroup.getNumItems();
+            /* [3307] */
+            if (numItems > 0 && ftItemCount + numItems <= maxItems) {
+                /* [3308 - 3314] */
+                int thisIntSize = (intervalSize < 0 || numItems == 1) ?
+                    1 : intervalSize * (numItems - 1);
+
+                /* [3318] Randomly select an item position for this ft item group */
+                Random rng = new Random();
```
millerts: Can this be moved out of the loop?  Do you need to create a new one each time?

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+                int nextPosition = (rng.nextInt(1000) % thisIntSize) + intervalIndex;
+
+                /* [3339] Mark item group as "used" */
+                assignedGroupIds.add(fieldTestItemGroup.getGroupId());
```
millerts: You're adding to this collection but I never see the collection accessed again in this method.

service/src/main/resources/db/migration/V1483732508__exam_create_table_field_test_item_group.sql
```diff
@@ -0,0 +1,43 @@
+/***********************************************************************************************************************
+  File: V1483732508__exam_create_table_field_test_item_group.sql
+
+  Desc: Creates the field_test_item_group table
```
millerts: Can you give a note what the table is used for?

service/src/main/java/tds/exam/models/FieldTestItemGroup.java
```diff
@@ -0,0 +1,297 @@
...
+            return this;
+        }
+
+        public Builder withAdministeredAt(Instant administeredAt) {
+            this.administeredAt = administeredAt;
+            return this;
+        }
+
+        public Builder fromFieldTestItemGroup(final FieldTestItemGroup fieldTestItemGroup) {
+            this.segmentKey = fieldTestItemGroup.segmentKey;
```
millerts: Doesn't `id` need to be in here so when you do an update it works as expected?

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
millerts: We probably just need to talk about this suggestion and @jtreuting's earlier comment.

service/src/main/java/tds/exam/models/FieldTestItemGroup.java
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * Represents a field test item group
+ */
+public class FieldTestItemGroup {
+    private long id;
+    private String segmentKey;
+    private UUID examId;
+    private int position;
+    private int numItems;
```
millerts: Can we just spell out his word?

service/src/main/resources/db/migration/V1483732508__exam_create_table_field_test_item_group.sql
```diff
@@ -0,0 +1,43 @@
...
+USE exam;
+
+DROP TABLE  IF EXISTS field_test_item_group_event;
+DROP TABLE IF EXISTS field_test_item_group;
+
+CREATE TABLE field_test_item_group (
+  id BIGINT NOT NULL AUTO_INCREMENT,
+  exam_id VARBINARY(16) NOT NULL,
+  position INT(11) NOT NULL,
+  num_items INT(11) DEFAULT NULL,
```
millerts: Can we change this to `item_count` or `total_item_count`?

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+        /* This loop begins at [3246] */
+        for (FieldTestItemGroup fieldTestItemGroup : selectedFieldTestItemGroups) {
+            /* Skip [3248-3274] - This code is just selecting a single item group that is unassigned and not frequently used
+              (as sorted by FT_Prioritize_2012())
+              Skip [3276-3285] - debug code */
+            int numItems = fieldTestItemGroup.getNumItems();
+            /* [3307] */
+            if (numItems > 0 && ftItemCount + numItems <= maxItems) {
+                /* [3308 - 3314] */
+                int thisIntSize = (intervalSize < 0 || numItems == 1) ?
```
millerts: I believe @jeffjohnson9046 prefers the following style:

```
(intervalSize< 0 || numItems == 1)
? 1
: intervalSize * (numItems - 1);
```

millerts: Since it is on multiple lines.

service/src/main/java/tds/exam/services/impl/EqualDistributionFieldTestItemGroupSelector.java
burzum619: Removing cache and replacing with a ConcurrentHashMap implementation

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: Probably not clear from github, but there is a comment above (not inside) the method that does just that

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -57,9 +80,95 @@ public boolean isFieldTestEligible(Exam exam, Assessment assessment, String segm
...
+
+        /* StudentDLL [3119-3126] ftcount is just fieldTestItemGroups.size()
+           [3126] - Skip, debug is always = 0                                       */
+        Integer startPosition = currentSegment.getFieldTestStartPosition();
+        Integer endPosition = currentSegment.getFieldTestEndPosition();
+        int maxItems = currentSegment.getFieldTestMaxItems();
+        int minItems = currentSegment.getFieldTestMinItems();
+        int ftItemCount = previouslyAssignedFieldTestItemGroups.size(); // Initialize it as the size of the existing item groups assigned to this student
+
+        if (startPosition == null || endPosition == null) {
```
burzum619: Done, AssessmentMapper will throw an IllegalStateException if no ft end/start dates are defined in a segment with min/max ft items defined. I'll create a separate mini-PR for assessment service changes either today or early monday.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: Added to comments. The reason is because in legacy, the FT item group selection method returns a list of all field test item groups (sorted by least used) and only a subset are selected. In the new code, the selector takes in the # of items to select and only returns as many items as are needed.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: Nope, nice catch.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
burzum619: Removed, nice catch. Leftover from some refactoring.

service/src/main/java/tds/exam/models/FieldTestItemGroup.java
burzum619: Thanks, didn't unit test this apparently. Added

jtreuting: Can you add a quick unit test for that as well as fixing the code?

service/src/main/java/tds/exam/repositories/impl/FieldTestItemGroupQueryRepositoryImpl.java
burzum619: There are no updates done by any of the code that was done within this PR, but I would think updates would be done using the natural key (segmentKey, examId, and groupKey/groupid)

service/src/main/resources/db/migration/V1483732508__exam_create_table_field_test_item_group.sql
burzum619: I'll go with item_count

PR #48, opened 2017-01-15
Feature/diagnostics api

mtrupkin:
Calls the dependent micro-services health indicators and aggregates the result

### DIFF
--- a/service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
+++ b/service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
@@ -1,0 +1,39 @@
+package tds.exam.health;
+
+import com.google.common.collect.ImmutableMap;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.health.*;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+import tds.common.health.HealthIndicatorClient;
+import tds.exam.configuration.ExamServiceProperties;
+
+import java.util.Map;
+
+import static com.google.common.collect.Maps.transformValues;
+
+/**
+ * Returns the health status of this service's dependencies.
+ */
+@Component
+public class ServicesHealthIndicator implements HealthIndicator {
+    private final HealthIndicatorClient healthIndicatorClient;
+    private final Map<String, String> services;
+
+    private final HealthAggregator healthAggregator = new OrderedHealthAggregator();
+
+    @Autowired
+    ServicesHealthIndicator(final RestTemplate restTemplate, final ExamServiceProperties examServiceProperties) {
+        this.healthIndicatorClient = new HealthIndicatorClient(restTemplate);
+        services = ImmutableMap.of(
+                "assessment", examServiceProperties.getAssessmentUrl(),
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
```diff
@@ -0,0 +1,39 @@
+package tds.exam.health;
+
+import com.google.common.collect.ImmutableMap;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.health.*;
```
millerts: Splat import.  Please use the IntelliJ settings pinned to the sbac-tds channel.

service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
```diff
@@ -0,0 +1,39 @@
...
+public class ServicesHealthIndicator implements HealthIndicator {
+    private final HealthIndicatorClient healthIndicatorClient;
+    private final Map<String, String> services;
+
+    private final HealthAggregator healthAggregator = new OrderedHealthAggregator();
+
+    @Autowired
+    ServicesHealthIndicator(final RestTemplate restTemplate, final ExamServiceProperties examServiceProperties) {
+        this.healthIndicatorClient = new HealthIndicatorClient(restTemplate);
+        services = ImmutableMap.of(
```
millerts: Is the plan for ExamService to provide the health for all the microservices rather than each service provide its own status and the diagnostic service checking each service?

For May 2017 I think that is fine but moving forward I think we'll need each service to provide health.  And I don't think we should be duplicating health checks for services long term.

mtrupkin: Currently, each microservice provides its own health status.  The ExamService aggregates the results of its dependencies.

service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
```diff
@@ -0,0 +1,43 @@
...
```
jtreuting: Can we keep those consistent with using the `this.` or not.  I think both don't technically need the `this` but i think in practice we are using `this.variableName` from the constructor.

service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
```diff
@@ -0,0 +1,43 @@
...
+                "assessment", examServiceProperties.getAssessmentUrl(),
+                "student", examServiceProperties.getStudentUrl(),
+                "config",  examServiceProperties.getConfigUrl(),
+                "session", examServiceProperties.getSessionUrl());
+    }
+
+    @Override
+    public Health health() {
+        return healthAggregator.aggregate(transformValues(services, healthIndicatorClient::health));
```
jtreuting: Good find.  That's great that they offer this out of the box.

service/src/main/java/tds/exam/health/ServicesHealthIndicator.java
mtrupkin: removed the unnecessary reference to the current object

PR #50, opened 2017-01-19
Status transition validator util and test coverage. ExamService.pause…

burzum619:
Created a new StatusTransitionValidator utility class for managing the validation of exam state/status changes. Most of this class is composed of a static immutable map that maps currentStatus -> set of valid next statuses. In ExamService, pauseExam() has been renamed and made more generic.

I included some test coverage for the validator, but I did not create tests for every possible test case as there are hundreds.

The mappings in the legacy application are contained within **CommonDLL._IsValidStatusTransition_FN() - Line 465** in a very large case statement.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -259,7 +259,7 @@ class ExamServiceImpl implements ExamService {
 
         if (!StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), newStatus.getCode())) {
             return Optional.of(new ValidationError(ValidationErrorCode.EXAM_STATUS_TRANSITION_FAILURE,
-                String.format("Bad status transition from %s to %s", exam.getStatus().getCode(), newStatus.getCode())));
+                String.format("Transitioning exam status from %s to %s is not allowed", exam.getStatus().getCode(), newStatus.getCode())));
         }
 
         Exam updatedExam = new Exam.Builder()
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -47,6 +47,7 @@ import tds.exam.services.ExamService;
 import tds.exam.services.SessionService;
 import tds.exam.services.StudentService;
 import tds.exam.services.TimeLimitConfigurationService;
+import tds.exam.utils.StatusTransitionValidator;
 import tds.session.ExternalSessionConfiguration;
 import tds.session.Session;
 import tds.student.RtsStudentPackageAttribute;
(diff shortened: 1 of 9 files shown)


service/src/main/java/tds/exam/utils/StatusTransitionValidator.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import tds.exam.ExamStatusCode;
+
+/**
+ * A utility class for validating transitions between various exam states
+ */
+public class StatusTransitionValidator {
```
millerts: Suggest creating a private constructor since this class will never be instantiated.  Wouldn't make sense to instantiate but that way nothing can.

service/src/main/java/tds/exam/utils/StatusTransitionValidator.java
```diff
@@ -0,0 +1,155 @@
...
+    // Build a map of current states -> valid next states
+    private static final Map<String, Set<String>> stateMap = ImmutableMap.<String, Set<String>>builder()
+        .put(ExamStatusCode.STATUS_PENDING, new HashSet<>(Arrays.asList(
```
millerts: Since we're using Guava you could use [Sets.newHashSet()](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Sets.html#newHashSet-E...-)

service/src/main/java/tds/exam/utils/StatusTransitionValidator.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import tds.exam.ExamStatusCode;
+
+/**
+ * A utility class for validating transitions between various exam states
+ */
+public class StatusTransitionValidator {
+    // Build a map of current states -> valid next states
```
jeffjohnson9046: Maybe add a reference to where the map came from in the legacy app (`CommonDLL._IsValidStatusTransition_FN()`)?  Not critical, but might be nice to know where the map came from in case we need to check it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -247,22 +248,27 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+    @Override
+    public Optional<ValidationError> updateExamStatus(UUID examId, ExamStatusCode newStatus, String statusChangeReason) {
         Exam exam = examQueryRepository.getExamById(examId)
             .orElseThrow(() -> new NotFoundException(String.format("Exam could not be found for id %s", examId)));
 
-        if (!statusesThatCanTransitionToPaused.contains(exam.getStatus().getCode())) {
+        if (!StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), newStatus.getCode())) {
             return Optional.of(new ValidationError(ValidationErrorCode.EXAM_STATUS_TRANSITION_FAILURE,
-                String.format("Bad status transition from %s to %s", exam.getStatus().getCode(), ExamStatusCode.STATUS_PAUSED)));
+                String.format("Bad status transition from %s to %s", exam.getStatus().getCode(), newStatus.getCode())));
```
millerts: I know this isn't critical but can we tweak this to be `Transitioning exam status from %s to %s is not allowed`

millerts:
I would like better testing around this logic but I agree that we can revisit it later.

PR #51, opened 2017-01-20
Feature/field test select items - item "cohort" code

burzum619:
This is an addendum to the field test item group selection PR. I've attempted to explain the need for the changes as best as possible in the comments within the code (FieldTestServiceImpl.java, specifically).

In summary, we need to account for the possible case where there is an item group containing a combination of field test items and non-field test items. A field test item group should only be selected if the exam has enough "space" for it, and non-ft items also need to be factored into this comparison. 

The `shouldSkipGroupDueToTooManyItemsInFTGroup()` test case in `FieldTestServiceImplTest` covers a case where there is space for only the first item group (which contains three FT items, as well as an additional non-field test item,). In this case, the second field test item group cannot be selected as there is room for only four items.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
@@ -93,7 +93,7 @@ public class FieldTestServiceImpl implements FieldTestService {
         Integer endPosition = currentSegment.getFieldTestEndPosition();
         int maxItems = currentSegment.getFieldTestMaxItems();
         int minItems = currentSegment.getFieldTestMinItems();
-        int ftItemCount = previouslyAssignedFieldTestItemGroups.size(); // Initialize it as the size of the existing item groups assigned to this student
+        int totalFtItemCount = previouslyAssignedFieldTestItemGroups.size(); // Initialize it as the size of the existing item groups assigned to this student
 
         /* [3131-3137] Note tht if endPos - startPos < numIntervals, the integer division below results in 0, which
         *  results in  a division-by-zero at line [3320] */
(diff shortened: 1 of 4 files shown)


service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -114,24 +140,35 @@ public int selectItemGroups(final Exam exam, final Assessment assessment, final
...
+            int ftItemCount = fieldTestItemGroup.getItemCount();
+
+            // Skip this group if the cohortItemCount is greater than or equal to the maximum number of field test items for this segment
+            // Ultimately we want to make sure that there aren't more items (including non-ft items) in the group
+            // than what will fit into the exam.
+            if (cohortGroupCount == 0 || cohortItemCount >= maxItems) {
+                /* This break corresponds to [3301-3305] - instead of "continuing" the selection loop, we can just break out and stop selecting.
+                *  The legacy app simply continues to loop unnecessarily. At this point, cohortItemCount and maxItems will never change. */
+                break;
+            } else if (ftItemCount > 0 && totalFtItemCount + ftItemCount <= maxItems) { /* [3307] */
```
jtreuting: Couldn't these 2 if statements be combined to have a single inclusion criteria like

```
if (cohortGroupCount != 0 
	&& cohortItemCount < maxItems
	&& ftItemCount > 0 && totalFtItemCount + ftItemCount <= maxItems) {
```

burzum619: I was going to do that, but it is slightly less efficient... in that case we will "continue" the loop until we process all remaining FieldTestItemGroups. Instead i'm  just breaking out of the loop altogether. If that condition is hit, itemCount >= maxItems always, no point in continuing the loop

burzum619: its a "break" inside that if. If it was a "continue" like in the legacy app, then the if statement you just posted would be equivalent

jtreuting: You're right, that makes sense.

PR #52, opened 2017-01-22
Feature/accommodations in assessment

millerts:
There were some issues when communicating with assessment service from config service in docker.

The endpoints that were required assessment service always belonged in assessment service after we decided to build endpoints based on intent rather than strictly by DB.  Rather than investigate the docker issues I moved the endpoints since that was on the TODO list.

I need this for other work so I'll be merging this code but will make follow up changes depending on review.

### DIFF
--- a/service/src/main/java/tds/exam/services/AssessmentService.java
+++ b/service/src/main/java/tds/exam/services/AssessmentService.java
@@ -1,8 +1,12 @@
 package tds.exam.services;
 
+import java.util.List;
 import java.util.Optional;
 
+import tds.accommodation.Accommodation;
 import tds.assessment.Assessment;
+import tds.assessment.AssessmentWindow;
+import tds.session.ExternalSessionConfiguration;
 
 /**
  * Service handles assessment interaction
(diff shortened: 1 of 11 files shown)


service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
```diff
@@ -45,4 +52,60 @@ public AssessmentServiceImpl(RestTemplate restTemplate, ExamServiceProperties ex
...
+
+        ResponseEntity<List<Accommodation>> responseEntity = restTemplate.exchange(builder.toUriString(),
+            HttpMethod.GET, null, new ParameterizedTypeReference<List<Accommodation>>() {
+            });
+
+        return responseEntity.getBody();
+    }
+
+    @Override
+    public List<Accommodation> findAssessmentAccommodationsByAssessmentId(String clientName, String assessmentId) {
```
jtreuting: The interface has `final` for both parameters

service/src/main/java/tds/exam/services/AssessmentService.java
```diff
@@ -16,4 +20,34 @@
...
+    /**
+     * Finds the assessment windows for an exam
+     *
+     * @param clientName    environment's client name
+     * @param assessmentId  the assessment id for the assessment
+     * @param studentId     identifier to the student
+     * @param configuration {@link tds.session.ExternalSessionConfiguration} for the environment
+     * @return array of {@link tds.assessment.AssessmentWindow}
+     */
+    List<AssessmentWindow> findAssessmentWindows(String clientName, String assessmentId, long studentId, ExternalSessionConfiguration configuration);
```
jtreuting: All these parameters should be `final` I believe.  Here and in the impl.

service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
```diff
@@ -45,4 +52,60 @@ public AssessmentServiceImpl(RestTemplate restTemplate, ExamServiceProperties ex
 
         return maybeAssessment;
     }
+
+    @Override
+    @Cacheable(CacheType.MEDIUM_TERM)
+    public List<AssessmentWindow> findAssessmentWindows(String clientName,
+                                                        String assessmentId,
+                                                        long studentId,
+                                                        ExternalSessionConfiguration configuration) {
```
jtreuting: Since this method is being cached with the `@Cacheable` we need to make sure to implement any object parameters `hashCode` method so that an object with the same properties will generate the same hashCode and therefore take advantage of caching.

https://github.com/SmarterApp/TDS_SessionService/blob/develop/client/src/main/java/tds/session/ExternalSessionConfiguration.java

millerts: Sounds good.  I will do that throughout because we are not consistent in doing it.

PR #53, opened 2017-01-23
Feature/get page content

jeffjohnson9046:
Implementation for [TDS-354](https://jira.fairwaytech.com/browse/TDS-354): get page content


service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -90,10 +145,17 @@ public int getPosition() {
     }
 
     /**
-     * @return The item type (e.g. MI for "Matching Item", MC for "Multiple Choice"
+     * @return True if the item has been selected; otherwise false
      */
-    public String getType() {
-        return type;
+    public boolean isSelected() {
+        return selected;
```
jeffjohnson9046: Couldn't determine what this field means or where it comes from in legacy code. Doing a search for `isSelected` in GitHub returns quite a few results, more than can be dealt with.  I couldn't think of another way to pare down the results (because this field is quite possibly used in the UI somewhere)

service/src/main/java/tds/exam/repositories/impl/ExamPageCommandRepositoryImpl.java
```diff
@@ -58,29 +65,30 @@ public void deleteAll(final UUID examId) {
...
-            "   exam_page_id, UTC_TIMESTAMP(), started_at \n" +
-            "FROM \n" +
-            "   exam_page_event PE\n" +
-            "JOIN \n" +
-            "   exam_page P\n " +
-            "ON \n" +
-            "   PE.exam_page_id = P.id \n" +
-            "WHERE \n " +
-            "   P.exam_id = :examId";
+                "   exam_page_event (exam_page_id, deleted_at, started_at) \n" +
```
jeffjohnson9046: Formatting change; looks like this got caught by asking IntelliJ to format the code.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
         return jdbcTemplate.query(SQL, parameters, examPageRowMapper);
     }
 
+    @Override
+    public Optional<ExamPage> find(UUID examId, int position) {
```
jeffjohnson9046: This probably shouldn't be `Optional`: I think by the time we're at the point in the workflow where we're calling for pages (i.e. the exam has already been started), the page we're looking for should _definitely_ exist.  If it doesn't, then we probably have a bigger issue.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
...
+                    examPageRowMapper));
+        } catch (IncorrectResultSizeDataAccessException e) {
+            LOG.debug("{} did not return results for examId {}, position {}", SQL, examId, position);
+        }
+
+        return maybeExamPage;
+    }
+
+    @Override
+    public Optional<ExamPage> findPageWithItems(UUID examId, int position) {
```
jeffjohnson9046: This probably shouldn't be `Optional`: I think by the time we're at the point in the workflow where we're calling for pages (i.e. the exam has already been started), the page we're looking for should _definitely_ exist.  If it doesn't, then we probably have a bigger issue.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
...
+        final MapSqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examId))
+            .addValue("position", position);
+
+        final String SQL =
+            "SELECT \n" +
```
jeffjohnson9046: Opted for getting a fully hydrated `ExamPage` here.  Made the query a bit more complicated, but saved a round-trip to the database (i.e. don't have to get the `ExamPage` _then_ its `ExamItem`s collection)

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
...
+                "   segment.segment_position \n" +
+                "FROM \n" +
+                "   exam_page page \n" +
+                "JOIN ( \n" +
+                "   SELECT \n" +
+                "       exam_page_id, \n" +
+                "       MAX(id) AS id \n" +
+                "   FROM \n" +
+                "       exam_page_event \n" +
+//                "   WHERE \n" +
```
jeffjohnson9046: Need to remove commented code.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
...
+                "JOIN \n" +
+                "   exam_item item \n" +
+                "   ON page.id = item.exam_page_id \n" +
+                "LEFT JOIN \n" +
+                "   (SELECT \n" +
+                "       exam_item_id, \n" +
+                "       MAX(id) AS id\n" +
+                "   FROM \n" +
+                "       exam_item_response \n" +
+//                "   WHERE \n" +
```
jeffjohnson9046: Need to remove commented code.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
...
+     * @return Return the text of the response as provided by the student, if one is provided
+     */
+    public String getResponseText() {
+        if (response == null) {
+            return "";
+        }
+
+        return response.getResponse() == null
+            ? ""
+            : response.getResponse();
```
jtreuting: Feels like the logic can be combined either by moving this null check and combing above like:

```
if (response == null || response.getResponse() == null) {
    return "";
}

return response.getResponse();
```

Or by having it become:

```
return response == null || response.getResponse() = null ? "" : response.getResponse();
```

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
...
+     */
+    public int getResponseLength() {
+        if (response == null) {
+            return 0;
+        }
+
+        return response.getResponse() == null
+            ? 0
+            : response.getResponse().length();
+    }
```
jtreuting: Same for here

service/src/main/java/tds/exam/models/ExamPage.java
```diff
@@ -91,20 +153,61 @@ public int getPagePosition() {
...
+     */
+    public int getGroupItemsRequired() {
+        // This value is only ever 0 or -1 in the database.  When the value is -1 all items are required (from comment
+        // in https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/cebc3996ab000dc604b539da51235eaf20039d0d/database%20scripts%20-%20mysql/Session/User-Defined%20Functions/iscomplete.sql#L46
+
+        // In legacy, this value is stored in the session.testeeresponse table.  Since it pertains to an item group
+        // (which is a super-set of items), it seems appropriate to store it at the page level.  From here, it can be
+        // mapped to the legacy OpportunityItem.
+        return groupItemsRequired;
+    }
```
jtreuting: Based on your comments, does it make more sense to store this as a `boolean` in the new DB instead?

service/src/main/java/tds/exam/repositories/ExamAccommodationQueryRepository.java
```diff
@@ -14,7 +14,7 @@
      *
      * @param examId             The id of the {@link tds.exam.Exam} that owns the {@link tds.exam.ExamAccommodation}s
      * @param segmentKey         The key of the segment to which the {@link tds.exam.ExamAccommodation}s apply
-     * @param accommodationTypes a list of types of {@link tds.exam.ExamAccommodation}s to find
+     * @param accommodationTypes a list of types of {@link tds.exam.ExamAccommodation}s to getPage
```
jtreuting: I'm not sure I follow this change?  Maybe find and replace happened?

service/src/main/java/tds/exam/repositories/ExamResponseQueryRepository.java
```diff
@@ -10,7 +10,7 @@
     /**
      * Gets the item position of the last item that has a response
      *
-     * @param examId the id of the {@link tds.exam.Exam} to find the position for
+     * @param examId the id of the {@link tds.exam.Exam} to getPage the position for
```
jtreuting: Same find and replace possibly

service/src/main/java/tds/exam/services/ExamPageService.java
```diff
@@ -3,17 +3,16 @@
...
  */
-public interface ExamItemService {
+public interface ExamPageService {
     /**
      * Inserts a {@link java.util.List} of {@link tds.exam.models.ExamPage}s
      *
-     * @param examPage
+     * @param examPage A collection of {@link tds.exam.models.ExamPage} to insert
      */
     void insertPages(List<ExamPage> examPage);
```
jtreuting: I know this is outside the changes you made, but shouldn't this param be plural: `List<ExamPage> examPages`

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    public List<ExamPage> findAllPages(final UUID examId) {
+        return examPageQueryRepository.findAll(examId);
+    }
+
+    @Override
+    public ExamPage getPage(UUID examId, int pageNumber) {
+        // TODO: Validate testee access equivalent?
+        // Collect data
+        Exam exam = examQueryRepository.getExamById(examId)
+            .orElseThrow(() -> new NotFoundException(String.format("Could not getPage an exam for id %s", examId)));
```
jtreuting: same find and replace?  "Could not **getPage** an exam ..."

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -142,7 +142,7 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         //the reference to those parts that require it.
         Optional<Session> maybeSession = sessionService.findSessionById(openExamRequest.getSessionId());
         if (!maybeSession.isPresent()) {
-            throw new IllegalArgumentException(String.format("Could not find session for id %s", openExamRequest.getSessionId()));
+            throw new IllegalArgumentException(String.format("Could not getPage session for id %s", openExamRequest.getSessionId()));
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -154,7 +154,7 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
 
         if (!openExamRequest.isGuestStudent()) {
             studentService.getStudentById(openExamRequest.getStudentId()).orElseThrow((Supplier<RuntimeException>) ()
-                -> new IllegalArgumentException(String.format("Could not find student for id %s", openExamRequest.getStudentId()))
+                -> new IllegalArgumentException(String.format("Could not getPage student for id %s", openExamRequest.getStudentId()))
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -316,7 +316,7 @@ public void pauseAllExamsInSession(UUID sessionId) {
         }
 
         Session session = sessionService.findSessionById(approvalRequest.getSessionId())
-            .orElseThrow(() -> new IllegalArgumentException("Could not find session for id " + approvalRequest.getSessionId()));
+            .orElseThrow(() -> new IllegalArgumentException("Could not getPage session for id " + approvalRequest.getSessionId()));
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -332,7 +332,7 @@ public void pauseAllExamsInSession(UUID sessionId) {
         // RULE:  Student should not be able to start an exam if the TA check-in window has expired.
         TimeLimitConfiguration timeLimitConfig =
             timeLimitConfigurationService.findTimeLimitConfiguration(approvalRequest.getClientName(), exam.getAssessmentId())
-                .orElseThrow(() -> new IllegalArgumentException(String.format("Could not find time limit configuration for client name %s and assessment id %s", approvalRequest.getClientName(), exam.getAssessmentId())));
+                .orElseThrow(() -> new IllegalArgumentException(String.format("Could not getPage time limit configuration for client name %s and assessment id %s", approvalRequest.getClientName(), exam.getAssessmentId())));
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -532,7 +532,7 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
 
         //OpenTestServiceImpl line 367 - 368 validation check.  no window no exam
         if (!maybeWindow.isPresent()) {
-            return new Response<Exam>(new ValidationError(NO_OPEN_ASSESSMENT_WINDOW, "Could not find an open assessment window"));
+            return new Response<Exam>(new ValidationError(NO_OPEN_ASSESSMENT_WINDOW, "Could not getPage an open assessment window"));
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -37,7 +37,7 @@ public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
     private Set<Item> getItemPool(final UUID examId, final List<ItemConstraint> itemConstraints, final List<Item> items, Boolean isFieldTest) {
         /*
             This method is meant to replace StudentDLL._AA_ItempoolString_FNOptimized() [1643]
-            The purpose of this method is to find the list of items to include in the segment by taking the following steps:
+            The purpose of this method is to getPage the list of items to include in the segment by taking the following steps:
```
jtreuting: change `getPage` to `find`

service/src/main/java/tds/exam/services/impl/ItemPoolServiceImpl.java
```diff
@@ -60,7 +60,7 @@ public ItemPoolServiceImpl(ExamAccommodationService examAccommodationService) {
                 .map(itemConstraint -> accommodation))
             .collect(Collectors.toSet());
 
-        // For the included accommodations above, find the list of compatible item ids
+        // For the included accommodations above, getPage the list of compatible item ids
```
jtreuting: change `getPage` to `find`

service/src/main/resources/db/migration/V1484676942__exam_add_exam_item_columns.sql
```diff
@@ -0,0 +1,20 @@
...
+Desc: Adds columns to the exam_item table for supporting the get page content feature
+
+***********************************************************************************************************************/
+USE exam;
+
+ALTER TABLE exam_item
+  ADD is_selected BIT NOT NULL DEFAULT b'0' AFTER is_required;
+
+ALTER TABLE exam_item
+  ADD is_marked_for_review BIT NULL AFTER is_selected;
```
jtreuting: Shouldn't this be NOT NULL and default to `false`?  It's not a mark for review unless the student clicks to make it one.

service/src/test/java/tds/exam/builder/SegmentBuilder.java
```diff
@@ -8,7 +8,8 @@
 import tds.assessment.Segment;
 
 public class SegmentBuilder {
-    private String key = "(SBAC_PT)IRP-Perf-ELA-3-Summer-2015-2016";
+    public static final String DEFAULT_SEGMENT_KEY = "(SBAC_PT)IRP-Perf-ELA-3-Summer-2015-2016";
+    private String key = DEFAULT_SEGMENT_KEY;
```
jtreuting: small thing.  line break between the public static and private

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -61,6 +69,171 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParamete
...
+                "       exam_page_id \n" +
+                ") last_event \n" +
+                "   ON P.id = last_event.exam_page_id \n" +
+                "JOIN \n" +
+                "   exam_page_event PE \n" +
+                "   ON PE.id = last_event.id \n" +
+                "WHERE \n" +
+                "   exam_id = :examId \n" +
+                "   AND page_position = :position \n" +
+                "   AND PE.deleted_at IS NULL";
```
jtreuting: The next method SQL statement uses fuller alias names like `page` and `page_event` so it seems like it might be nice to be consistent even though the `P` and `PE` are easier to understand in this query than the next.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
jtreuting: Is this comment supposed to be for the next method `findPageWithItems`?

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
jtreuting: How come this WHERE clause is commented out in this SQL but not in the one above that is similar but doesn't get items?  Wouldn't this make this subquery smaller and only include the items that are needed?

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -1,37 +1,80 @@
...
+
     public ExamItem(Builder builder) {
-        this.id = builder.id;
-        this.itemKey = builder.itemKey;
-        this.position = builder.position;
-        this.type = builder.type;
-        this.fieldTest = builder.fieldTest;
-        this.segmentId = builder.segmentId;
-        this.required = builder.required;
+        id = builder.id;
```
millerts: In setters and builder we usually use this but that is mostly do to the fact that the argument being passed into the setter is the same name as the instance variable so the `this` is required.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
...
 
     /**
-     * @return Flag indicating whether this is a required item.
+     * @return The bank key of the {@link tds.assessment.Item} this item represents
      */
-    public boolean isRequired() {
-        return required;
+    public long getAssessmentBankKey() {
+        // Key convention for an item: [bank key]-[item key], e.g. 187-2602
+        return Long.parseLong(assessmentItem.getId().split("-")[0]);
```
millerts: I think we should do this parsing in the setters or object construction.  No need to parse each time and we'd rather see the parsing error when creating the object rather than when the `get` is called.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
...
+     * @return True if the {@link tds.assessment.Item} is marked as required; otherwise false
+     */
+    public boolean getAssessmentItemIsRequired() {
+        return assessmentItem.isRequired();
+    }
+
+    /**
+     * @return Return the text of the response as provided by the student, if one is provided
+     */
+    public String getResponseText() {
```
millerts: Why do we have getters for the response object properties in this class?  This class already has a getter for the `Response` object.  I would expect class to do `item.getResponse().getResponse()`.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
     }
 
     /**
-     * @return The id of the {@link tds.assessment.Segment} the item belongs to
+     * @return The most recent {@link tds.exam.models.ExamItemResponse} for this item
+     */
+    public ExamItemResponse getResponse() {
```
millerts: I feel this should return `Optional<ExamItemResponse>` since this can be null.

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -1,37 +1,80 @@
...
-        this.required = builder.required;
+        id = builder.id;
+        examPageId = builder.examPageId;
+        itemKey = builder.itemKey;
+        position = builder.position;
+        selected = builder.selected;
+        markedForReview = builder.markedForReview;
+        fieldTest = builder.fieldTest;
+        response = builder.response;
+        assessmentItem = builder.assessmentItem;
```
millerts: The getters are written as if assessmentItem is never null accessing it without doing null checks.  Either those getters need to be rewritten to be null safe or we should force the builder to be created with the value.

```{java}
public Builder(AssessmentItem assessmentItem) {
  this.assessmentItem = checkNotNull(assessmentItem);
}
```

service/src/main/java/tds/exam/models/ExamItem.java
```diff
@@ -104,17 +166,101 @@ public boolean isFieldTest() {
...
+    public long getAssessmentItemKey() {
+        // Key convention for an item: [bank key]-[item key], e.g. 187-2602
+        return Long.parseLong(assessmentItem.getId().split("-")[1]);
+    }
+
+    /**
+     * @return The path to the XML file for the {@link tds.assessment.Item} this item represents
+     */
+    public String getAssessmentItemFilePath() {
+        return assessmentItem.getItemFilePath();
```
millerts: How did you determine when to provide these getters on objects within the class and when to just have the calling code do ExamItem.getAssessmentItem()?  Same comment with the response object.  This class provides some getters into the sub objects.  

Will we duplicate getters for every getter in response/assessmentItem?

service/src/main/java/tds/exam/services/ExamPageService.java
jeffjohnson9046: Yep - it should be.  But I didn't change it because it was outside the changes for this feature.  As I'm going through and addressing other comments, I'll fix this.

service/src/main/java/tds/exam/models/ExamItem.java
millerts: I lean towards these methods that are reaching into response and assessment item should be removed.

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
jeffjohnson9046: Sigh... this is why I so rarely use find/replace and do it myself.

service/src/main/java/tds/exam/models/ExamPage.java
millerts: I think it should.  We can map the boolean to the legacy via @jeffjohnson9046  comments.

service/src/main/java/tds/exam/repositories/ExamResponseQueryRepository.java
millerts: Yes.  When doing refactors using IntelliJ I almost always tell it to ignore any text matches.  I find it is more trouble than help.

service/src/main/java/tds/exam/repositories/impl/ExamPageCommandRepositoryImpl.java
```diff
@@ -31,24 +31,31 @@ public ExamPageCommandRepositoryImpl(@Qualifier("commandJdbcTemplate") NamedPara
...
         final String examPageSQL =
             "INSERT INTO exam_page (\n" +
-            "   page_position, item_group_key, exam_id\n" +
-            ") \n" +
-            "VALUES (\n" +
-            "   :pagePosition, \n" +
-            "   :itemGroupKey, \n" +
-            "   :examId\n" +
-            ")";
+                "   page_position, exam_segment_key, item_group_key, group_items_required, exam_id\n" +
```
millerts: Can you add carriage returns after these since we're doing the same below?

service/src/main/java/tds/exam/repositories/impl/ExamPageCommandRepositoryImpl.java
millerts: Cool.  I thought I formatted all this code.  Must have missed a couple classes or were added after the formatting.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -37,10 +39,12 @@
 @RequestMapping("/exam")
 public class ExamController {
     private final ExamService examService;
+    private final ExamPageService examPageService;
 
     @Autowired
-    public ExamController(ExamService examService) {
+    public ExamController(ExamService examService, ExamPageService examPageService) {
```
millerts: Rather than add page and item stuff to this controller can we just create a new controller to have that information?

service/src/main/resources/db/migration/V1484849630__exam_add_columns_to_exam_page.sql
```diff
@@ -0,0 +1,19 @@
...
+Desc: Adds columns to the exam_page table for supporting the get page content feature
+
+***********************************************************************************************************************/
+USE exam;
+
+ALTER TABLE exam_page
+  ADD group_items_required INT NOT NULL DEFAULT -1 AFTER item_group_key;
+
+ALTER TABLE exam_page
+  ADD exam_segment_key VARCHAR(250) NULL AFTER page_position;
```
millerts: Why should this ever be null?

service/src/test/java/tds/exam/repositories/impl/ExamItemCommandRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,133 @@
...
+    private NamedParameterJdbcTemplate commandJdbcTemplate;
+
+    private ExamItemCommandRepository examItemCommandRepository;
+    private ExamPageQueryRepository examPageQueryRepository;
+
+    private Exam mockExam = new ExamBuilder().build();
+    private ExamPage mockPage = new ExamPageBuilder()
+        .withExamId(mockExam.getId())
+        .build();
+    @Before
```
millerts: Can you add a line break between the variable instantiation and the before annotation?

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+                .fromExamItem(examItem)
+                .withAssessmentItem(assessmentItemsForSegmentMap.get(examItem.getItemKey()))
+                .build())
+            .collect(Collectors.toList());
+
+        // Build up the completed exam page with all its components
+        ExamPage completeExamPage = new ExamPage.Builder()
+            .fromExamPage(examPage)
+            .withExamItems(examItemsWithAssessmentItems)
+            .withStartedAt(Instant.now())
```
millerts: So each time a student views a page the time spent on a page resets?  So if the student spent 20 minutes when originally answering the question but then goes back to double check that they've answered the questions spending 10 seconds the 10 seconds is all that is kept?  Not 20 minutes and 10 seconds?

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public ExamPage getPage(UUID examId, int pageNumber) {
+        // TODO: Validate testee access equivalent?
+        // Collect data
+        Exam exam = examQueryRepository.getExamById(examId)
+            .orElseThrow(() -> new NotFoundException(String.format("Could not getPage an exam for id %s", examId)));
+
+        Assessment assessment = assessmentService.findAssessment(exam.getClientName(), exam.getAssessmentKey())
+            .orElseThrow(() -> new NotFoundException(String.format("Could not find an assessment for client id %s and key %s", exam.getClientName(), exam.getAssessmentKey())));
```
millerts: Nit but I think this should say `"Could not find the associated assessment for exam %s", examId`.  That'll make for a better explanation to a caller the problem.

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public List<ExamPage> findAllPages(final UUID examId) {
+        return examPageQueryRepository.findAll(examId);
+    }
+
+    @Override
+    public ExamPage getPage(UUID examId, int pageNumber) {
+        // TODO: Validate testee access equivalent?
```
millerts: What is the answer to this TODO?  I know the current code does this check before providing the page.

service/src/main/java/tds/exam/models/ExamItem.java
jeffjohnson9046: Pursuant to code review, this field is being ported to the `ExamItem`

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
burzum619: No, it should keep the total duration (the sum of the two "visits"). 

The legacy app calculates the duration on the client side and is sent across to the server during /updateResponses. This duration is then summed and set in the "update" query for testeeresponse. We'll have to take a different approach in our new implementation since we are not making updates. 

Could we possibly create a service/repository method that the student report processor code would call when constructing the TRT? Then we could look at the individual "exam_page_event" records to calculate the sum of durations for each page/item.

service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
jeffjohnson9046: The answer is "yes - the validate testee access equivalent should be called".  I'll be sure to make this change along with the other changes I'm making.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -71,7 +71,7 @@
 
     private final ExamQueryRepository examQueryRepository;
     private final ExamCommandRepository examCommandRepository;
-    private final ExamItemService examItemService;
+    private final ExamPageService examItemService;
```
burzum619: May wanna update the property name to examPageService

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
jeffjohnson9046: It turns out that MySQL will not allow referencing the outer query in a subquery like this.  The `find` method in `ExamPageQueryRepository` was left over from my previous attempt to implement this feature and will be removed.  It wouldn't have worked either.

We may need to come up with another way to get at our event data.  For the immediate release, I'm going to remove the commented out `WHERE` clauses in the subqueries.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
```diff
@@ -35,45 +35,45 @@ public ExamAccommodationQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") Name
...
-            "   ea.`type`, \n" +
-            "   ea.code, \n" +
-            "   ea.description, \n" +
-            "   eae.denied_at, \n" +
-            "   ea.created_at, \n" +
-            "   ea.allow_change, \n" +
-            "   ea.value, \n" +
-            "   ea.segment_position, \n" +
-            "   eae.selectable, \n" +
+                "   ea.id, \n" +
```
jeffjohnson9046: Just a formatting change; caught by `cmd` + `alt` + `l`

service/src/main/resources/db/migration/V1484676942__exam_add_exam_item_columns.sql
jeffjohnson9046: Addressed in `V1485214567__exam_exam_item_column_changes.sql`

service/src/main/resources/db/migration/V1484849630__exam_add_columns_to_exam_page.sql
jeffjohnson9046: Addressed in `V1485218648__exam_exam_page_schema_changes.sql`

client/src/main/java/tds/exam/ExamItemResponse.java
```diff
@@ -0,0 +1,126 @@
...
+    private String response;
+    private int sequence;
+    private boolean valid;
+    private Instant createdAt;
+
+    private ExamItemResponse() {
+    }
+
+    public ExamItemResponse(Builder builder) {
+        this.id = builder.id;
```
jtreuting: Can we be consistent with using the `this` qualifier within our constructors with builders.  I honestly can't remember if we came to a conclusion on this when we discussed it last or not.  But in the ExamItem constructor above it does not use `this.`

client/src/main/java/tds/exam/ExamPage.java
```diff
@@ -0,0 +1,231 @@
...
+    private Instant startedAt;
+
+    /**
+     * For frameworks
+     */
+    private ExamPage() {
+    }
+
+    public ExamPage(Builder builder) {
+        id = builder.id;
```
jtreuting: Same consistency as mentioned above.

service/src/main/java/tds/exam/services/impl/ExamApprovalServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+            return Optional.of(new ValidationError(ValidationErrorCode.EXAM_APPROVAL_SESSION_CLOSED, "The session is not available for testing, please check with your test administrator."));
+        }
+
+        // RULE:  If the session has no proctor, there is nothing to approve.  This is either a guest session or an
+        // otherwise proctor-less session.
+        if (session.isProctorless()) {
+            return Optional.empty();
+        }
+
+        // RULE:  Student should not be able to start an exam if the TA check-in window has expired.
```
jtreuting: Might be worth mentioning more details around the TA check-in window because it was hard for us to figure out at first.  There is a scheduled call from the proctor application to an endpoint every 3 minutes (though that is configurable) that is a heartbeat essentially and says that the proctor is still on the site and there.  If they go away for too long it assumes there is no proctor and shuts down all exams in the session.

I mention it because it is feels odd that it is part of this check since it has nothing to do with this student that is being validated.

service/src/main/java/tds/exam/web/endpoints/ExamApprovalController.java
```diff
@@ -0,0 +1,44 @@
...
+@RequestMapping("/exam")
+public class ExamApprovalController {
+    private final ExamApprovalService examApprovalService;
+
+    @Autowired
+    public ExamApprovalController(ExamApprovalService examApprovalService) {
+        this.examApprovalService = examApprovalService;
+    }
+
+    @RequestMapping(value = "/{id}/get-approval", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
jtreuting: Seems like the `get-` part of the path is redundant since this is making an HTTP `GET` for an Approval.

PR #54, opened 2017-01-23
Feature/add client name lookup

millerts:
The biggest change is that we get the client name from session rather than having the legacy application pass the client name

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
@@ -83,7 +83,7 @@ class AssessmentServiceImpl implements AssessmentService {
     public List<Accommodation> findAssessmentAccommodationsByAssessmentKey(final String clientName, final String assessmentKey) {
         UriComponentsBuilder builder =
             UriComponentsBuilder
-                .fromHttpUrl(String.format("%s/%s/accommodations", examServiceProperties.getAssessmentUrl(), clientName))
+                .fromHttpUrl(String.format("%s/%s/assessments/accommodations", examServiceProperties.getAssessmentUrl(), clientName))
                 .queryParam("assessmentKey", assessmentKey);
 
         ResponseEntity<List<Accommodation>> responseEntity = restTemplate.exchange(builder.toUriString(),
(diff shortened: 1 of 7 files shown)


service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
```diff
@@ -83,7 +83,7 @@ public AssessmentServiceImpl(RestTemplate restTemplate, ExamServiceProperties ex
     public List<Accommodation> findAssessmentAccommodationsByAssessmentKey(final String clientName, final String assessmentKey) {
         UriComponentsBuilder builder =
             UriComponentsBuilder
-                .fromHttpUrl(String.format("%s/%s/accommodations", examServiceProperties.getAssessmentUrl(), clientName))
+                .fromHttpUrl(String.format("%s/%s/assessments/accommodations", examServiceProperties.getAssessmentUrl(), clientName))
```
millerts: I missed this in the earlier accommodations changes.

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -74,7 +74,7 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
                 .withCode(accommodation.getCode())
                 .withType(accommodation.getType())
                 .withDescription(accommodation.getValue())
-                .withSegmentKey(accommodation.getSegmentKey())
+                .withSegmentKey(accommodation.getSegmentKey() != null ? accommodation.getSegmentKey() : exam.getAssessmentKey())
```
millerts: SegmentKey is not null in the DB.  This will be null from the accommodations service if it is a single segmented exam.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -147,6 +138,16 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
+        //Line 5602 in StudentDLL.  This has been moved to earlier in the flow than the original because it is used throughout.  The original
+        //fetches the external configuration multiple times in the different layers.
+        Optional<ExternalSessionConfiguration> maybeExternalSessionConfiguration = sessionService.findExternalSessionConfigurationByClientName(currentSession.getClientName());
+        if (!maybeExternalSessionConfiguration.isPresent()) {
+            throw new IllegalStateException(String.format("External Session Configuration could not be found for client name %s", currentSession.getClientName()));
+        }
+
+        ExternalSessionConfiguration externalSessionConfiguration = maybeExternalSessionConfiguration.get();
+
+
```
jtreuting: How come this moved down below?

millerts: The call to get `ExternalSessionConfiguration` requires the clientName which we get from the session.  So this now needs to come after we get the current session.

jtreuting: Ooops, totally missed that.

PR #55, opened 2017-01-24
Changed ID type to UUIDs for ExamAccommodation and ExamPage

mtrupkin:
Changed db table primary keys to char(36) to store UUIDs:
  exam_accommodation
  exam_page

Changed tables with foreign key references:
  exam_accommodation_event
  exam_item
  exam_page_event

### DIFF
--- a/service/src/main/java/tds/exam/models/ExamPage.java
+++ b/service/src/main/java/tds/exam/models/ExamPage.java
@@ -8,7 +8,7 @@ import java.util.UUID;
  * Represents the page of an exam
  */
 public class ExamPage {
-    private long id;
+    private UUID id;
     private int pagePosition;
     private String itemGroupKey;
     private UUID examId;
@@ -29,7 +29,7 @@ public class ExamPage {
     }
 
     public static final class Builder {
-        private long id;
+        private UUID id = UUID.randomUUID();
         private int pagePosition;
         private String itemGroupKey;
         private UUID examId;
@@ -67,7 +67,7 @@ public class ExamPage {
             return this;
         }
 
-        public Builder withId(long id) {
+        public Builder withId(UUID id) {
             this.id = id;
             return this;
         }
@@ -80,7 +80,7 @@ public class ExamPage {
     /**
      * @return The id of the {@link tds.exam.models.ExamPage} record
      */
-    public long getId() {
+    public UUID getId() {
         return this.id;
     }
     /**
@@ -123,9 +123,5 @@ public class ExamPage {
      */
     public Instant getStartedAt() {
         return startedAt;
-    }
-
-    public void setId(long id) {
-        this.id = id;
     }
 }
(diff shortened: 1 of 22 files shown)


service/src/main/java/tds/exam/models/ExamPage.java
```diff
@@ -29,7 +29,7 @@ public ExamPage(Builder builder) {
     }
 
     public static final class Builder {
-        private long id;
+        private UUID id = UUID.randomUUID();
```
millerts: Can we make this consistent across builders?  We have this style and the following:

```
Builder(UUID id) {
  this.id = id;
}
```

And some where we don't even set it when building the builder.  Personally I like the constructor because it is clear to the callers the ID is being set.  And if we do that I also think the `fromObject` type methods should be static always returning a new builder.  That way you don't have `new Builder(id).from(obj)`

millerts: Since your changing this.  I know I haven't been consistent.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
```diff
@@ -30,27 +30,24 @@ public ExamAccommodationCommandRepositoryImpl(@Qualifier("commandJdbcTemplate")
...
     public void insert(List<ExamAccommodation> accommodations) {
-        String SQL = "INSERT INTO exam_accommodation(exam_id, segment_key, type, code, description, allow_change, value, segment_position) \n" +
-            "VALUES(:examId, :segmentKey, :type, :code, :description, :allowChange, :value, :segmentPosition)";
+        String SQL = "INSERT INTO exam_accommodation(exam_id, id, segment_key, type, code, description, allow_change, value, segment_position) \n" +
+            "VALUES(:examId, :id, :segmentKey, :type, :code, :description, :allowChange, :value, :segmentPosition)";
 
-        accommodations.forEach(examAccommodation -> {
-            SqlParameterSource parameters = new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examAccommodation.getExamId()))
+        SqlParameterSource[] parameters = accommodations.stream().map(examAccommodation ->
+            new MapSqlParameterSource("examId", UuidAdapter.getBytesFromUUID(examAccommodation.getExamId()))
```
millerts: Can we also change Exam to use a character?  Then we can remove this UuidAdapter for examId.

service/src/main/java/tds/exam/models/ExamPage.java
jeffjohnson9046: I certainly haven't been consistent about this either; but it would be nice to have a convention for this.  I agree w/Todd's comment.

jtreuting: I agree with this at a high-level.  I like info provided in the constructor and agree that the `fromX()` methods should always be static.

I think in practice there are questions about what should be part of the constructor but that is probably a bigger conversation.  For this case I think the ID is good to provide there.

PR #61, opened 2017-02-06
Feature/save responses

jeffjohnson9046:
[TDS-501](https://jira.fairwaytech.com/browse/TDS-501): Save responses (no scoring)

### Notes
* Scoring has been stubbed out; an "empty" score is returned by the `ExamItemResponseScoringService` implementation.
  * The empty score that is returned emulates the score that is created by the legacy student application when a response should be scored asynchronously (`ItemScoringService#updateResponses(), line 287)
* The `ExamPage` returned by `ExamItemServiceImpl#insertResponses()` does not contain the next batch of `ExamItem`s; the functionality to determine the `ExamItem`s that should appear on an `ExamPage` is beyond the scope of TDS-501.  A separate JIRA story will be drafted to implement that feature.


service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
```diff
@@ -0,0 +1,49 @@
...
+        this.examItemService = examItemService;
+    }
+
+    @RequestMapping(value = "/{id}/page/{position}/responses", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<Response<ExamPage>> insertResponses(@PathVariable final UUID id,
+                                                       @PathVariable final int position,
+                                                       @RequestParam final UUID sessionId,
+                                                       @RequestParam final UUID browserId,
+                                                       @RequestBody final ExamItemResponse[] responses) {
+        ApprovalRequest approvalRequest = new ApprovalRequest(id, sessionId, browserId);
```
millerts: What do you think about renaming the `ApprovalRequest` class?  I know it is used for approval in the internals but it seems like a better name would be `ExamInfo` since it seems to contain the basic information to get an exam.

service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        this.examPageCommandRepository = examPageCommandRepository;
+        this.examPageQueryRepository = examPageQueryRepository;
+        this.examQueryRepository = examQueryRepository;
+        this.examApprovalService = examApprovalService;
+        this.examItemResponseScoringService = examItemResponseScoringService;
+    }
+
+    @Override
+    public Response<ExamPage> insertResponses(final ApprovalRequest request, final int mostRecentPagePosition, final ExamItemResponse... responses) {
+        Response<ExamApproval> approval = examApprovalService.getApproval(request);
```
millerts: Both this and line 63 fetch an exam from the database.  Can we add an approval method that takes in an `Exam` if we've already fetched it?

service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            return new ExamItemResponse.Builder()
+                .fromExamItemResponse(response)
+                .withScore(score)
+                .build();
+        }).toArray(ExamItemResponse[]::new);
+
+        examItemCommandRepository.insertResponses(scoredResponses);
+
+        // Create a record for the next page
+        ExamPage currentPage = examPageQueryRepository.find(request.getExamId(), mostRecentPagePosition)
```
millerts: Should this be up higher in the flow?   Otherwise we're saving responses for a page that doesn't exist.

service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
```diff
@@ -0,0 +1,49 @@
...
+@RequestMapping("/exam")
+public class ExamItemController {
+    private final ExamItemService examItemService;
+
+    @Autowired
+    public ExamItemController(ExamItemService examItemService) {
+        this.examItemService = examItemService;
+    }
+
+    @RequestMapping(value = "/{id}/page/{position}/responses", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: What do you think about using these [annotations](https://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-ann-requestmapping-composed) rather than the `@RequestMapping` annotations?

client/src/main/java/tds/exam/ExamItemResponseScore.java
```diff
@@ -0,0 +1,132 @@
+package tds.exam;
+
+import com.google.common.base.Optional;
+import org.joda.time.Instant;
+
+/**
+ * Score and scoring metadata for the {@link tds.exam.ExamItemResponse}
+ */
+public class ExamItemResponseScore {
```
millerts: I don't think we need to spend the cycle to convert this but maybe next time we add a new class we can check out using `@AutoValue`

service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
jeffjohnson9046: Actually, there already is a method that will take in an `Exam`:  `ExamApproval#verifyAccess`.  I'll change to use that method instead.

service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
jtreuting: I personally like them, they feel easier to see at a glance what it is.  I would vote for using them moving forward.  But @millerts didn't you say you ran into an issue somewhere with them?  Maybe testing related?

jeffjohnson9046: Sure - I can get behind using the `@PostMapping` annotation here.  I thought there was some issue w/the Spring MVC testing framework and using these annotations, which is why I've been hesitant to use these.  However, the integration tests and unit test for this controller worked for this controller.

service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
jeffjohnson9046: `ApprovalRequest` has always kinda bothered me.  I've made the switch to `ExamInfo`, which feels better.

client/src/main/java/tds/exam/ExamItemResponseScore.java
```diff
@@ -0,0 +1,132 @@
...
+        return Optional.fromNullable(scoredAt);
+    }
+
+    /**
+     * @return A string representing an XML fragment for this {@link tds.exam.ExamItemResponse}'s scoring dimensions.
+     */
+    public String getScoringDimensionsXml() {
+        // Output of StudentDLL.buildSCoreInfoNode(), called on line 1907.  When StudentDLL.T_UpdateScoredResponse_SP is
+        // called (line 1907), the scoreDimension attribute is always set to "overall"
+        return String.format("<ScoreInfo scorePoint=\"%d\" scoreDimension=\"overall\" scoreStatus=\"%s\"><SubScoreList /></ScoreInfo>", score, scoringStatus);
```
jeffjohnson9046: From what I can tell, this _is_ the easy way to construct it on the way out.  I suppose it doesn't have to be a `String.format`, but that seemed like the easiest way to build the score dimensions value.  I initially thought about just doing this in the `getScoreDimensions()` getter, but I'm not sure if something later on in the scoring pipeline changes/modifies the score dimensions value.

My assumption is that as we move forward with implementing scoring, this will likely change.

millerts: Why store the XML in the DB?  It seems we just need to have `scorePoint` and `scoreStatus`.  This getter can remain for now.

jtreuting: ^ This is what I was saying.  Seemed like when we need to the XML we can construct it instead of storing that in the DB.  Maybe we don't have enough info now to know if that is the case, but I was asking what Todd is asking, just not as clearly.

jeffjohnson9046: My (admittedly naive) assumption is that the legacy Student application stores the XML string in the database because some other system (e.g. the item scoring logic) reads it and can potentially change it.

@jtreuting :  I agree that right now I don't have enough information to know if the XML can be constructed as opposed to read from the database.  Since this is a partial implementation of saving responses (no scoring is applied; a "dummy" score is generated prior to inserting the responses to the database), it's entirely possible that the XML string will never change and can be generated some other way.  I suspect we'll find that out as we move forward with implementing scoring.

PR #62, opened 2017-02-06
Bugfix/build ur iproperly

millerts:
This was a tricky one to track down.  It came down to the way we were building our URLs.  We were building the URLs with `toUriString` which RestTemplate would encode again resulting in the wrong assessment key string.

Moved all services to pass in the URI into the RestTemplate rather than rely on it to do the encoding.

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -57,10 +57,6 @@
         <!-- Spring -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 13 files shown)


service/pom.xml
```diff
@@ -57,10 +57,6 @@
         <!-- Spring -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
```
millerts: This isn't needed because the hateoas brings it in.

millerts:
This PR also has the necessary changes to build with the changes in https://github.com/SmarterApp/TDS_StudentService/pull/14

PR #64, opened 2017-02-08
Bugfix/use session proctor

millerts:
The proctor id being passed in was a mistaken idea by me during early development.  We can now get that from the session which is present even for guests.  The proctor id in that situation is just null.


millerts:
FYI - if there is a proctor when creating a new exam the status will be `pending` instead of the flip/flop @mtrupkin reported ealrier today

PR #65, opened 2017-02-09
Feature/format system messages

jtreuting:
This creates a method for formatting localized messages from the DB.  It includes getting the message from the ConfigService endpoint and then doing the replacements, if any.

ValidationErrors from the legacy app that were included in the ExamService have been updated.  Some messages appear to be new and not in the database, and for those I've left them returning the message as it currently was.  We will want to consider if we should add these to the message translations tables in the future.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ConfigServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ConfigServiceImpl.java
@@ -25,6 +25,8 @@ import static tds.exam.configuration.SupportApplicationConfiguration.CONFIG_APP_
  */
 @Service
 class ConfigServiceImpl implements ConfigService {
+    private static final String DEFAULT_LANGUAGE = "ENU";
+
     private final RestTemplate restTemplate;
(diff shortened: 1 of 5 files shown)


service/src/main/java/tds/exam/services/impl/ConfigServiceImpl.java
```diff
@@ -54,4 +59,50 @@ public ConfigServiceImpl(RestTemplate restTemplate, ExamServiceProperties examSe
...
+                    clientName,
+                    context,
+                    messageKey,
+                    languageCode))
+                .queryParam("grade", grade)
+                .queryParam("subject", subject);
+
+        String messageTemplate;
+        try {
+            messageTemplate = restTemplate.getForObject(builder.buildAndExpand().toUri(), String.class);
```
millerts: The buildAndExpand isn't necessary because we are building the URI above. The build and expand method is forcwhen you want to pass in variables for URI construction. You can just do build here.

PR #66, opened 2017-02-11
Feature/student attrs

jeffjohnson9046:
[TDS-665](https://jira.fairwaytech.com/browse/TDS-665):  Save student attributes and relationships to the `exam` database.  This PR effectively ports `CommonDLL._SetTesteeAttributes_SP()` (`CommonDLL`, line 1185).

#### Notes
* Related to the [feature/student-attrs](https://github.com/SmarterApp/TDS_StudentService/pull/15) pull request in the *TDS_StudentService* repo
* So far, the `ExamineeService` implementation is used internally; there is no need for a controller endpoint
* The `ExamineeQueryRepository` implementation has been provided primarily to facilitate testing

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamineeQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamineeQueryRepositoryImpl.java
@@ -30,17 +30,32 @@ public class ExamineeQueryRepositoryImpl implements ExamineeQueryRepository {
 
         final String SQL =
             "SELECT \n" +
-                "   id, \n" +
-                "   exam_id, \n" +
-                "   context, \n" +
-                "   attribute_name, \n" +
-                "   attribute_value, \n" +
-                "   created_at \n" +
+                "   attribute.id, \n" +
+                "   attribute.exam_id, \n" +
+                "   attribute.context, \n" +
+                "   attribute.attribute_name, \n" +
+                "   attribute.attribute_value, \n" +
+                "   attribute.created_at \n" +
                 "FROM \n" +
-                "   examinee_attribute \n" +
+                "   examinee_attribute attribute \n" +
+                "JOIN ( \n" +
+                "   SELECT \n" +
+                "       MAX(id) AS id \n" +
+                "   FROM \n" +
+                "       examinee_attribute \n" +
+                "   WHERE \n" +
+                "       exam_id = :examId \n" +
+                "       AND context = :context \n" +
+                "   GROUP BY \n" +
+                "       exam_id, \n" +
+                "       context, \n" +
+                "       attribute_name \n" +
+                ") most_recent_record \n" +
+                "   ON \n" +
+                "       most_recent_record.id = attribute.id \n" +
                 "WHERE \n" +
-                "   exam_id = :examId \n" +
-                "   AND context = :context";
+                "   attribute.exam_id = :examId	\n" +
+                "   AND attribute.context = :context";
 
         return jdbcTemplate.query(SQL, parameters, (rs, r) -> new ExamineeAttribute.Builder()
             .withId(rs.getLong("id"))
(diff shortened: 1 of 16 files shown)


service/src/main/java/tds/exam/services/impl/ExamineeServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    @Transactional
+    public void insertAttributesAndRelationships(UUID examId, ExamineeContext context) {
+        Exam exam = examService.findExam(examId)
+            .orElseThrow(() -> new NotFoundException(String.format("Could not find exam for id %s", examId)));
+
+        Student student = studentService.getStudentById(exam.getStudentId())
+            .orElseThrow(() -> (new NotFoundException(String.format("Could not find student for exam id %s and student id %d", examId, exam.getStudentId()))));
```
jtreuting: No need for the parenthesis around the new NotFoundException.  Also is consistent with the previous call on line 40

service/src/main/resources/db/migration/V1486592454__exam_create_examinee_tables.sql
```diff
@@ -0,0 +1,30 @@
...
+***********************************************************************************************************************/
+USE exam;
+
+CREATE TABLE IF NOT EXISTS examinee_attribute (
+  id BIGINT(20) NOT NULL AUTO_INCREMENT,
+  exam_id CHAR(36) NOT NULL,
+  context VARCHAR(10) NOT NULL,
+  attribute_name VARCHAR(50) NOT NULL,
+  attribute_value VARCHAR(400) NOT NULL,
+  created_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
```
jtreuting: Should be `UTC_TIMESTAMP()`

service/src/main/resources/db/migration/V1486592454__exam_create_examinee_tables.sql
```diff
@@ -0,0 +1,30 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS examinee_relationship (
+  id BIGINT(20) NOT NULL AUTO_INCREMENT,
+  exam_id CHAR(36) NOT NULL,
+  attribute_name VARCHAR(50) NOT NULL,
+  attribute_value VARCHAR(500) NOT NULL,
+  attribute_relationship VARCHAR(100) NOT NULL,
+  context VARCHAR(10) NOT NULL,
+  created_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
```
jtreuting: `UTC_TIMESTAMP`

service/src/test/java/tds/exam/repositories/impl/ExamineeCommandRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,121 @@
...
+    private ExamineeCommandRepository examineeCommandRepository;
+    private ExamineeQueryRepository examineeQueryRepository;
+    private ExamCommandRepository examCommandRepository;
+    private Exam mockExam = new ExamBuilder().build();
+
+    @Before
+    public void setUp() {
+        examineeCommandRepository = new ExamineeCommandRepositoryImpl(jdbcTemplate);
+        examineeQueryRepository = new ExamineeQueryRepositoryImpl(jdbcTemplate);
+
```
jtreuting: extra space

service/src/main/java/tds/exam/repositories/impl/ExamineeQueryRepositoryImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    public ExamineeQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdbcTemplate jdbcTemplate) {
+        this.jdbcTemplate = jdbcTemplate;
+    }
+
+    @Override
+    public List<ExamineeAttribute> findAllAttributes(final UUID examId, final ExamineeContext context) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("examId", examId.toString())
+            .addValue("context", context.toString());
+
+        final String SQL =
```
millerts: FYI, during open exam there is a possibility that new values will be inserted for certain fields.  So this will fetch all attributes including duplicates.  For example, if a student package school changes between opening an exam the school will be twice.

I think both queries in this repository need to do the same fetch latest as we do on other event like tables.

service/src/main/resources/db/migration/V1486592454__exam_create_examinee_tables.sql
jeffjohnson9046: It turns out that you cannot create a default of `UTC_TIMESTAMP()` on a table (the third answer on this [SO post](http://dba.stackexchange.com/questions/20217/mysql-set-utc-time-as-default-timestamp)).  I couldn't find any other migrations in our codebase that use the `UTC_TIMESTAMP()` function as a default.  Also, I think we're enforcing the usage of UTC at the application level.

service/src/main/resources/db/migration/V1486592454__exam_create_examinee_tables.sql
jeffjohnson9046: It turns out that you cannot create a default of `UTC_TIMESTAMP()` on a table (the third answer on this [SO post](http://dba.stackexchange.com/questions/20217/mysql-set-utc-time-as-default-timestamp)).  I couldn't find any other migrations in our codebase that use the `UTC_TIMESTAMP()` function as a default.  Also, I think we're enforcing the usage of UTC at the application level.

service/src/main/java/tds/exam/repositories/impl/ExamineeQueryRepositoryImpl.java
jeffjohnson9046: Agreed.  Right now, this query repository is only used for testing.  But if there's a need to query these values, the most recent attributes/relationships should be fetched.  Queries for both methods have been updated and test coverage added.

service/src/main/resources/db/migration/V1486592454__exam_create_examinee_tables.sql
jtreuting: Interesting.  Yeah we are enforcing in the app, but I was hoping it would default to UTC as well.  Oh well, I assumed that would be easy :)

PR #67, opened 2017-02-11
Feature/exam status update endpoint

burzum619:
This PR corresponds to JIRA TDS-518. This endpoint will be called when student or proctor needs to update a status for an exam.

A new map has been added to ExamStatusCode to map status' to their default stage, if no stage is provided by the caller of this API.


millerts:
Code looks good.  I think the question I'm starting to have is are the statuses and stages really configurable?

PR #68, opened 2017-02-13
Add approve/deny exam endpoints

mtrupkin:


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -158,16 +158,16 @@ public ExamAccommodationServiceImpl(ExamAccommodationQueryRepository examAccommo
         
         if (!maybeSession.isPresent()) {
             return Optional.of(new ValidationError(ValidationErrorCode.EXAM_NOT_ENROLLED_IN_SESSION, "The test opportunity is not enrolled in this session"));
-        } else if (maybeSession.isPresent() && !maybeSession.get().isProctorless()) {
+        } else if (maybeSession.isPresent() && maybeSession.get().isProctorless()) {
```
jtreuting: Seems like the `!` should not have been removed based on the validation error message.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -121,4 +127,31 @@ public ExamController(ExamService examService) {
     public void pauseExamsInSession(@PathVariable final UUID sessionId) {
         examService.pauseAllExamsInSession(sessionId);
     }
+
+    @RequestMapping(value = "/approve/{examId}", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
```
burzum619: I think for URI's we've generally been following the pattern "/exam/{examId}/<verb>" as opposed to "/exam/<verb>/{examId}". I don't feel too strongly about it either way, but perhaps we may want to keep it consistent. For reference see startExam, updateStatus, and pauseExam methods in ExamController.

PR #70, opened 2017-02-14
Feature/open exam save attrs

jeffjohnson9046:
[TDS-672](https://jira.fairwaytech.com/browse/TDS-672): Add a call to `ExamineeService.insertAttributesAndRelationships()` into the `ExamServiceImpl`.  This call emulates lines 409 - 410 in `OpenTestServiceImpl` within the legacy student application.

### DIFF
--- a/service/src/main/java/tds/exam/services/ExamineeService.java
+++ b/service/src/main/java/tds/exam/services/ExamineeService.java
@@ -1,7 +1,5 @@
 package tds.exam.services;
 
-import java.util.UUID;
-
 import tds.exam.Exam;
 import tds.exam.ExamineeContext;
 
@@ -13,8 +11,7 @@ public interface ExamineeService {
      * Insert the {@link tds.exam.ExamineeAttribute}s and {@link tds.exam.ExamineeRelationship}s for this
      * {@link tds.exam.Exam} for the specified {@link tds.exam.ExamineeContext}.
      *
-     * @param examId The unique identifier of the {@link tds.exam.Exam} for which attributes and relationships are being
-     *               stored
+     * @param exam    The {@link tds.exam.Exam} for which attributes and relationships are being stored
      * @param context The {@link tds.exam.ExamineeContext} that describes the state of the examniee's information when
      *                these attributes and relationships were collected
      */
--- a/service/src/main/java/tds/exam/services/ExamineeService.java
+++ b/service/src/main/java/tds/exam/services/ExamineeService.java
@@ -2,6 +2,7 @@ package tds.exam.services;
 
 import java.util.UUID;
 
+import tds.exam.Exam;
 import tds.exam.ExamineeContext;
 
 /**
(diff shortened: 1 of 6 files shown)


service/src/main/java/tds/exam/services/ExamineeService.java
```diff
@@ -17,5 +18,5 @@
      * @param context The {@link tds.exam.ExamineeContext} that describes the state of the examniee's information when
      *                these attributes and relationships were collected
      */
-    void insertAttributesAndRelationships(UUID examId, ExamineeContext context);
+    void insertAttributesAndRelationships(Exam exam, ExamineeContext context);
```
jtreuting: Update javadoc params

service/src/main/java/tds/exam/services/impl/ExamineeServiceImpl.java
```diff
@@ -5,46 +5,37 @@
...
     @Override
     @Transactional
-    public void insertAttributesAndRelationships(UUID examId, ExamineeContext context) {
-        Exam exam = examService.findExam(examId)
-            .orElseThrow(() -> new NotFoundException(String.format("Could not find exam for id %s", examId)));
-
+    public void insertAttributesAndRelationships(Exam exam, ExamineeContext context) {
         Student student = studentService.getStudentById(exam.getStudentId())
-            .orElseThrow(() -> new NotFoundException(String.format("Could not find student for exam id %s and student id %d", examId, exam.getStudentId())));
+            .orElseThrow(() -> new NotFoundException(String.format("Could not find student for exam id %s and student id %d", exam.getId(), exam.getStudentId())));
```
millerts: We don't necessarily need to change this now, but I think this method should take in a Student.  The reason being is that we've already fetched it during [openExam](https://github.com/SmarterApp/TDS_ExamService/blob/develop/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java#L166) before this is called.

If we have a need for an implementation that needs to fetch the Student we could do something like this:

```
public void insertAttributesAndRelationships(Exam exam, ExamineeContext context) {
  Student student = fetch student;
  insertAttributesAndRelationships(exam, context, student)
}

public void insertAttributesAndRelationships(Exam exam, ExamineeContext context, Student student) {
//logic
}

File: service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java [hunk 3]
```diff
@@ -512,6 +517,9 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
 
         examCommandRepository.insert(exam);
 
+        // OpenTestServiceImpl lines 409 - 410
+        examineeService.insertAttributesAndRelationships(exam, ExamineeContext.INITIAL);
```
millerts: We need to confirm this with a breakpoint in the legacy code but I'm pretty sure we only need to call this for students that are not guest students.

millerts:
I posted my comments for clarity.  I'll double check and address them in a separate PR.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jtreuting: Yep you are right, I missed that.

https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/cebc3996ab000dc604b539da51235eaf20039d0d/tds-dll-mysql/src/main/java/tds/dll/mysql/CommonDLL.java#L1185

PR #71, opened 2017-02-15
Bugfix/ignore guest students on examinee attributes

millerts:
There are three changes that rippled.

1. StudentService.findStudentById was being used to fetch the `Student` the URL was incorrect so attributes and relationships were not coming back from that call.  The change to the interface to include client rippled.
2. Added a new ExamineeService method that takes in a student and leverage that in the open exam flow. 
3. examinee_attribute `attribute_value` column needs to be nullable.  Checked against legacy student and even though values are not null in the final state they are when opening an exam.


millerts:
This must be merged with https://github.com/SmarterApp/TDS_StudentService/pull/16

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -518,7 +519,9 @@ private static ExamConfiguration getExamConfiguration(Exam exam, Assessment asse
         examCommandRepository.insert(exam);
 
         // OpenTestServiceImpl lines 409 - 410
-        examineeService.insertAttributesAndRelationships(exam, ExamineeContext.INITIAL);
+        if(!openExamRequest.isGuestStudent()) {
```
jtreuting: Does this check make more sense inside the `examineeService.insertAttributesAndRelationships` based on the `student`?  I know that is how they did it in the legacy code and then if this is called in multiple places we don;t have to worry about wrapping them all.  But maybe that isn;t an issue because it isn't called from more places?

millerts: I will add it to the method that does not take a student but instead fetches the student.  I think that is why the check is necessary in the old code.  If you have a Student object that means it is a real student.

jtreuting: Makes sense

PR #72, opened 2017-02-15
Use the external configuration service for configurations.  Add a loc…

wtritch:
…al-development configuration profile with sample property values.

This PR updates the Exam Service to optionally pull configurations from a Spring Cloud Config service.
* Pull docker configuration values out of the docker start script
* Add default properties to an application.yml file under a "local-development" profile
* Add bootstrap configuration properties that will connect to a remote configuration service if CONFIG_SERVICE_ENABLED and CONFIG_SERVICE_URL environment properties are set.

There is a [related pull request](https://github.com/SmarterApp/TDS_Build/pull/3) for TDS_Build.

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -1,0 +1,39 @@
+spring:
+  profiles: local-development
+  ds_queries:
+    jdbcUrl: ${mysql.url}
+    username: ${mysql.user}
+    password: ${mysql.password}
+    driver-class-name: 'com.mysql.jdbc.Driver'
+  ds_commands:
+    jdbcUrl: ${mysql.url}
+    username: ${mysql.user}
+    password: ${mysql.password}
+    driver-class-name: 'com.mysql.jdbc.Driver'
+
+server:
+  port: 8081
+  undertow:
+    buffer-size: 16384
+    buffers-per-region: 20
+    io-threads: 64
+    worker-threads: 512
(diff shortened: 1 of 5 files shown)


service/src/main/resources/application.yml
```diff
@@ -0,0 +1,39 @@
...
+exam-service:
+  session-url: http://localhost:32842
+  student-url: http://localhost:32840
+  assessment-url: http://localhost:32841
+  config-url: http://localhost:32843
+
+mysql:
+  url: jdbc:mysql://localhost:3306/exam
+  user: username
+  password: password
```
wtritch: The mysql.url, mysql.user, and mysql.password properties will need to be updated to real values before the "local-development" profile can be used.  However, if I remove them from this application.yml file, the system will pull MYSQL_URL, MYSQL_USER, and MYSQL_PASSWORD environment variables.
Should I remove them from the application.yml? What do you all think?

millerts: So one problem that we had was that developers kept committing and then including their application.properties  which is why it is included in our .gitignore.  I don't think it is something that would stop this PR.

jtreuting: To Todd's point, it seems like we should remove them so that we don't have that issue movng forward with checking in the changes by accident.

wtritch: Sounds like a plan, removed.

PR #73, opened 2017-02-16
Now updating exam language during openTest and checkApproval flow

burzum619:
Now updating the Exam's languageCode property in open test and potentially in check approval, if the exam language differs from the default for the assessment.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
@@ -16,13 +16,14 @@ import java.util.stream.Collectors;
 import tds.accommodation.Accommodation;
 import tds.assessment.Assessment;
 import tds.common.ValidationError;
-import tds.exam.ExamInfo;
 import tds.exam.ApproveAccommodationsRequest;
 import tds.exam.Exam;
 import tds.exam.ExamAccommodation;
+import tds.exam.ExamInfo;
 import tds.exam.error.ValidationErrorCode;
 import tds.exam.repositories.ExamAccommodationCommandRepository;
 import tds.exam.repositories.ExamAccommodationQueryRepository;
+import tds.exam.repositories.ExamCommandRepository;
 import tds.exam.repositories.ExamQueryRepository;
 import tds.exam.services.AssessmentService;
 import tds.exam.services.ExamAccommodationService;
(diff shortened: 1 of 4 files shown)


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -341,4 +346,26 @@ private static boolean isEqual(Instant instant, Instant instant2) {
             && instant != null
             && instant.equals(instant2);
     }
+    
+    private Exam updateExamWithLanguageCode(final Exam exam, final Set<ExamAccommodation> examAccommodations) {
```
millerts: I think we should discuss this change.  The reason ExamAccommodationService has to update the Exam is because we have chosen to store a `languageCode` on the `exam_event` database table.  Due to the way we're doing the eventing tables we chose to copy the data rather than have a complex query to get at the data.

Normally I would either get all the exam accommodations when getting the exam (like an ORM) allowing us to get the language or do a join.  Due to the complexity of our queries due to the eventing tables we've shied away from that style.

I think we should discuss this and decide if we want to continue to copy data to different tables or not.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -747,4 +748,22 @@ private Exam updateExamWithCustomAccommodations(final Exam exam, final List<Exam
 
         return exam;
     }
+    
+    private Exam updateExamWithLanguageCode(final Exam exam, final List<ExamAccommodation> examAccommodations) {
+        Optional<ExamAccommodation> maybeExamAccommodation = examAccommodations.stream()
+          .filter(accommodation -> accommodation.getType().equals(Accommodation.ACCOMMODATION_TYPE_LANGUAGE))
+          .findFirst();
+        
+        if (maybeExamAccommodation.isPresent()) {
```
millerts: Will this ever not be present?  And if it isn't present should it result in an exception?

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
jeffjohnson9046: Well, we could reconsider how we're fetching the exam from the database.  It seems like a fair amount of workflow is driven by the accommodations, in particular the language code.  Would it be worthwhile to explore having an `ExamAssembler` that would give us back a richer `Exam` that included its accommodations?

I'm not intimately familiar with the workflow/pipeline, so there might be some timing issues or other complications with fetching the exam's associated accommodations.

PR #75, opened 2017-02-17
Exam repository query will now join on exam_accommodations to retriev…

burzum619:
…e language accommodation

I needed to insert an ExamAccommodation during ExamQuery/Command integration tests so that the "find" queries would not fail. an Exam should __always__ have an exam language at the assessment-level (segment_position = 0) by the time it is access with a query repository.

Note that the "segment_position = 0" is to ensure that we only retrieve one accommodation record. 
TDS-660 is a an open bug where exam accommodations are inserted at position "1" instead of "0" during open test flow. Once this is addressed, the segment_position = 0 clause is extraneous and can be removed. 

https://jira.fairwaytech.com/browse/TDS-660?filter=-2

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -53,12 +53,12 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
         "e.assessment_window_id, \n" +
         "e.assessment_algorithm, \n" +
         "e.segmented, \n" +
+        "lang.code AS language_code, \n" +
         "ee.attempts, \n" +
         "ee.status, \n" +
         "ee.status_changed_at, \n" +
         "ee.max_items, \n" +
         "ee.expires_at, \n" +
-        "ee.language_code, \n" +
         "ee.status_change_reason, \n" +
         "ee.deleted_at, \n" +
         "ee.changed_at, \n" +
@@ -101,7 +101,12 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
                 "  ON last_event.exam_id = ee.exam_id AND \n" +
                 "     last_event.id = ee.id\n" +
                 "JOIN exam.exam_status_codes esc \n" +
-                "  ON esc.status = ee.status";
+                "  ON esc.status = ee.status \n" +
+                "JOIN exam.exam_accommodation lang \n" +
+                "  ON lang.exam_id= e.id \n" +
+                "WHERE \n" +
+                "   lang.type = 'Language' AND \n" +
+                "   lang.segment_position = 0";
 
         Optional<Exam> examOptional;
         try {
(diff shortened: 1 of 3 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -101,7 +101,12 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 "  ON last_event.exam_id = ee.exam_id AND \n" +
                 "     last_event.id = ee.id\n" +
                 "JOIN exam.exam_status_codes esc \n" +
-                "  ON esc.status = ee.status";
+                "  ON esc.status = ee.status \n" +
+                "JOIN exam.exam_accommodation lang \n" +
```
millerts: Doesn't this need to join on the `exam_accommodation_event` table?  For example, the language is ENU by default which is the initial language inserted into the exam_event.  The proctor then comes in and sets the language to FRN.  Wouldn't that result in two language ExamAccommodations in the exam with one of them with an `exam_accommodation_event` record populated `deleted_at`?

jtreuting: Since there will always be at least 1 language (I assume that is correct), I don't think we need to join on the event table, but we would need to get the latest exam_accommodation where the type is Language since the last record won't ever be deleted for language.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -101,7 +101,12 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
+                "  ON lang.exam_id= e.id \n" +
+                "WHERE \n" +
+                "   lang.type = 'Language' AND \n" +
+                "   lang.segment_position = 0";
```
jtreuting: For consistency, let's put the AND at the front of the second like like the SQL changes you did below.  I think that is pretty much our standard.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -324,9 +337,13 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
                 "JOIN exam.exam_status_codes esc \n" +
                 "  ON esc.status = ee.status \n" +
+                "JOIN exam.exam_accommodation lang \n" +
+                "  ON lang.exam_id= e.id \n" +
                 "WHERE \n" +
                 "  ee.session_id = :sessionId AND \n" +
-                "  ee.status IN (:statusSet)";
+                "  ee.status IN (:statusSet) AND \n" +
+                "  lang.type = 'Language' AND \n" +
+                "  lang.segment_position = 0";
```
jtreuting: Same AND comment here.  Plus, it probably goes without saying, whatever changes to the SQL made above will need to be made here as well.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
millerts: Why do we have an exam_accommodation_event table then?  Since the exam_accommodations as whole can change.  Seems like we should chuck the event table and treat exam accommodation as the event table.  Like we do responses

millerts: I don't disagree with your comment and nothing needs to change as part of this PR.  It is just that I don't see what the event table is buying us.

burzum619: exam_accommodations can be "deleted" I believe.

burzum619: I agree with @jtreuting , I'll order by and select the latest.

burzum619: There's also a "denied_at" field, although I'm not sure how its used. Can proctors reject exam_accommodations that have already been inserted??

millerts: I think we should table this and not have this discussion as part of this PR.

PR #77, opened 2017-02-17
Feature/legacy db views

jeffjohnson9046:
[TDS-650](https://jira.fairwaytech.com/browse/TDS-650):  Add views to the `exam` database that make the `exam` tables look like their equivalent tables in the legacy `session` database.

### Notes
* These views are not (and should not be) leveraged by any application; they are primarily to assist the QA effort of verifying that data stored by the exam microservice is equivalent to what the legacy student application stored in its `session` database.
* In many cases, columns were not migrated from the `session` database to the `exam` database.  In that case, the view will return a value of `'not migrated'` for that column.
* As features are added to the exam microservice and/or integration work is done (e.g. scoring logic implementation or integration with saving responses) that require additional database columns, the idea is that these views will be updated if appropriate via a flyway migration.

### DIFF
--- a/service/src/main/resources/db/migration/V1487352270__exam_create_legacy_db_views.sql
+++ b/service/src/main/resources/db/migration/V1487352270__exam_create_legacy_db_views.sql
@@ -206,13 +206,13 @@ CREATE OR REPLACE VIEW qa_session_testeeaccommodations AS
 -- ----------------------------------------------------------------------------
 CREATE OR REPLACE VIEW qa_session_testeeresponse AS
 	SELECT
-		page.exam_id AS _fk_testopportunity,
-		item.assessment_item_key AS _efk_itsitem,
-		item.assessment_item_bank_key AS _efk_itsbank,
-		exam.session_id AS _fk_session,
-		exam.attempts AS opportunityrestart,
-		page.page_position AS page,
-		item.position AS position,
+	  page.exam_id AS _fk_testopportunity,
+	  item.assessment_item_key AS _efk_itsitem,
+	  item.assessment_item_bank_key AS _efk_itsbank,
+	  exam.session_id AS _fk_session,
+	  exam.attempts AS opportunityrestart,
+	  page.page_position AS page,
+	  item.position AS position,
 		'not migrated' AS answer,
 		'not migrated' AS scorepoint,
 		'not migrated' AS format,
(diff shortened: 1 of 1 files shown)


service/src/main/resources/db/migration/V1487352270__exam_create_legacy_db_views.sql
```diff
@@ -0,0 +1,352 @@
...
+		'not migrated' AS testeecontrol,
+		ea.exam_id AS _fk_testopportunity,
+		CASE
+			WHEN eae.deleted_at IS NULL THEN 1
+			ELSE 0
+		END AS isapproved,
+		eae.selectable AS isselectable,
+		ea.segment_position AS segment,
+		'not migrated' AS valuecount,
+		'not migrated' AS recordusage
```
wtritch: super-nitpick, indentations seems to be a little varied.

service/src/main/resources/db/migration/V1487352270__exam_create_legacy_db_views.sql
```diff
@@ -0,0 +1,352 @@
...
+  FROM
+    exam_page_event
+  GROUP BY
+    exam_page_id;
+
+-- ----------------------------------------------------------------------------
+-- A view for finding the latest exam_item_response record for an exam_item
+--
+-- NOTE:  This view is required because MySQL 5.6 does not support using
+-- subqueries in the FROM clause of a view:  "Before MySQL 5.7.7, the SELECT
```
jtreuting: Probably too late, but AWS RDS does have version 5.7.11 and 5.7.16 that we could upgrade to.  For some reason it defaults 5.6.27

jeffjohnson9046: Nope - not too late (as far as I'm concerned, anyway).  If we've agreed that 5.7.x is the MySQL version we're going to support, then I can change the views.  Although it might make more sense to leave them this way:

* I think they're a little bit more readable
* Some other implementor might _not_ want to go with MySQL 5.7.x for whatever reason.  Leaving the views without subqueries provides more backwards compatibility for older MySQL versions.

jtreuting: I think the MySQL support level is 2 separate questions.  Since this view is not used for the application itself, we can have it in our environment even if we say that the application itself supports MySQL 5.6 (which I believe should be a separate question, but I would lean towards it's fine to choose anything available in RDS)

PR #79, opened 2017-02-18
Feature/add transactions

millerts:
I was a little overzealous when adding the Transactional annotations.  The ones that are removed here are services that simply do batch inserts to a particular single table.  This is fine as long as we allow the database to run internal transactions so we get the ACID features one batch inserts.

https://dev.mysql.com/doc/refman/5.7/en/set-transaction.html


millerts:
After some external conversation we've decided to hold off on removing the transactions.

PR #81, opened 2017-02-20
Feature/get exam segments 

burzum619:
This endpoint is to support the  completion of /startExam integration with legacy student. This endpoint will be fetched by the RemoteOpportunityService.getSegments() method, which is called in MasterShellHandler during the /startExam flow. Another PR in TDS_Student will be created to cover the code that will call this API.


service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -69,11 +71,21 @@
      * to the new status; otherwise {@code Optional.empty()}.\
      */
     Optional<ValidationError> updateExamStatus(final UUID examId, final ExamStatusCode newStatus);
-
+
```
millerts: These are odd to me.  It is like the formatter is adding a tab of 4 spaces to any newline.

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -28,16 +31,19 @@
...
     private final SegmentPoolService segmentPoolService;
     private final FormSelector formSelector;
     private final FieldTestService fieldTestService;
 
     @Autowired
-    public ExamSegmentServiceImpl(final ExamSegmentCommandRepository commandRepository,
-                                  final SegmentPoolService segmentPoolService,
-                                  final FormSelector formSelector,
-                                  final FieldTestService fieldTestService) {
+    public ExamSegmentServiceImpl(ExamSegmentCommandRepository commandRepository,
```
millerts: Can you add the `final` keyword back to all of these?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -768,4 +769,29 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
 
         return exam;
     }
+    
+    @Override
+    @Transactional
```
millerts: We shouldn't need to use Transactional for methods that only do lookups.  We can save transactions for when we impact multiple records in the DB.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -768,4 +769,29 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
...
+    public Response<List<ExamSegment>> findExamSegments(final UUID examId, final UUID sessionId, final UUID browserId) {
```
millerts: Why did you decide to put this in ExamService rather than ExamSegmentService?

client/src/main/java/tds/exam/ExamSegment.java
```diff
@@ -114,8 +116,8 @@ public Builder withFieldTestItemCount(int newFieldTestItemCount) {
             return this;
         }
 
-        public Builder withIsPermeable(boolean newIsPermeable) {
-            this.isPermeable = newIsPermeable;
+        public Builder withIsPermeable(boolean newPermeable) {
+            this.permeable = newPermeable;
```
wtritch: super-nit, with boolean values I like just withPermeable =)

client/src/main/java/tds/exam/ExamSegment.java
```diff
@@ -21,14 +21,16 @@
...
-    private boolean isSatisfied;
+    private boolean satisfied;
     private Instant exitedAt;
     private Set<String> itemPool;
     private int poolCount;
     private Instant createdAt;
+    
+    private ExamSegment() {}
 
     public ExamSegment(Builder builder) {
```
wtritch: Interesting.  Usually I implement a builder with a .build() method rather than a model constructor that accepts a Builder...

service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -69,11 +71,21 @@
...
+    
     /**
      * Update the status of all {@link tds.exam.Exam}s in the specified {@link tds.session.Session} to "paused"
      *
      * @param sessionId The unique identifier of the session that has been closed
      */
     void pauseAllExamsInSession(final UUID sessionId);
+    
+    /**
+     * Fetches the {@link tds.exam.ExamSegment}s for the exam id after validating the exam and session.
```
wtritch: Does this method perform validation?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -768,4 +769,29 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
...
+        /* This method is a port of the legacy OpportunityRepository.getOpportunitySegments() [241] and
+        *  StudentDLL.T_GetOpportunitySegments_SP [10212]*/
+        ExamInfo examInfo = new ExamInfo(examId, sessionId, browserId);
+        Optional<Exam> maybeExam = findExam(examId);
+        
+        if (!maybeExam.isPresent()) {
```
millerts: I understand you're following a pattern that I believe I started but I think we should remove this check.  Instead I think this flow should call `ExamApprovalService.checkApproval(ExamInfo)` which underneath the hood does the fetch and verify exam flow.

client/src/main/java/tds/exam/ExamSegment.java
millerts: So you were thinking something like this in the Builder.build() method?

```
        public ExamSegment build() {
            ExamSegment segment = new ExamSegment();
            segment.sessionId = this.sessionId;
            return request;
        }```

millerts: I'm not opposed to it.  I probably added the constructor takes a builder pattern.  Not sure there is much difference though between the two since they are all private.

wtritch: Pretty much.  I just like my models to be true POJOs, this way the builder needs getter methods and the model constructor needs a list of all properties.

millerts: The builder doesn't need getter methods.  Since the builder is within the class they can access each others properties.

millerts: And I pulled this pattern from effective Java.  Not saying that it makes it correct.  

http://www.informit.com/articles/article.aspx?p=1216151&seqNum=2

service/src/main/java/tds/exam/services/ExamService.java
burzum619: Yes it does, it calls ExamService.verifyAccess which performs validation on the Exam.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Yep, I thought about that after the fact, thanks

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: It needs to call verifyAccess, which is in ExamService, and I also need to fetch the Exam object. If it was in ExamSegmentService, I would not be able to add the ExamService as a dependency or it would cause a circular dependency.

PR #83, opened 2017-02-22
Expandable exams  + findExamsBySession service code

burzum619:
[TDS-709]  - This PR covers code for the ExpandableExam, as well as a service method populating some additional pieces of data needed by the proctor application, such as the exam accommodations and the number of items a student has answered (responded to). 

Included in this PR are changes to some ExamAccommodationService/Repository methods that previously took in a single UUID (exam id) as an argument. These have been expanded to take in a variable number of exam id arguments in order to reduce the number of jdbc/database calls. 

TDS-646, which depends on this PR, covers the query found in TDS_Proctor - TestSessionDaoImpl.getCurrentSessionTestees() [line 88]. ExamService.findExamsBySessionId() is essentially a port of this query, with some unused fields removed. That query, formatter and with some of my personal comments is found below:

```
select 
  tOpp._efk_AdminSubject,                         --exam.assessmentKey
  tOpp._fk_TestOpportunity as opportunityKey,     --exam.id
  tOpp._efk_Testee,                               --exam.studentId
  tOpp._efk_TestID,                               --exam.assessmentId
  tOpp.Opportunity,                               --exam.attempts
  tOpp.TesteeName,                                --exam.studentName
  TesteeID,                                       --exam.loginSsid  
  tOpp.Status,                                    --exam.status
  tOpp.DateCompleted,                             --exam.completedAt
  tOpp._fk_Session,                               --exam.sessionId    (uuid)
  tOpp.SessID as SessionID,                       --exam.sessionKey  (ADM-99)
  '' as sessionName,  
  maxitems as ItemCount,                          --exam.maxItems
  case when tOpp.status = 'paused' and datePaused is not null 
    then timestampdiff(MINUTE, datePaused, now(3)) 
  else 
    cast(null as CHAR) end as pauseMinutes,       -- null if not paused, otherwise the number of mins the exam has been paused for
  numResponses as ResponseCount,                   
  (
    select 
      count(*) 
    from 
      testopprequest REQ 
    where 
      REQ._fk_TestOpportunity = tOpp._fk_TestOpportunity 
      and REQ._fk_Session = ${sessionKey}
      and DateFulfilled is null 
      and DateSubmitted > ${midnightAM} 
      and DateSubmitted < ${midnightPM}
  ) as RequestCountN,                             -- the number of unfulfilled requests for this exam/session          
  (
    select 
      value as score 
    from 
      testopportunityscores S
    join configs.client_testscorefeatures F
      on S.MeasureOf = F.MeasureOf 
      and S.MeasureLabel = F.MeasureLabel 
    where 
      F.ClientName = ${clientname} 
      and ReportToProctor = 1 
      and S._fk_TestOpportunity = tOpp._fk_TestOpportunity 
      and S.IsOfficial = 1 
    limit 1
  ) as Score,                                 -- not needed
  AccommodationString as Accommodations,      -- ExamService will return entire list of exam accommodations
  tOpp.customAccommodations,                  -- exam.customAccommodations
  tOpp.mode,                                  -- Not needed, always "online"
  ctp.msb                                     -- TODO: Need to fetch MSB flag from assessment
from testopportunity_readonly tOpp 
left outer join                       -- Join not necessary - only used to validate testid and clientname (part of exam) and to get msb flag. 
  configs.client_testproperties ctp 
  on tOpp._efk_testid = ctp.testid 
  and tOpp.clientname = ctp.clientname 
where 
  _fk_Session = ${sessionKey} 
  and tOpp.DateChanged > ${midnightAM}
  and tOpp.DateChanged < ${midnightPM} 
  and tOpp.status not in ('pending', 'suspended', 'denied');
```

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
@@ -13,6 +13,7 @@ import java.sql.SQLException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import tds.exam.ExamAccommodation;
 import tds.exam.repositories.ExamAccommodationQueryRepository;
@@ -79,17 +80,23 @@ public class ExamAccommodationQueryRepositoryImpl implements ExamAccommodationQu
     }
 
     @Override
-    public List<ExamAccommodation> findAccommodations(final UUID examId) {
-        return getAccommodations(examId, false);
+    public List<ExamAccommodation> findAccommodations(final UUID... examIds) {
+        return getAccommodations(false, examIds);
     }
 
     @Override
-    public List<ExamAccommodation> findApprovedAccommodations(final UUID examId) {
-        return getAccommodations(examId, true);
+    public List<ExamAccommodation> findApprovedAccommodations(final UUID... examIds) {
+        return getAccommodations(true, examIds);
     }
 
-    private List<ExamAccommodation> getAccommodations(final UUID examId, final boolean excludeDenied) {
-        final SqlParameterSource parameters = new MapSqlParameterSource("examId", examId.toString());
+    private List<ExamAccommodation> getAccommodations(final boolean excludeDenied, final UUID... examIds) {
+        final SqlParameterSource parameters = (examIds.length > 1)
+            ? new MapSqlParameterSource("examIds", Arrays.stream(examIds).map(UUID::toString).collect(Collectors.toSet()))
+            : new MapSqlParameterSource("examId", examIds[0].toString());
+
+        final String WHERE_CLAUSE = examIds.length > 1
+            ? "ea.exam_id IN (:examIds) \n"
+            : "ea.exam_id = :examId \n";
 
         String SQL =
             "SELECT \n" +
@@ -121,7 +128,7 @@ public class ExamAccommodationQueryRepositoryImpl implements ExamAccommodationQu
                 "JOIN exam_accommodation_event eae \n" +
                 "  ON last_event.id = eae.id \n" +
                 "WHERE \n" +
-                "   ea.exam_id = :examId" +
+                WHERE_CLAUSE +
                 "   AND eae.deleted_at IS NULL";
 
         if (excludeDenied) {
--- a/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
@@ -90,13 +90,8 @@ public class ExamAccommodationQueryRepositoryImpl implements ExamAccommodationQu
     }
 
     private List<ExamAccommodation> getAccommodations(final boolean excludeDenied, final UUID... examIds) {
-        final SqlParameterSource parameters = (examIds.length > 1)
-            ? new MapSqlParameterSource("examIds", Arrays.stream(examIds).map(UUID::toString).collect(Collectors.toSet()))
-            : new MapSqlParameterSource("examId", examIds[0].toString());
-
-        final String WHERE_CLAUSE = examIds.length > 1
-            ? "ea.exam_id IN (:examIds) \n"
-            : "ea.exam_id = :examId \n";
+        final SqlParameterSource parameters = new MapSqlParameterSource("examIds",
+            Arrays.stream(examIds).map(UUID::toString).collect(Collectors.toSet()));
 
         String SQL =
             "SELECT \n" +
@@ -128,7 +123,7 @@ public class ExamAccommodationQueryRepositoryImpl implements ExamAccommodationQu
                 "JOIN exam_accommodation_event eae \n" +
                 "  ON last_event.id = eae.id \n" +
                 "WHERE \n" +
-                WHERE_CLAUSE +
+                "   ea.exam_id IN (:examIds) \n" +
                 "   AND eae.deleted_at IS NULL";
 
         if (excludeDenied) {
--- a/service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
@@ -2,12 +2,21 @@ package tds.exam.repositories.impl;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.dao.DataAccessException;
+import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.stereotype.Repository;
 
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import tds.exam.repositories.ExamItemQueryRepository;
 
@@ -45,5 +54,48 @@ public class ExamItemQueryRepositoryImpl implements ExamItemQueryRepository {
                 "   PE.deleted_at IS NULL;\n";
 
         return jdbcTemplate.queryForObject(SQL, params, Integer.class);
+    }
+
+    @Override
+    public Map<UUID, Integer> getResponseCounts(final UUID... examIds) {
+        // Needs to toString() each UUID individually
+        final SqlParameterSource params = new MapSqlParameterSource("examIds",
+            Arrays.stream(examIds).map(UUID::toString).collect(Collectors.toSet()));
+        final String SQL =
+            "SELECT \n" +
+                "   exam_id, \n" +
+                "   COUNT(id) as response_count \n" +
+                "FROM ( \n" +
+                "   SELECT\n" +
+                "       P.exam_id, \n" +
+                "       I.id\n" +
+                "   FROM \n" +
+                "       exam.exam_item I \n" +
+                "   JOIN \n" +
+                "       exam.exam_item_response IR \n" +
+                "   ON \n" +
+                "       I.id = IR.exam_item_id \n" +
+                "   JOIN \n" +
+                "       exam.exam_page P \n" +
+                "   ON \n" +
+                "       P.id = I.exam_page_id \n" +
+                "   JOIN \n" +
+                "       exam.exam_page_event PE \n" +
+                "   ON \n" +
+                "       PE.exam_page_id = P.id \n" +
+                "   WHERE \n" +
+                "       PE.deleted_at IS NULL \n" +
+                "       AND P.exam_id IN (:examIds) \n" +
+                "   GROUP BY P.exam_id, I.id \n" +
+                ") examIdCounts \n" +
+                "GROUP BY exam_id";
+
+        return jdbcTemplate.query(SQL, params, (ResultSetExtractor<Map<UUID, Integer>>) rs -> {
+            HashMap<UUID, Integer> examIdResponseCounts = new HashMap<>();
+            while(rs.next()){
+                examIdResponseCounts.put(UUID.fromString(rs.getString("exam_id")), rs.getInt("response_count"));
+            }
+            return examIdResponseCounts;
+        });
     }
 }
--- a/service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
@@ -4,6 +4,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Stream;
@@ -102,5 +103,10 @@ public class ExamItemServiceImpl implements ExamItemService {
     @Override
     public int getExamPosition(final UUID examId) {
         return examItemQueryRepository.getCurrentExamItemPosition(examId);
+    }
+
+    @Override
+    public Map<UUID, Integer> getResponseCounts(UUID... examIds) {
+        return examItemQueryRepository.getResponseCounts(examIds);
     }
 }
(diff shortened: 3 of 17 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
```diff
@@ -79,17 +80,23 @@ public ExamAccommodationQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") fina
...
-    public List<ExamAccommodation> findApprovedAccommodations(final UUID examId) {
-        return getAccommodations(examId, true);
+    public List<ExamAccommodation> findApprovedAccommodations(final UUID... examIds) {
+        return getAccommodations(true, examIds);
     }
 
-    private List<ExamAccommodation> getAccommodations(final UUID examId, final boolean excludeDenied) {
-        final SqlParameterSource parameters = new MapSqlParameterSource("examId", examId.toString());
+    private List<ExamAccommodation> getAccommodations(final boolean excludeDenied, final UUID... examIds) {
+        final SqlParameterSource parameters = (examIds.length > 1)
```
burzum619: This is done for optimization purposes. If a single examId is passed, we will do an "=" check instead of a "where-in".

service/src/main/java/tds/exam/services/impl/ExamItemServiceImpl.java
```diff
@@ -103,4 +104,9 @@ public ExamItemServiceImpl(final ExamItemQueryRepository examItemQueryRepository
     public int getExamPosition(final UUID examId) {
         return examItemQueryRepository.getCurrentExamItemPosition(examId);
     }
+
+    @Override
+    public Map<UUID, Integer> getResponseCounts(UUID... examIds) {
```
millerts: final on the `examIds`

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -303,6 +305,42 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         examCommandRepository.update(pausedExams.toArray(new Exam[pausedExams.size()]));
     }
 
+    @Override
+    public List<ExpandableExam> findExamsBySessionId(final UUID sessionId, final String... expandableParams) {
+        final Set<String> params = Sets.newHashSet(expandableParams);
+        final Set<String> validExamStatuses = Sets.newHashSet(
```
millerts: What makes these the valid exam statuses?  And should they be passed by the caller via a query string rather than hard coded?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -769,4 +807,24 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
 
         return exam;
     }
+
```
millerts: extra line break

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
jtreuting: I believe that MySQL would optimize this and `col = 1` and `col IN (1)` would be handled the same.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -769,4 +807,24 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
...
+
+    private static void mapResponseCountsToExams(Map<UUID, ExpandableExam.Builder> examBuilders, Map<UUID, Integer> itemResponseCounts) {
```
millerts: Can we add the `final` keyword where appropriate?  There is a setting in code style that will make the arguments file when generating methods.

service/src/test/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryIntegrationTests.java
```diff
@@ -175,6 +176,21 @@ public void shouldGetTwoAccommodationsForExamAndSegmentAndIgnoreAccommodationTyp
...
+        final UUID examId1 = UUID.randomUUID();
+        final UUID examId2 = UUID.randomUUID();
+
+        ExamAccommodation accommodation1 = new ExamAccommodationBuilder().withExamId(examId1).build();
+        ExamAccommodation accommodation2 = new ExamAccommodationBuilder().withExamId(examId2).build();
+        ExamAccommodation accommodation3 = new ExamAccommodationBuilder().withExamId(examId2).build();
+
+        examAccommodationCommandRepository.insert(Arrays.asList(accommodation1, accommodation2, accommodation3));
+
+        assertThat(examAccommodationQueryRepository.findApprovedAccommodations(examId1)).hasSize(1);
```
millerts: Can we do a basic validation that the accommodations we do for each query are returning the correct accommodations?  

Maybe something like 

```
Set<UUID> exam1AccommodationIds = examAccommodationQueryRepository.findApprovedAccommodations(examId1).stream.map(accommodation::examId);
assertThat(exam1AccommodationsIds).containsExactly(examId1);
```

Something like that

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
burzum619: Ah, if thats so I can remove this piece. Great

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1333,6 +1337,90 @@ public void shouldNotCallUpdateWhenThereAreNoExamsToPauseInTheSession() {
...
+    @Test
+    public void shouldReturnExpandableExamsWithExamAccommodationsOnlyForSessionId() {
+        UUID sessionId = UUID.randomUUID();
+        Exam exam1 = new ExamBuilder().build();
+        Exam exam2 = new ExamBuilder().build();
+
+        ExamAccommodation exam1accommodation1 = new ExamAccommodationBuilder().withExamId(exam1.getId()).build();
+        ExamAccommodation exam1accommodation2 = new ExamAccommodationBuilder().withExamId(exam1.getId()).build();
+        ExamAccommodation exam2accommodation = new ExamAccommodationBuilder().withExamId(exam2.getId()).build();
+
```
millerts: extra line break

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1333,6 +1337,90 @@ public void shouldNotCallUpdateWhenThereAreNoExamsToPauseInTheSession() {
...
+
+        when(mockExamQueryRepository.findAllExamsInSessionWithStatus(eq(sessionId), any())).thenReturn(Arrays.asList(exam1, exam2));
+        when(mockExamAccommodationService.findApprovedAccommodations(any(), any()))
+            .thenReturn(Arrays.asList(exam1accommodation1, exam1accommodation2, exam2accommodation));
+        List<ExpandableExam> expandableExams = examService.findExamsBySessionId(sessionId, ExpandableExam.EXPANDABLE_PARAMS_EXAM_ACCOMMODATIONS);
+        verify(mockExamQueryRepository).findAllExamsInSessionWithStatus(eq(sessionId), any());
+        verify(mockExamAccommodationService).findApprovedAccommodations(any(), any());
+        // Should not call this
+        verify(mockExamItemService, never()).getResponseCounts(any());
```
millerts: I think you can do `verifyZeroInteractions(mockExamItemService);`.  It doesn't look like that mock should be called at all in this flow.  The never() is nice when the mock is called but you want to verify a particular method isn't called within the mock.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1333,6 +1337,90 @@ public void shouldNotCallUpdateWhenThereAreNoExamsToPauseInTheSession() {
...
+        when(mockExamItemService.getResponseCounts(any(), any())).thenReturn(ImmutableMap.of(
+            exam1.getId(), 4,
+            exam2.getId(), 1
+        ));
+        List<ExpandableExam> expandableExams = examService.findExamsBySessionId(sessionId,
+            ExpandableExam.EXPANDABLE_PARAMS_EXAM_ACCOMMODATIONS, ExpandableExam.EXPANDABLE_PARAMS_ITEM_RESPONSE_COUNT);
+        verify(mockExamQueryRepository).findAllExamsInSessionWithStatus(eq(sessionId), any());
+        verify(mockExamAccommodationService).findApprovedAccommodations(any(), any());
+        // Should not call this
+        verify(mockExamItemService).getResponseCounts(any(), any());
```
millerts: Same comment as above with the zero interactions.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -303,6 +305,42 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+            Map<UUID, Integer> itemResponseCounts = examItemService.getResponseCounts(examIds);
+            mapResponseCountsToExams(examBuilders, itemResponseCounts);
+        }
+
+        if (params.contains(ExpandableExam.EXPANDABLE_PARAMS_UNFULFILLED_REQUEST_COUNT)) {
+            //TODO: fetch count of unfulfilled print/emboss requests for each exam
+        }
+
+        // Build each exam
+        List<ExpandableExam> expandableExams = examBuilders.values().stream()
```
millerts: Can we just return here?  `return examBuilders.values().stream....`.  Pretty sure IntelliJ will warn you that there isn't a point in creating the variable since you immediately return it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -303,6 +305,42 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         examCommandRepository.update(pausedExams.toArray(new Exam[pausedExams.size()]));
     }
 
+    @Override
+    public Response<List<ExpandableExam>> findExamsBySessionId(final UUID sessionId, final String... expandableParams) {
+        final Set<String> params = Sets.newHashSet(expandableParams);
+        final Set<String> validExamStatuses = Sets.newHashSet(
```
burzum619: This is from the following portion of the query in the PR description:
`tOpp.status not in ('pending', 'suspended', 'denied');`

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
millerts: http://stackoverflow.com/questions/9926433/mysql-performance-in-clause-vs-equals-for-a-single-value

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Actually, just noticed a bug. These statuses come from the last part of the query in the PRs description:
`tOpp.status not in ('pending', 'suspended', 'denied');`

However, I'm doing the exact opposite below (only getting the exams in this status). I'll fix this.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
jtreuting: The last answer with 17 votes shows the `explain` command

service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
```diff
@@ -46,4 +55,47 @@ public int getCurrentExamItemPosition(final UUID examId) {
 
         return jdbcTemplate.queryForObject(SQL, params, Integer.class);
     }
+
+    @Override
+    public Map<UUID, Integer> getResponseCounts(final UUID... examIds) {
+        // Needs to toString() each UUID individually
+        final SqlParameterSource params = new MapSqlParameterSource("examIds",
+            Arrays.stream(examIds).map(UUID::toString).collect(Collectors.toSet()));
```
jtreuting: I would have to test locally, but passing in the list of UUIDs doesn't automatically get `.toString()`'d by the namemd jdbc template?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Yep, certainly could. Although after reviewing this myself, I realized I may want to return a Response<> wrapped List of ExpandableExams. So in this case, I think it makes it a bit cleaner to break it out into two lines. Let me know if you disagree (or if you think it shouldn't be wrapped in Response<>, since there aren't any validation/UI errors that occur here)

```
        // Build each exam
        List<ExpandableExam> expandableExams = examBuilders.values().stream()
            .map(builders -> builders.build())

        return new Response<>(expandableExams);
```

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1333,6 +1337,98 @@ public void shouldNotCallUpdateWhenThereAreNoExamsToPauseInTheSession() {
...
+        ));
+        Response<List<ExpandableExam>> response = examService.findExamsBySessionId(sessionId,
+            ExpandableExam.EXPANDABLE_PARAMS_EXAM_ACCOMMODATIONS, ExpandableExam.EXPANDABLE_PARAMS_ITEM_RESPONSE_COUNT);
+        assertThat(response.getData().isPresent()).isTrue();
+        assertThat(response.getError().isPresent()).isFalse();
+        List<ExpandableExam> expandableExams = response.getData().get();
+
+        verify(mockExamQueryRepository).findAllExamsInSessionWithStatus(eq(sessionId), any());
+        verify(mockExamAccommodationService).findApprovedAccommodations(any(), any());
+        // Should not call this
```
jtreuting: What does this comment mean?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: thx, preferences updated.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
millerts: Ya, not sure why that isn't the accepted answer.

service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
burzum619: Nope. I gave that a shot first, but unfortunately it didn't seem to work for me.

Which is strange considering the toStrings() evaluate to the same thing:

image

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
burzum619: Since that expandable param wasn't provided, that service method should not be called (hence the never()). I'll elaborate a bit more on the comment for clarity

PR #86, opened 2017-02-23
Feature/exams for session endpoint

burzum619:
This PR includes the endpoint that proctor will call to retrieve the list of "expandable exams" (exams with additional data) for a sessionId. 

Some notes:
- This, along with the code in ExamServiceImpl.findExamsBySessionId() covers the query found below. Please reference inline SQL comments below for explanations of where this data can/will be retrieved by the proctor application
- The service/endpoint will no longer return a Response-wrapped object. We do not need to wrap our data in a response since there is no possible validation errors or any errors that will be displayed in the UI.
- The query filters by exams with a status NOT IN 'pending', 'suspended', or 'denied'. The controller/service will not filter this implicitly, but these filter values will be be included as query parameters by the caller (the proctor app).
- The entrypoint in the legacy code that corresponds to this work is in TDS_Proctor - TestSessionDaoImpl.getCurrentSessionTestees()

```
select 
  tOpp._efk_AdminSubject,                         --exam.assessmentKey
  tOpp._fk_TestOpportunity as opportunityKey,     --exam.id
  tOpp._efk_Testee,                               --exam.studentId
  tOpp._efk_TestID,                               --exam.assessmentId
  tOpp.Opportunity,                               --exam.attempts
  tOpp.TesteeName,                                --exam.studentName
  TesteeID,                                       --exam.loginSsid  
  tOpp.Status,                                    --exam.status
  tOpp.DateCompleted,                             --exam.completedAt
  tOpp._fk_Session,                               --exam.sessionId    (uuid)
  tOpp.SessID as SessionID,                       --exam.sessionKey  (ADM-99)
  '' as sessionName,  
  maxitems as ItemCount,                          --exam.maxItems
  case when tOpp.status = 'paused' and datePaused is not null   -- Can be extracted from Exam if status == paused (diff of dateChanged and "now")
    then timestampdiff(MINUTE, datePaused, now(3)) 
  else 
    cast(null as CHAR) end as pauseMinutes,       -- null if not paused, otherwise the number of mins the exam has been paused for
  numResponses as ResponseCount,                   
  (
    select 
      count(*) 
    from 
      testopprequest REQ 
    where 
      REQ._fk_TestOpportunity = tOpp._fk_TestOpportunity 
      and REQ._fk_Session = ${sessionKey}
      and DateFulfilled is null 
      and DateSubmitted > ${midnightAM} 
      and DateSubmitted < ${midnightPM}
  ) as RequestCountN,                             -- the number of unfulfilled requests for this exam/session          
  (
    select 
      value as score 
    from 
      testopportunityscores S
    join configs.client_testscorefeatures F
      on S.MeasureOf = F.MeasureOf 
      and S.MeasureLabel = F.MeasureLabel 
    where 
      F.ClientName = ${clientname} 
      and ReportToProctor = 1 
      and S._fk_TestOpportunity = tOpp._fk_TestOpportunity 
      and S.IsOfficial = 1 
    limit 1
  ) as Score,                                 -- Not needed
  AccommodationString as Accommodations,      -- ExamService will return entire list of exam accommodations
  tOpp.customAccommodations,                  -- exam.customAccommodations
  tOpp.mode,                                  -- Not needed, always "online"
  ctp.msb                                     -- TODO: Need to fetch MSB flag from assessment
from testopportunity_readonly tOpp 
left outer join                       -- Join not necessary - only used to validate testid and clientname (part of exam) and to get msb flag. 
  configs.client_testproperties ctp 
  on tOpp._efk_testid = ctp.testid 
  and tOpp.clientname = ctp.clientname 
where 
  _fk_Session = ${sessionKey} 
  and tOpp.DateChanged > ${midnightAM} -- midnight date in UTC
  and tOpp.DateChanged < ${midnightPM} 
  and tOpp.status not in ('pending', 'suspended', 'denied');
  ```

### DIFF
--- a/service/src/main/java/tds/exam/web/endpoints/ExamController.java
+++ b/service/src/main/java/tds/exam/web/endpoints/ExamController.java
@@ -106,8 +106,8 @@ public class ExamController {
 
     @RequestMapping(value = "/session/{sessionId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
     ResponseEntity<List<ExpandableExam>> findExamsForSessionId(@PathVariable final UUID sessionId,
-                                                                         @RequestParam final Set<String> statusNot,
-                                                                         @RequestParam("expandable") final String... expandableParams) {
+                                                               @RequestParam final Set<String> statusNot,
+                                                               @RequestParam("expandable") final String... expandableParams) {
         final List<ExpandableExam> exams = examService.findExamsBySessionId(sessionId, statusNot,
             expandableParams);
 
--- a/service/src/main/java/tds/exam/web/endpoints/ExamController.java
+++ b/service/src/main/java/tds/exam/web/endpoints/ExamController.java
@@ -106,10 +106,9 @@ public class ExamController {
 
     @RequestMapping(value = "/session/{sessionId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
     ResponseEntity<List<ExpandableExam>> findExamsForSessionId(@PathVariable final UUID sessionId,
-                                                               @RequestParam final Set<String> statusNot,
-                                                               @RequestParam(value = "expandable", required = false) final String... expandableParams) {
-        final List<ExpandableExam> exams = examService.findExamsBySessionId(sessionId, statusNot,
-            expandableParams);
+                                                               @RequestParam(required = false) final Set<String> statusNot,
+                                                               @RequestParam(required = false) final String... embed) {
+        final List<ExpandableExam> exams = examService.findExamsBySessionId(sessionId, statusNot, embed);
 
         if (exams.isEmpty()) {
             return new ResponseEntity<>(exams, HttpStatus.NO_CONTENT);
--- a/service/src/main/java/tds/exam/web/endpoints/ExamController.java
+++ b/service/src/main/java/tds/exam/web/endpoints/ExamController.java
@@ -15,7 +15,9 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.util.List;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by burzum619** [comment]
General comment - the query filters by what appears to be a time-zone adjusted midnight value. The point of this seems to be to filter values by "active" exams in the session (exams that have registered a date-change or proctor print/emboss request in "this" day). I have ignored this for now as I don't see a clear reason for this (sessions do not span multiple days), but should we perhaps handle this case? If so, we can filter within the exam service, after retrieving all Exams in the session. Thoughts?

File: service/src/main/java/tds/exam/web/endpoints/ExamController.java [hunk 1]
```diff
@@ -79,28 +82,42 @@ public ExamController(ExamService examService) {
...
         final HttpHeaders headers = new HttpHeaders();
         headers.add("Location", link.getHref());
-    
+
         return new ResponseEntity<>(headers, HttpStatus.NO_CONTENT);
     }
 
+    @RequestMapping(value = "/session/{sessionId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<List<ExpandableExam>> findExamsForSessionId(@PathVariable final UUID sessionId,
+                                                               @RequestParam final Set<String> statusNot,
```
millerts: I think this and the expandable should be [`required=false`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html#required--).

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -79,28 +82,42 @@ public ExamController(ExamService examService) {
...
+                                                               @RequestParam("expandable") final String... expandableParams) {
```
millerts: I prefer this to follow the naming that @wtritch mentioned in our conversation which is to name this `embed` since your asking to embed more objects

PR #87, opened 2017-02-25
Feature/on completed paused events

jeffjohnson9046:
[TDS-604](https://jira.fairwaytech.com/browse/TDS-604):  Implementation for the `CommonDLL#_OnStatusCompleted_SP()` and `CommonDLL#_OnStatusPaused_SP()`

**NOTE:** This PR depends on this [TDS_Common pull request](https://github.com/SmarterApp/TDS_Common/pull/26).


service/src/main/java/tds/exam/repositories/ExamItemQueryRepository.java
```diff
@@ -14,7 +14,8 @@
      * @param examId the id of the {@link tds.exam.Exam} to find the position for
      * @return the item position of the last item responded to
      */
-    int getCurrentExamItemPosition(final UUID examId);
+
+    int getCurrentExamItemPosition(UUID examId);
```
jtreuting: Did you mean to remove the `final`?

service/src/main/java/tds/exam/repositories/FieldTestItemGroupQueryRepository.java
```diff
@@ -17,5 +17,13 @@
...
-    List<FieldTestItemGroup> find(final UUID examId, final String segmentKey);
+    List<FieldTestItemGroup> find(UUID examId, String segmentKey);
+
+    /**
+     * Find all {@link tds.exam.models.FieldTestItemGroup}s that were delivered/administered in an exam.
+     *
+     * @param examId The id of the {@link tds.exam.Exam} to fetch item groups by
+     * @return The list of all {@link tds.exam.models.FieldTestItemGroup}s in the specified exam
+     */
+    List<FieldTestItemGroup> findUsageInExam(UUID examId);
```
jtreuting: `final`

service/src/main/java/tds/exam/repositories/FieldTestItemGroupQueryRepository.java
```diff
@@ -17,5 +17,13 @@
      * @param segmentKey The id of the {@link tds.assessment.Segment} to fetch item groups by
      * @return The list of {@link tds.exam.models.FieldTestItemGroup}s fetched
      */
-    List<FieldTestItemGroup> find(final UUID examId, final String segmentKey);
+    List<FieldTestItemGroup> find(UUID examId, String segmentKey);
```
jtreuting: `final`

service/src/main/java/tds/exam/repositories/FieldTestItemGroupCommandRepository.java
```diff
@@ -14,5 +14,12 @@
...
      */
-    void insert(final List<FieldTestItemGroup> fieldTestItemGroups);
+    void insert(List<FieldTestItemGroup> fieldTestItemGroups);
+
+    /**
+     * Update a collection of {@link tds.exam.models.FieldTestItemGroup}s
+     *
+     * @param fieldTestItemGroups The {@link tds.exam.models.FieldTestItemGroup}s to update
+     */
+    void update(FieldTestItemGroup... fieldTestItemGroups);
```
jtreuting: `final`s

service/src/main/java/tds/exam/repositories/impl/FieldTestItemGroupQueryRepositoryImpl.java
```diff
@@ -73,6 +74,73 @@ public FieldTestItemGroupQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") fin
...
+                "   AND segment.segment_key = ftitem_group.segment_key \n" +
+                "   AND page.item_group_key = ftitem_group.group_key \n" +
+                "JOIN ( \n" +
+                "   SELECT \n" +
+                "   field_test_item_group_id, \n" +
+                "   MAX(id) AS id \n" +
+                "FROM \n" +
+                "   field_test_item_group_event \n" +
+                "GROUP BY \n" +
+                "   field_test_item_group_id) AS last_event \n" +
```
jtreuting: Lines 112 - 117 should be indented for readability I think

client/src/main/java/tds/exam/ExamSegment.java
```diff
@@ -71,6 +74,27 @@ public ExamSegment(Builder builder) {
         private int poolCount;
         private Instant createdAt;
 
+        public Builder fromSegment(ExamSegment segment) {
```
millerts: Can we make this static like [this](https://github.com/SmarterApp/TDS_ExamService/blob/develop/client/src/main/java/tds/exam/ExamAccommodation.java#L129) ?

pom.xml
```diff
@@ -26,7 +26,7 @@
         <unit-tests.skip>false</unit-tests.skip>
         <integration-tests.skip>true</integration-tests.skip>
 
-        <tds-common.version>0.0.1</tds-common.version>
+        <tds-common.version>0.0.2-SNAPSHOT</tds-common.version>
```
millerts: I don't think this should be committed.  The way we should do this is:

1. Merge common changes
2. Release common
3. Bump this released version.

I'll let @wtritch confirm, but our snapshots shouldn't be committed.  We can add a maven profile at some point to make it so this doesn't need to be changed to build using snapshots.

service/src/main/java/tds/exam/repositories/ExamItemQueryRepository.java
millerts: I think this file shouldn't have changed or be committed.

service/src/main/java/tds/exam/repositories/FieldTestItemGroupCommandRepository.java
```diff
@@ -14,5 +14,12 @@
      *
      * @param fieldTestItemGroups the {@link tds.exam.models.FieldTestItemGroup}s to insert
      */
-    void insert(final List<FieldTestItemGroup> fieldTestItemGroups);
+    void insert(List<FieldTestItemGroup> fieldTestItemGroups);
```
millerts: looks like a bad merge.  Everything (as @jtreuting) should have final keyword where appropriate.

service/src/main/java/tds/exam/repositories/impl/FieldTestItemGroupQueryRepositoryImpl.java
```diff
@@ -73,6 +74,73 @@ public FieldTestItemGroupQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") fin
...
+                "   ftitem_event.deleted_at";
+
+        return jdbcTemplate.query(SQL, parameters, (rs, r) -> new FieldTestItemGroup.Builder()
+            .withId(rs.getLong("id"))
+            .withExamId(UUID.fromString(rs.getString("exam_id")))
+            .withLanguageCode(rs.getString("language_code"))
+            .withGroupKey(rs.getString("item_group_key"))
+            .withSegmentKey(rs.getString("segment_key"))
+            .withPositionAdministered(rs.getInt("position_administered"))
+            .withDeletedAt(ResultSetMapperUtility.mapTimestampToInstant(rs, "deleted_at"))
```
millerts: I'll have to dig into the logic but it seems odd to me that we're purposefully fetching deleted field test items.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -281,6 +287,8 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
 
         examCommandRepository.update(updatedExam);
 
+        changeListeners.forEach(listener -> listener.accept(exam, updatedExam));
```
millerts: I believe these listeners should be executed asynchornously otherwise this will block the response until the listener finishes.  Just looking at the logic for onComplete that could take a bit of time.

millerts: Also, are you expecting this whole flow to fail if the listener throws which will roll back all the changes done prior to this call?

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -0,0 +1,90 @@
...
+        // If the status has not changed between exam instances or the status has not already been set to "completed" on
+        // the new version of the exam, exit
+        if (oldExam.getStatus().equals(newExam.getStatus())
+            || !newExam.getStatus().getCode().equals(ExamStatusCode.STATUS_COMPLETED)) {
+            return;
+        }
+
+        // CommonDLL#_OnStatus_Completed_SP, line 1425: Update the exam to indicate this segment is not permeable.
+        // Legacy code sets isPermeable to -1
+        ExamSegment segment = examSegmentQueryRepository.findByExamIdAndSegmentPosition(newExam.getId(),
```
millerts: I think line 62 to 69 should be in the ExamSegmentService not here. l I think services should be the only things that use repositories.  That allows us to keep our layered architecture clear.  Repositories leaking to other areas makes that architecture blurry for future coders.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -0,0 +1,90 @@
...
+        // relationships
+        examineeService.insertAttributesAndRelationships(newExam, ExamineeContext.FINAL);
+
+        // CommonDLL#_OnStatus_Completed_SP, line 1431: not importing CommonDLL#_RecordBPSatisfaction_SP.  The
+        // tables/queries that are referenced in that method are only used by the loader stored procedures, schema
+        // creation/modification scripts and/or SimDLL.java (which is related to the simulator).
+
+        // CommonDLL#_OnStatus_Completed_SP, lines 1445 - 1453: Find all the field test items that were administered
+        // during an exam and record their usage.
+        List<FieldTestItemGroup> fieldTestItemGroupsToUpdate =
```
millerts: Same comment about using a service for lines 81 - 86.  If we don't have a FieldTestService then let's add one.

service/src/main/java/tds/exam/utils/listeners/OnPausedStatusExamChangeListener.java
```diff
@@ -0,0 +1,67 @@
...
+            || !newExam.getStatus().getCode().equals(ExamStatusCode.STATUS_PAUSED)) {
+            return;
+        }
+
+        // CommonDLL#_OnStatus_Paused_SP, line 1493: Update the exam to indicate this segment is not permeable if the
+        // segment was previously permeable and the restore permeable condition is either "segment" or "paused".  Legacy
+        // code sets isPermeable to -1, which means false.
+        // Omit CommonDLL#_OnStatus_Paused_SP, line 1489 - 1493: If the segment's getRestorePermeableCondition is set to
+        // "segment" or "paused", then it is not equal to "completed" thus the update to the segment should happen.
+        ExamSegment segment = examSegmentQueryRepository.findByExamIdAndSegmentPosition(newExam.getId(),
```
millerts: Please move lines 49 - 62 to the ExamSegmentService.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -142,6 +143,12 @@
     @Mock
     private ExamineeService mockExamineeService;
 
+    @Mock
+    private ChangeListener<Exam> mockOnCompletedExamChangeListener;
+
+    @Mock
+    private ChangeListener<Exam> mockOnPausedExamChangeListener;
```
millerts: Let's remove one of these.  If you verify a single one is called I think that will sufficiently prove that the service calls the listener.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -0,0 +1,90 @@
...
+            fieldTestItemGroupQueryRepository.findUsageInExam(newExam.getId());
+
+        if (fieldTestItemGroupsToUpdate.size() > 0) {
```
millerts: I prefer us to use the `if(!fieldTestItemgroupsToUpdate.isEmpty())` style rather than checking size.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -96,6 +98,8 @@
 
     private final Set<String> statusesThatCanTransitionToPaused;
 
+    private final Collection<ChangeListener<Exam>> changeListeners;
```
millerts: Can we rename this variable to examStatusChangeListeners?  The reason being is that the only time this change listener is called is when the exam status changes.  As named right now I would expect it to be called whenever an exam changes.

millerts: I notice these are called when updating the status through the update status service method?  Should it also be called when reopening or starting an exam?  The status changes in those flows as well.  

When someone implements one of these how do they know exactly when the listener will be called?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -111,7 +115,8 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
                            ExamStatusQueryRepository examStatusQueryRepository,
                            ExamAccommodationService examAccommodationService,
                            ExamApprovalService examApprovalService,
-                           ExamineeService examineeService) {
+                           ExamineeService examineeService,
+                           Collection<ChangeListener<Exam>> changeListeners) {
```
millerts: Do you think we should have a `@Configuration` class that defines this collection?  What happens when this is reused in another class for some other Exam change that doesn't care about status changes?  Those would be included in this list, right?  Is that your intention?

service/src/main/java/tds/exam/repositories/ExamItemQueryRepository.java
jeffjohnson9046: I did mean to remove `final`, but this might not be the appropriate PR to do it.  Declaring `final` on an interface doesn't have any practical effect.  But I can back it out.

service/src/main/java/tds/exam/repositories/FieldTestItemGroupQueryRepository.java
jeffjohnson9046: Same comment as above; it's my understanding that adding `final` to a parameter on an interface doesn't have any practical effect.

service/src/main/java/tds/exam/repositories/ExamItemQueryRepository.java
millerts: I agree. However, JT mentioned that it signals intent. I see no issue with including it.

pom.xml
wtritch: yessir, we shouldn't ever be committing with an "x.x.x-SNAPSHOT" dependency.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
wtritch: Ideally, we wouldn't have separate groups of ChangeListeners, they shouldn't care how or why an Exam has changed, just if they should do something about it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
wtritch: That'd be my expectation, otherwise we could end up in an undefined state where a change listener may or may not have executed on a change.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
wtritch: +1 I picture a change listener as a light-weight trigger, (closer to a Controller.)

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: I guess my async comment is premature optimization.  If it takes a long time we can come back to threading this stuff.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: I see your point.  My suggestion is made because right now is there is a very narrow focus on when these change listeners are called.

wtritch: I see the ChangeListeners as a way of externalizing a stupid-long if-chain that would otherwise be embedded in the update method, rather than asynchronous, separate processes.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Ignore this flow.  @wtritch and I had an offline discussion and I've replied below

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: I agree. I wasn't being clear in my earlier comment.  This is what I think (and I believe Will agrees) should happen.

We should add a new method to ExamService

```
private Exam save(Exam exam, Exam updatedExam) {
  examCommandRepository.update(exam);
  changeListeners.forEach(listener -> listener.accept(exam, updatedExam));
}
```

Then everywhere we're calling `examCommandRepository.update(exam);` should instead call that method.  This is assuming listeners are only called when exams are updated.

service/src/main/java/tds/exam/repositories/impl/FieldTestItemGroupQueryRepositoryImpl.java
jeffjohnson9046: In legacy, the query to get the field test items for an opportunity looks like this (`CommonDLL#_OnStatus_Completed_SP`, line 1445):

```
"insert into ${groupsTableName} (gid, bid, seg, pos) "
+ " select R.groupID, I.blockID, R.segment, min(R.position) from testeeresponse R, ft_opportunityitem I "
+ " where R._fk_TestOpportunity = ${oppkey} and I._fk_TestOpportunity = ${oppkey} and R.segment = I.segment and R.groupID = I.groupID and R.IsFieldTest = 1 "
+ " group by R.segment, R.groupID, I.blockID";
```

On `CommonDLL#_OnStatus_Completed_SP` line 1455, the values inserted into the temporary table cited above are used to update `session.ft_opportunityitem`.

There is a `deleted` column on the `session.ft_opportunityitem` table, but it doesn't seem to factor in here (in either the `SELECT` or `UPDATE`).  So the intent with line 140 of this PR is to preserve the `deletedAt` value; if the `FieldTestItemGroup` was deleted before the update, it will remain deleted after the updated.

millerts: Agreed.  We have some questions around some of the FT fields, but this isn't critical right now.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -793,10 +798,23 @@ private Exam updateExamWithCustomAccommodations(final Exam exam,
...
     }
+
+    /**
+     * Perform the update to the new {@link tds.exam.Exam} then execute the
+     * {@link tds.common.entity.utils.ChangeListener}s to apply any rules/business logic as a result of the update.
+     *
+     * @param exam        The {@link tds.exam.Exam} in its original state
+     * @param updatedExam The {@link tds.exam.Exam} with new values to persist
+     */
+    private void updateExam(final Exam exam, final Exam updatedExam) {
```
wtritch: I like having a central update method.  I have to admit, I was a little surprised by this, however.  I was expecting more of a:
```
public Exam save(final Exam exam);
```
method.  But I can see how we wouldn't want to re-fetch the exam if we already have the original that we've processed.

millerts: Which right now is always the case when we are updating an exam.

service/src/main/java/tds/exam/services/impl/FieldTestServiceImpl.java
```diff
@@ -196,9 +197,19 @@ public int selectItemGroups(final Exam exam, final Assessment assessment, final
...
+    @Override
+    public void update(final FieldTestItemGroup... fieldTestItemGroups) {
+        fieldTestItemGroupCommandRepository.update(fieldTestItemGroups);
+    }
+
     /*
-        This helper method is a null-tolerant Instant/date comparison for the test window
-     */
+            This helper method is a null-tolerant Instant/date comparison for the test window
+         */
```
millerts: This formatting doesn't look right.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -0,0 +1,83 @@
...
+        // CommonDLL#_OnStatus_Completed_SP, line 1430: insert the final version of the student's attributes and
+        // relationships
+        examineeService.insertAttributesAndRelationships(newExam, ExamineeContext.FINAL);
+
+        // CommonDLL#_OnStatus_Completed_SP, line 1431: not importing CommonDLL#_RecordBPSatisfaction_SP.  The
+        // tables/queries that are referenced in that method are only used by the loader stored procedures, schema
+        // creation/modification scripts and/or SimDLL.java (which is related to the simulator).
+
+        // CommonDLL#_OnStatus_Completed_SP, lines 1445 - 1453: Find all the field test items that were administered
+        // during an exam and record their usage.
```
wtritch: As a total noob to this codebase, I find the comments a little confusing.

service/src/test/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListenerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Test
+    public void shouldUpdateFieldTestItemGroupsWhenAnExamIsSetToCompleted() {
+        Exam oldExam = new ExamBuilder().build();
+        Exam newExam = new ExamBuilder()
+            .withStatus(new ExamStatusCode(ExamStatusCode.STATUS_COMPLETED, ExamStatusStage.IN_PROGRESS), Instant.now())
+            .build();
+        ExamSegment mockSegment = new ExamSegmentBuilder()
+            .withExamId(newExam.getId())
+            .build();
+        FieldTestItemGroup mockFirstFtItemGroup = new FieldTestItemGroup.Builder()
```
millerts: Not a huge deal but you could've leveraged that random value creator that Ernie introduced into exam and then only set the values you have to have to make the test flow.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
jtreuting: the goal is to provide an easy way to get back to the legacy code that we ported from so that we can compare them.  does that address the comment?  or was it something else?

millerts: And my hope would be that we can remove the comments in a future release once we're confident the system continues to work as previously implemented.

PR #88, opened 2017-02-27
Check for code changes for accommodations of the same type and replac…

jtreuting:
…e changed accommodations

Addresses TDS-682: https://jira.fairwaytech.com/browse/TDS-682

Basically it was using contains to check if accommodations had changed, but that included the code in the check, so when an accommodation code changed, it was just being inserted again and the old one was not removed.

Changed the logic to see if one exists for the same type and segment, and replace it (delete then insert).  I did not think that adding code to the _event table made sense since it is only because of the flow of starting an exam that makes this "update" necessary.  Ideally (in the future), we will alter the flow to only insert accommodations once instead of adding default then updating, once we are confident that it won't have an unintended effect somewhere else.


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -302,37 +302,54 @@ private static String getOtherAccommodationValue(String formattedValue) {
...
                 }
             } else {
                 examAccommodationsToInsert.add(examAccommodation);
             }
         }
 
-        if (!examAccommodationsToInsert.isEmpty()) {
-            examAccommodationCommandRepository.insert(examAccommodationsToInsert);
+        if (!examAccommodationsToDelete.isEmpty()) {
+            examAccommodationCommandRepository.delete(examAccommodationsToDelete);
```
burzum619: Do we have test coverage that hits this conditional (and verifies the repository.delete() call? Seems like we might have missed it

jtreuting: i'm guessing not... good point.

i made sure that all existing tests passed but i'll look into that

PR #89, opened 2017-02-27
Exam Print Request endpoint

burzum619:
[TDS-711] This PR is for creating an endpoint for creating exam print and emboss requests. The student application will call this endpoint whenever a student select to print or emboss a passage or item, or if the auto-embossing accommodation is enabled. The proctor application will consume these requests and decide whether to allow or deny a print request for a student.  

- For testing purposes, and because it will be required later, I have added the ExamPrintRequestQueryRepository/Impl classes and a method to read the counts of unfulfilled requests for a session and list of examIds.

### DIFF
--- a/client/src/main/java/tds/exam/ExamPrintRequest.java
+++ b/client/src/main/java/tds/exam/ExamPrintRequest.java
@@ -1,0 +1,241 @@
+package tds.exam;
+
+import org.joda.time.Instant;
+
+import java.util.UUID;
+
+/**
+ * Represents a request to print or emboss and exam item, passage, or page.
+ */
+public class ExamPrintRequest {
+    public static final String REQUEST_TYPE_EMBOSS_ITEM = "EMBOSSITEM";
+    public static final String REQUEST_TYPE_EMBOSS_PASSAGE = "EMBOSSPASSAGE";
+    public static final String REQUEST_TYPE_PRINT_ITEM = "PRINTITEM";
+    public static final String REQUEST_TYPE_PRINT_PASSAGE = "PRINTPASSAGE";
+
+    private UUID id;
+    private UUID examId;
+    private UUID sessionId;
+    private String type;
+    private String value;
+    private int itemPosition;
+    private int pagePosition;
+    private String parameters;
+    private String description;
+    private Instant createdAt;
+    private Instant approvedAt;
+    private Instant deniedAt;
+    private String reasonDenied;
+
+    private ExamPrintRequest() {
+    }
+
+    public ExamPrintRequest(Builder builder) {
+        this.createdAt = builder.createdAt;
+        this.pagePosition = builder.pagePosition;
+        this.itemPosition = builder.itemPosition;
+        this.approvedAt = builder.approvedAt;
+        this.examId = builder.examId;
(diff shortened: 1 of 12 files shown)


client/src/main/java/tds/exam/ExamPrintRequest.java
```diff
@@ -0,0 +1,241 @@
+package tds.exam;
+
+import org.joda.time.Instant;
+
+import java.util.UUID;
+
+/**
+ * Represents a request to print or emboss and exam item, passage, or page.
+ */
+public class ExamPrintRequest {
```
millerts: Not critical but this would've been a good pojo to use the @AutoValue stuff

service/src/main/java/tds/exam/services/ExamPrintRequestService.java
```diff
@@ -0,0 +1,28 @@
...
+    void insert(final ExamPrintRequest examPrintRequest);
+
+    /**
+     * Retrieves a map of exam ids to their respective count of unfulfilled {@link tds.exam.ExamPrintRequest}
+     *
+     * @param sessionId The session of the exams
+     * @param examIds   The list of exam ids for each {@link tds.exam.ExamPrintRequest}
+     * @return A map of exam ids to their request count
+     */
+    Map<UUID, Integer> findRequestCountsForExamIds(final UUID sessionId, final UUID... examIds);
```
millerts: This seems to only be used in a test.  Is it going to be used at some point by exam or a client?   

If the intent is only to verify the data in integration test then I suggest this method be in the test src tree.

burzum619: That method will be called by ExamServiceImpl. See the TODO: in ExamServiceImpl.findExamsBySessionId(). I didn't want to include that work in this PR, but I did want to be able to test that my inserts were working.

client/src/main/java/tds/exam/ExamPrintRequest.java
burzum619: Good point, I definitely should have.

service/src/main/java/tds/exam/services/impl/ExamPrintRequestServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    @Autowired
+    public ExamPrintRequestServiceImpl(final ExamPrintRequestCommandRepository examPrintRequestCommandRepository,
+                                       final ExamPrintRequestQueryRepository examPrintRequestQueryRepository) {
+        this.examPrintRequestCommandRepository = examPrintRequestCommandRepository;
+        this.examPrintRequestQueryRepository = examPrintRequestQueryRepository;
+    }
+
+    @Override
+    public void insert(final ExamPrintRequest examPrintRequest) {
+        examPrintRequestCommandRepository.insert(examPrintRequest);
```
jtreuting: Since the `insert` method on the repository does an insert and then an update (for the event), we need this to be Transactional

burzum619: Nice catch, forgot about the update being a separate transaction.

PR #90, opened 2017-02-28
Integration exam print service into getExamsBySessionId flow (for pro…

burzum619:
…ctor)

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -12,6 +12,7 @@ import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -54,6 +55,7 @@ import tds.exam.services.ExamPrintRequestService;
 import tds.exam.services.ExamSegmentService;
 import tds.exam.services.ExamService;
 import tds.exam.services.ExamineeService;
+import tds.exam.services.ExpandableExamMapper;
 import tds.exam.services.SessionService;
 import tds.exam.services.StudentService;
 import tds.exam.services.TimeLimitConfigurationService;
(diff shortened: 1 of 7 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -819,4 +824,12 @@ private static void mapExamAccommodationsToExams(final Map<UUID, ExpandableExam.
             builder.withExamAccommodations(sortedExamAccommodations);
         });
     }
+
+    private void mapPrintRequestCountsToExams(final Map<UUID, ExpandableExam.Builder> examBuilders, final Map<UUID, Integer> examRequestCounts) {
```
millerts: Now that I see this in practice I feel like we should have a common interface that takes in this data.  We can then create objects that can process the `params` embedded object.

For example:
```
public interface ExpandableExamMapper {
  void updateExpandableMapper(params, session, examBuilders);
}

public class PrintRequestsExpandableMapper implements ExpandableExamMapper {
  //load whatever autowired dependencies are needed

  @Override
  public final void updateExpandableMapper(params, session, examBuilders) {
    Does this contain the param?  If not return
    Fetch data and update examBuilders
  }
}
```

I understand this may be the last for this particular use case but every time we want to add a new property to expandable exam we're going to add yet another if check in this flow.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -325,7 +329,8 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         }
 
         if (params.contains(ExpandableExam.EXPANDABLE_PARAMS_UNFULFILLED_REQUEST_COUNT)) {
-            //TODO: fetch count of unfulfilled print/emboss requests for each exam
+            Map<UUID, Integer> examRequestCounts = examPrintRequestService.findRequestCountsForExamIds(sessionId, examIds);
```
millerts: Can we rename `params` to something that has more meaning?  I believe the Set represents the `expandableExamAttribute`.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: The interface and implementations should live outside of ExamService but within exam project.  ExamService can take them in via constructor autowiring.  @jeffjohnson9046 did a similar thing with the listeners in his recent PR.

PR #93, opened 2017-03-03
Bugfix/student default accommodations

jtreuting:
Fixed bug TDS-740 where the student accommodations from ART are not being respected.

Student codes are now passed into initializeExamAccommodations so that the student accommodations are always used then the default assessment accommodations from the database are included for other accommodation types. 

Also, renamed `guestAccommodations` to `studentAccommodation` since it can be either guest or student accommodations from ART.

Since I'll be gone tomorrow, I'm going to merge this later tonight in a couple hours.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
@@ -76,12 +76,6 @@ class ExamAccommodationServiceImpl implements ExamAccommodationService {
 
     @Transactional
     @Override
-    public List<ExamAccommodation> initializeExamAccommodations(final Exam exam) {
-        return initializeExamAccommodations(exam, "");
-    }
-
-    @Transactional
-    @Override
     public List<ExamAccommodation> initializeExamAccommodations(final Exam exam, final String studentAccommodationCodes) {
         Instant now = Instant.now();
         // This method replaces StudentDLL._InitOpportunityAccommodations_SP.  One note is that the calls to testopportunity_readonly were not implemented because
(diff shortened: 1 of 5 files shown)


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -74,19 +76,37 @@ public ExamAccommodationServiceImpl(final ExamAccommodationQueryRepository examA
...
         // these tables are only used for proctor and that is handled via the proctor related endpoints.
 
         // StudentDLL fetches the key accommodations via CommonDLL.TestKeyAccommodations_FN which this call replicates.  The legacy application leverages
         // temporary tables for most of its data structures which is unnecessary in this case so a collection is returned.
         List<Accommodation> assessmentAccommodations = assessmentService.findAssessmentAccommodationsByAssessmentKey(exam.getClientName(), exam.getAssessmentKey());
 
+        // Get the list of accommodations from the student package, for the particular Exam subject
+        //  Take the accommodation code, lookup the accommodation and put into a Map by type for easy lookup
+        Map<String, Accommodation> studentAccommodations = new HashMap<>();
+        splitAccommodationCodes(exam.getSubject(), studentAccommodationCodes).forEach(code -> {
```
millerts: premature optimization warning; Do you think it would be better to convert the assessmentAccommodations List above to `Map<code, accommodation>`?  Then this wouldn't be iterating over the accommodations for every code and instead would be doing an O(1) lookup in the map?

Then below you would simply change it to `assessmentAccommodationsMap.values().stream...`

jtreuting: I like it.  No reason I can see not to.

PR #95, opened 2017-03-04
Wait for segment approval endpoint/service

burzum619:
This endpoint will be called by legacy student whenever approval is needed for segment entry or exit. A separate PR in TDS_Student will cover the changes to call this endpoint.

The entry point in the legacy for this port is in StudentDLL.T_WaitForSegment_SP, line 1840. This is called on the student side from OpportunityRepository.waitForSegment().

### DIFF
--- a/client/src/main/java/tds/exam/SegmentApprovalRequest.java
+++ b/client/src/main/java/tds/exam/SegmentApprovalRequest.java
@@ -1,0 +1,51 @@
+package tds.exam;
+
+import java.util.UUID;
+
+/**
+ * A request to wait for a segment approval
+ */
+public class SegmentApprovalRequest {
+    private UUID sessionId;
+    private UUID browserId;
+    private int segmentPosition;
+    private boolean entryApproval;
+
+    private SegmentApprovalRequest() {
+    }
+
+    public SegmentApprovalRequest(final UUID sessionId, final UUID browserId, final int segmentPosition, final boolean entryApproval) {
+      this.sessionId = sessionId;
+      this.browserId = browserId;
+      this.segmentPosition = segmentPosition;
+      this.entryApproval = entryApproval;
+    }
+
+    /**
+     * @return The session id of the exam requesting segment approval
+     */
+    public UUID getSessionId() {
+        return sessionId;
+    }
+
+    /**
+     * @return The browser id of the exam requesting segment approval
+     */
+    public UUID getBrowserId() {
+        return browserId;
+    }
+
+    /**
+     * @return The position of the exam segment to seek entry/exit approval for
+     */
+    public int getSegmentPosition() {
+        return segmentPosition;
+    }
+
+    /**
+     * @return If {@code true}, this is an entry approval request. If {@code false} this is an exit segment approval request
+     */
+    public boolean isEntryApproval() {
+        return entryApproval;
+    }
+}
--- a/service/src/main/java/tds/exam/services/ExamService.java
+++ b/service/src/main/java/tds/exam/services/ExamService.java
@@ -13,6 +13,7 @@ import tds.exam.ExamConfiguration;
 import tds.exam.ExamStatusCode;
 import tds.exam.ExpandableExam;
 import tds.exam.OpenExamRequest;
+import tds.exam.SegmentApprovalRequest;
 
 /**
  * Main entry point for interacting with {@link Exam}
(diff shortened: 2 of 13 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -277,6 +278,12 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
     @Transactional
     @Override
     public Optional<ValidationError> updateExamStatus(final UUID examId, final ExamStatusCode newStatus, final String statusChangeReason) {
+        return updateExamStatus(examId, newStatus, statusChangeReason, -1);
+    }
+
+    @Override
+    public Optional<ValidationError> updateExamStatus(final UUID examId, final ExamStatusCode newStatus, final String statusChangeReason,
```
millerts: Per my comment above let's make this private.

client/src/main/java/tds/exam/SegmentApprovalRequest.java
```diff
@@ -0,0 +1,51 @@
...
+    private UUID sessionId;
+    private UUID browserId;
+    private int segmentPosition;
+    private boolean entryApproval;
+
+    private SegmentApprovalRequest() {
+    }
+
+    public SegmentApprovalRequest(final UUID sessionId, final UUID browserId, final int segmentPosition, final boolean entryApproval) {
+      this.sessionId = sessionId;
```
millerts: I see how this is used.  The approval service is always checking for session id.  I guess to keep it consistent we can continue to require it.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -338,6 +346,28 @@ public void pauseAllExamsInSession(final UUID sessionId) {
             .collect(Collectors.toList());
     }
 
+    @Override
+    public Optional<ValidationError> waitForSegmentApproval(final UUID examId, final SegmentApprovalRequest request) {
```
millerts: Transactional?

service/src/test/java/tds/exam/web/endpoints/ExamControllerIntegrationTests.java
```diff
@@ -220,4 +226,38 @@ public void shouldThrowWithNoStatusProvided() throws Exception {
...
+        SegmentApprovalRequest request = random(SegmentApprovalRequest.class);
+        when(mockExamService.waitForSegmentApproval(eq(examId), any())).thenReturn(Optional.of(new ValidationError("some", "error")));
+
+        ObjectWriter ow = objectMapper
+            .writer().withDefaultPrettyPrinter();
+
+        http.perform(put(new URI(String.format("/exam/%s/segmentApproval/", examId)))
+            .contentType(MediaType.APPLICATION_JSON)
+            .content(ow.writeValueAsString(request)))
+            .andExpect(status().isUnprocessableEntity());
```
millerts: We should probably check the location if we're going to respond with it.

service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -55,11 +56,24 @@
...
      * Change the {@link tds.exam.Exam}'s status to a new status.
      *
+     * @param examId                    The id of the exam whose status is being changed
+     * @param newStatus                 The {@link tds.exam.ExamStatusCode} to transition to
+     * @param statusChangeReason        The reason why the {@link tds.exam.Exam} status is being updated
+     * @param waitingForSegmentPosition The position of the exam segment to pause for approval
+     * @return {@code Optional<ValidationError>} if the {@link tds.exam.Exam} cannot be updated from its current status
+     * to the new status; otherwise {@code Optional.empty()}.
+     */
+    Optional<ValidationError> updateExamStatus(final UUID examId, final ExamStatusCode newStatus, final String statusChangeReason,
```
burzum619: Yep, you're right. Nice catch

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Is it necessary? updateExamStatus() is the only place where we have (multiple) writes, and that is already tagged @Transactional

PR #96, opened 2017-03-06
Exit Segment endpoint

burzum619:
This PR is for the Exam Service endpoint for exiting an exam segment. This endpoint, along with "waitForSegmentApproval" will be called by the student legacy application. The code for calling these two endpoints will be included in a separate PR on TDS_Student

For reference, the legacy code that this endpoint replace is found in `OpportunityService.exitSegment()` , `OpportunityRepository.exitSegment()`, and `StudentDLL.T_ExitSegment_SP`

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
@@ -1,5 +1,6 @@
 package tds.exam.services.impl;
 
+import org.joda.time.Instant;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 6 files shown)


service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
```diff
@@ -41,4 +49,21 @@ public ExamSegmentController(final ExamSegmentService examSegmentService) {
...
+    ResponseEntity<NoContentResponseResource> exitSegment(@PathVariable final UUID examId,
+                                                          @PathVariable final int segmentPosition) {
+        Optional<ValidationError> maybeError = examSegmentService.exitSegment(examId, segmentPosition);
+
+        if (maybeError.isPresent()) {
+            NoContentResponseResource response = new NoContentResponseResource(maybeError.get());
+            return new ResponseEntity<>(response, HttpStatus.UNPROCESSABLE_ENTITY);
+        }
+
+        Link link = linkTo(methodOn(ExamController.class).getExamById(examId)).withSelfRel();
```
millerts: Why are we including the location to the exam?  I would expect the exam segment.

millerts: https://en.wikipedia.org/wiki/HTTP_location makes me think that this shouldn't be included at all

service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -192,4 +195,22 @@ public int initializeExamSegments(final Exam exam, final Assessment assessment)
     public void update(final ExamSegment... examSegments) {
         commandRepository.update(Arrays.asList(examSegments));
     }
+
+    @Override
+    public Optional<ValidationError> exitSegment(final UUID examId, final int segmentPosition) {
+        Optional<ExamSegment> maybeExamSegment = queryRepository.findByExamIdAndSegmentPosition(examId, segmentPosition);
```
millerts: I know this isn't your change but can we name queryRepository something like examSegmentQueryReposiotry.

PR #104, opened 2017-03-10
Proctor Print request endpoints - Deny, approve, fetch approved

burzum619:
Implementation of multiple endpoints used by TDS_Proctor print request interaction.

The following APIs have been added:
- find and approve print request
- find all approved print requests for session
- deny print request
- find unfulfilled requests

These endpoints will be called by TDS_Proctor. Those changes will be made in a separate PR.

These are ports of data-access code found in StudentDLL and TesteeRequestRepository. For reference, please see the following methods in TesteeRequestRepository.java (TDS_Proctor)

- getTesteeRequestValues() (find and approve)
- getApprovedTesteeRequests() (find approved requests for session)
- getCurrentTesteeRequests() (find unfulfilled requests for session and exam)
- denyTesteeRequest()

### DIFF
--- a/client/src/main/java/tds/exam/ExamPrintRequest.java
+++ b/client/src/main/java/tds/exam/ExamPrintRequest.java
@@ -25,8 +25,8 @@ public class ExamPrintRequest {
     private String parameters;
     private String description;
     private Instant createdAt;
-    private Instant approvedAt;
-    private Instant deniedAt;
+    private Instant changedAt;
+    private ExamPrintRequestStatus status;
     private String reasonDenied;
 
     private ExamPrintRequest() {
@@ -36,10 +36,10 @@ public class ExamPrintRequest {
         this.createdAt = builder.createdAt;
         this.pagePosition = builder.pagePosition;
         this.itemPosition = builder.itemPosition;
-        this.approvedAt = builder.approvedAt;
+        this.changedAt = builder.changedAt;
         this.examId = builder.examId;
         this.parameters = builder.parameters;
-        this.deniedAt = builder.deniedAt;
+        this.status = builder.status;
         this.value = builder.value;
         this.id = builder.id;
         this.sessionId = builder.sessionId;
@@ -59,8 +59,8 @@ public class ExamPrintRequest {
         private String parameters;
         private String description;
         private Instant createdAt;
-        private Instant approvedAt;
-        private Instant deniedAt;
+        private Instant changedAt;
+        private ExamPrintRequestStatus status;
         private String reasonDenied;
 
         public Builder(UUID id) {
@@ -112,13 +112,13 @@ public class ExamPrintRequest {
             return this;
         }
 
-        public Builder withApprovedAt(Instant approvedAt) {
-            this.approvedAt = approvedAt;
+        public Builder withChangedAt(Instant changedAt) {
+            this.changedAt = changedAt;
             return this;
         }
 
-        public Builder withDeniedAt(Instant deniedAt) {
-            this.deniedAt = deniedAt;
+        public Builder withStatus(ExamPrintRequestStatus status) {
+            this.status = status;
             return this;
         }
 
@@ -131,10 +131,10 @@ public class ExamPrintRequest {
             this.createdAt = request.createdAt;
             this.pagePosition = request.pagePosition;
             this.itemPosition = request.itemPosition;
-            this.approvedAt = request.approvedAt;
+            this.changedAt = request.changedAt;
             this.examId = request.examId;
             this.parameters = request.parameters;
-            this.deniedAt = request.deniedAt;
+            this.status = request.status;
             this.value = request.value;
             this.id = request.id;
             this.sessionId = request.sessionId;
@@ -220,17 +220,17 @@ public class ExamPrintRequest {
     }
 
     /**
-     * @return If approved, the {@link org.joda.time.Instant} the request was approved by a proctor
+     * @return The {@link org.joda.time.Instant} of when the {@link tds.exam.ExamPrintRequestStatus} last changed
      */
-    public Instant getApprovedAt() {
-        return approvedAt;
+    public Instant getChangedAt() {
+        return changedAt;
     }
 
     /**
-     * @return If denied, the {@link org.joda.time.Instant} the request was denied by a proctor
+     * @return The {@link tds.exam.ExamPrintRequestStatus} of the {@link tds.exam.ExamPrintRequest}
      */
-    public Instant getDeniedAt() {
-        return deniedAt;
+    public ExamPrintRequestStatus getStatus() {
+        return status;
     }
 
     /**
@@ -255,8 +255,7 @@ public class ExamPrintRequest {
         if (value != null ? !value.equals(that.value) : that.value != null) return false;
         if (parameters != null ? !parameters.equals(that.parameters) : that.parameters != null) return false;
         if (description != null ? !description.equals(that.description) : that.description != null) return false;
-        if (approvedAt != null ? !approvedAt.equals(that.approvedAt) : that.approvedAt != null) return false;
-        if (deniedAt != null ? !deniedAt.equals(that.deniedAt) : that.deniedAt != null) return false;
+        if (status != null ? !status.equals(that.status) : that.status != null) return false;
         return reasonDenied != null ? reasonDenied.equals(that.reasonDenied) : that.reasonDenied == null;
     }
 
@@ -270,8 +269,8 @@ public class ExamPrintRequest {
         result = 31 * result + pagePosition;
         result = 31 * result + (parameters != null ? parameters.hashCode() : 0);
         result = 31 * result + (description != null ? description.hashCode() : 0);
-        result = 31 * result + (approvedAt != null ? approvedAt.hashCode() : 0);
-        result = 31 * result + (deniedAt != null ? deniedAt.hashCode() : 0);
+        result = 31 * result + (changedAt != null ? changedAt.hashCode() : 0);
+        result = 31 * result + (status != null ? status.hashCode() : 0);
         result = 31 * result + (reasonDenied != null ? reasonDenied.hashCode() : 0);
         return result;
     }
--- a/client/src/main/java/tds/exam/ExamPrintRequest.java
+++ b/client/src/main/java/tds/exam/ExamPrintRequest.java
@@ -239,4 +239,40 @@ public class ExamPrintRequest {
     public String getReasonDenied() {
         return reasonDenied;
     }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        final ExamPrintRequest that = (ExamPrintRequest) o;
+
+        if (itemPosition != that.itemPosition) return false;
+        if (pagePosition != that.pagePosition) return false;
+        if (!examId.equals(that.examId)) return false;
+        if (!sessionId.equals(that.sessionId)) return false;
+        if (!type.equals(that.type)) return false;
+        if (value != null ? !value.equals(that.value) : that.value != null) return false;
+        if (parameters != null ? !parameters.equals(that.parameters) : that.parameters != null) return false;
+        if (description != null ? !description.equals(that.description) : that.description != null) return false;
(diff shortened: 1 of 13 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamPrintRequestQueryRepositoryImpl.java
```diff
@@ -63,4 +74,156 @@ public ExamPrintRequestQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final
...
+                "   FROM \n" +
+                "       exam.exam_print_request_event \n" +
+                "   GROUP BY exam_print_request_id \n" +
+                ") last_event \n" +
+                "   ON PR.id = last_event.exam_print_request_id \n" +
+                "JOIN exam.exam_print_request_event PRE \n" +
+                "   ON last_event.exam_print_request_id = PRE.exam_print_request_id \n" +
+                "   AND last_event.id = PRE.id \n" +
+                "WHERE \n" +
+                "   exam_id = :examId \n" +
```
millerts: Can we use aliases here as well?

service/src/main/java/tds/exam/repositories/impl/ExamPrintRequestQueryRepositoryImpl.java
```diff
@@ -63,4 +74,156 @@ public ExamPrintRequestQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final
...
+                "   FROM \n" +
+                "       exam.exam_print_request_event \n" +
+                "   GROUP BY exam_print_request_id \n" +
+                ") last_event \n" +
+                "   ON PR.id = last_event.exam_print_request_id \n" +
+                "JOIN exam.exam_print_request_event PRE \n" +
+                "   ON last_event.exam_print_request_id = PRE.exam_print_request_id \n" +
+                "   AND last_event.id = PRE.id \n" +
+                "WHERE \n" +
+                "   approved_at IS NOT NULL \n" +
```
millerts: Aliases here as well?

service/src/main/java/tds/exam/services/impl/ExamPrintRequestServiceImpl.java
```diff
@@ -34,4 +37,42 @@ public void insert(final ExamPrintRequest examPrintRequest) {
...
     }
+
+    @Override
+    public List<ExamPrintRequest> findUnfulfilledRequests(final UUID examId, final UUID sessionId) {
+        return examPrintRequestQueryRepository.findUnfulfilledRequests(examId, sessionId);
+    }
+
+    @Override
+    public void denyRequest(final UUID id, final String reason) {
+        final ExamPrintRequest request = new ExamPrintRequest.Builder(id)
```
millerts: Why don't we follow the same update path as when we approve?  When we approve we take the previous print request and update the approved at field.  On deny we have a different flow.  I would expect the same flow except in one instance we populate denied at instead of approved at.  With deny we lose any other data that was on the request prior to denied.

Also, why do we have two date fields for this?  In my mind we have three states.  `requested`, `approved`, `denied`.  If we need to know the date we can get it from created_at, right?  I don't have the whole picture of usages so maybe I'm missing how these dates would be used.  The added benefit to using a column named state or status would be we could collapse all approve/deny into a single flow.

service/src/main/java/tds/exam/web/endpoints/ExamPrintRequestController.java
```diff
@@ -27,4 +33,31 @@ public ExamPrintRequestController(final ExamPrintRequestService examPrintRequest
...
+
+    @RequestMapping(value = "/{sessionId}/{examId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<List<ExamPrintRequest>> findUnfulfilledRequests(final @PathVariable UUID examId, final @PathVariable UUID sessionId) {
+        return ResponseEntity.ok(examPrintRequestService.findUnfulfilledRequests(examId, sessionId));
+    }
+
+    @RequestMapping(value = "/deny/{id}", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void findUnfulfilledRequests(final @PathVariable UUID id, final @RequestBody String reason) {
+        examPrintRequestService.denyRequest(id, reason);
```
millerts: We check if we can find a print request and respond differently if we can't find the print request in the approve flow.  In this flow we'd just throw a 500 if the print request cannot be found based on the service implementation.  I don't think that is the intended flow, right?  Or am I missing something?

client/src/main/java/tds/exam/ExamPrintRequest.java
```diff
@@ -239,4 +239,40 @@ public Instant getDeniedAt() {
...
+        if (pagePosition != that.pagePosition) return false;
+        if (!examId.equals(that.examId)) return false;
+        if (!sessionId.equals(that.sessionId)) return false;
+        if (!type.equals(that.type)) return false;
+        if (value != null ? !value.equals(that.value) : that.value != null) return false;
+        if (parameters != null ? !parameters.equals(that.parameters) : that.parameters != null) return false;
+        if (description != null ? !description.equals(that.description) : that.description != null) return false;
+        if (approvedAt != null ? !approvedAt.equals(that.approvedAt) : that.approvedAt != null) return false;
+        if (deniedAt != null ? !deniedAt.equals(that.deniedAt) : that.deniedAt != null) return false;
+        return reasonDenied != null ? reasonDenied.equals(that.reasonDenied) : that.reasonDenied == null;
```
wtritch: Personal preference, but I like using:
```
return Objects.equals(itemPosition, that.itemPosition) &&
  Objects.equals(pagePosition, that.pagePosition) &&
  ...
```
It'll work for primitives/objects/nulls.

burzum619: I didn't write this, it was generated by IntelliJ. Is there a way to configure it to use that pattern?

service/src/main/java/tds/exam/web/endpoints/ExamPrintRequestController.java
burzum619: The legacy code simply does an update on the `testopprequest` record. In the case of approve, we're also fetching the ExamPrintRequest, which is needed by Proctor to generate the next view. For deny, no return value is necessary. It also saves us an addition database call to fetch the exam record (as we don't need any other info). 

However if you prefer to keep things consistent at the expense of the extra query call, I'm fine with making that change.

burzum619: Also, I noticed/fixed the controller method name not being correct here.

service/src/main/java/tds/exam/services/impl/ExamPrintRequestServiceImpl.java
burzum619: For the first point, please see my followup comment to your comment on /deny.

For your second point, we could make that change if you'd prefer, but I'm not seeing much of a benefit to it personally, other than we remove an extra column. I think the flow for deny/approve will have to be different because approve requires the request object to be returned while deny does not.

service/src/main/java/tds/exam/repositories/impl/ExamPrintRequestQueryRepositoryImpl.java
```diff
@@ -63,4 +74,156 @@ public ExamPrintRequestQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final
...
+                "   PR.type, \n" +
+                "   PR.value, \n" +
+                "   PR.item_position, \n" +
+                "   PR.page_position, \n" +
+                "   PR.parameters, \n" +
+                "   PR.description, \n" +
+                "   PR.created_at, \n" +
+                "   PRE.approved_at, \n" +
+                "   PRE.denied_at, \n" +
+                "   PRE.reason_denied \n" +
```
wtritch: I know we don't have a pattern for slicing and dicing our SQL, but I wish there were a way to avoid this duplication.  Even just a `public static final String PRINT_REQUEST_SELECT` might help if we ever add/remove properties.

service/src/main/java/tds/exam/web/endpoints/ExamPrintRequestController.java
```diff
@@ -27,4 +33,31 @@ public ExamPrintRequestController(final ExamPrintRequestService examPrintRequest
...
+    }
+
+    @RequestMapping(value = "/approve/{id}", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<ExamPrintRequest> findAndApprovePrintRequest(final @PathVariable UUID id) {
+        Optional<ExamPrintRequest> maybeExamPrintRequest = examPrintRequestService.findAndApprovePrintRequest(id);
+
+        if (!maybeExamPrintRequest.isPresent()) {
+            return new ResponseEntity<>(HttpStatus.UNPROCESSABLE_ENTITY);
```
wtritch: Should this instead throw a 404?

service/src/main/java/tds/exam/repositories/impl/ExamPrintRequestQueryRepositoryImpl.java
burzum619: Moving this SELECT ... statements into a static const string at the class level similar to what we do with Exam.java

client/src/main/java/tds/exam/ExamPrintRequest.java
burzum619: I actually have been looking for a way to configure IntelliJ to use the `Objects` static helpers for a while now, but i haven't quite figured it out.

wtritch: I don't believe so... At least, not that I'm aware of.  I usually hand-jam it if I'm overwriting equals/hashcode =)

service/src/main/java/tds/exam/repositories/impl/ExamPrintRequestQueryRepositoryImpl.java
burzum619: Yep, I just did exactly that actually, while I was aliasing the column names. Thanks!!

service/src/main/java/tds/exam/web/endpoints/ExamPrintRequestController.java
burzum619: Yep, that's more appropriate.

client/src/main/java/tds/exam/ExamPrintRequest.java
burzum619: Masochist! Manually writing boilerplate code is my bane

service/src/main/java/tds/exam/web/endpoints/ExamPrintRequestController.java
wtritch: Just my 2 cents but:
1. We shouldn't be spam-denying print requests so often that an additional query by id would be noticeable.
2. The client might appreciate knowing they just tried to deny a non-existent request (return 404 if not found)
3. It gives us a quick opportunity to check if it's already been denied, if we want to avoid inserting multiple denials (Maybe we want to do that for approvals, as well?)
4. I like consistency.  =)

client/src/main/java/tds/exam/ExamPrintRequest.java
wtritch: Yeah... I usually don't have to do it on classes with 10+ properties =/

service/src/main/java/tds/exam/services/impl/ExamPrintRequestServiceImpl.java
millerts: I think removing unnecessary columns has value regardless of flow.  A request can only be in a single state at any given time.  I wouldn't represent state as two different fields in a class so I wouldn't do it in a DB table.

client/src/main/java/tds/exam/ExamPrintRequest.java
millerts: I agree with Ernie on this one.  Boilerplate or let's use AutoValue which does it for us.

burzum619: Good point, I keep forgetting about Autovalue

PR #106, opened 2017-03-13
Bugfix/completestatus permeable update

jtreuting:
Updated PR.  Main points include:

- Pulled verify access logic from service to VerifyAccess annotation on controller endpoints
- Changed ExamSegments endpoints to follow the /exam/{examId} convention
- Updated services to remove verify access check and only pass in examId instead of ExamInfo since that was just for doing the verify access checks in the service
- Removed tests from services that was testing the verify access checks
- Added VerifyAccessInterceptor tests 
- Fixed the original bug in Complete status to update all appropriate segments with test coverage for that

Note: A separate PR will be coming to update calls to the ExamSegmentsController since the URL structure chagned

### DIFF
--- a/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
+++ b/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
@@ -5,7 +5,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
-import tds.exam.utils.VerifyAccessInterceptor;
+import tds.exam.web.interceptors.VerifyAccessInterceptor;
 
 @Configuration
 public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
--- a/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
+++ b/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
@@ -1,7 +1,6 @@
 package tds.exam.configuration.web;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
@@ -9,6 +8,9 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter
 import tds.exam.services.ExamApprovalService;
 import tds.exam.web.interceptors.VerifyAccessInterceptor;
 
+/**
+ * Configure additional interceptors for Spring
+ */
 @Configuration
 public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
     private final ExamApprovalService examApprovalService;
@@ -20,10 +22,6 @@ public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
 
     @Override
     public void addInterceptors(InterceptorRegistry registry) {
-        registry.addInterceptor(verifyAccessInterceptor(examApprovalService)).addPathPatterns("/exam/**");;
-    }
-
-    public VerifyAccessInterceptor verifyAccessInterceptor(ExamApprovalService examApprovalService) {
-        return new VerifyAccessInterceptor(examApprovalService);
+        registry.addInterceptor(new VerifyAccessInterceptor(examApprovalService)).addPathPatterns("/exam/**");;
     }
 }
--- a/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
+++ b/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
@@ -1,21 +1,29 @@
 package tds.exam.configuration.web;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
+import tds.exam.services.ExamApprovalService;
 import tds.exam.web.interceptors.VerifyAccessInterceptor;
 
 @Configuration
 public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
-    @Override
-    public void addInterceptors(InterceptorRegistry registry) {
-        registry.addInterceptor(verifyAccessInterceptor()).addPathPatterns("/**");;
+    private final ExamApprovalService examApprovalService;
+
+    @Autowired
+    public InterceptorConfiguration(ExamApprovalService examApprovalService) {
+        this.examApprovalService = examApprovalService;
     }
 
-    @Bean
-    public VerifyAccessInterceptor verifyAccessInterceptor() {
-        return new VerifyAccessInterceptor();
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        registry.addInterceptor(verifyAccessInterceptor(examApprovalService)).addPathPatterns("/exam/**");;
+    }
+
+    public VerifyAccessInterceptor verifyAccessInterceptor(ExamApprovalService examApprovalService) {
+        return new VerifyAccessInterceptor(examApprovalService);
     }
 }
--- a/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
+++ b/service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
@@ -1,0 +1,21 @@
+package tds.exam.configuration.web;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+
+import tds.exam.utils.VerifyAccessInterceptor;
+
+@Configuration
+public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        registry.addInterceptor(verifyAccessInterceptor()).addPathPatterns("/**");;
+    }
+
+    @Bean
+    public VerifyAccessInterceptor verifyAccessInterceptor() {
+        return new VerifyAccessInterceptor();
+    }
+}
--- a/service/src/main/java/tds/exam/utils/VerifyAccessInterceptor.java
+++ b/service/src/main/java/tds/exam/utils/VerifyAccessInterceptor.java
@@ -10,11 +10,10 @@ import java.util.Map;
 import java.util.UUID;
 
 import tds.common.Response;
-import tds.common.ValidationError;
-import tds.common.web.exceptions.ValidationException;
 import tds.exam.ExamApproval;
 import tds.exam.ExamInfo;
 import tds.exam.services.ExamApprovalService;
+import tds.exam.web.exceptions.ValidationException;
 
 public class VerifyAccessInterceptor extends HandlerInterceptorAdapter {
     @Autowired
@@ -56,8 +55,7 @@ public class VerifyAccessInterceptor extends HandlerInterceptorAdapter {
         Response<ExamApproval> approval = examApprovalService.getApproval(new ExamInfo(examId, sessionId, browserId));
 
         if (approval.getError().isPresent()) {
-            ValidationError validationError = approval.getError().get();
-            throw new ValidationException(validationError.getCode(), String.format("VerifyAccess: %s", validationError.getMessage()));
+            throw new ValidationException(approval.getError().get());
         }
 
         return true;
--- a/service/src/main/java/tds/exam/utils/VerifyAccessInterceptor.java
+++ b/service/src/main/java/tds/exam/utils/VerifyAccessInterceptor.java
@@ -1,0 +1,118 @@
+package tds.exam.utils;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.http.MediaType;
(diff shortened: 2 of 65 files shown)


service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -51,15 +54,20 @@ public void accept(final Exam oldExam, final Exam newExam) {
...
+            List<ExamSegment> examSegments = examSegmentsResponse.getData().get().stream()
+                .filter(examSegment -> examSegment.isPermeable())
+                .map(examSegment -> ExamSegment.Builder
+                    .fromSegment(examSegment)
+                    .withPermeable(false)
+                    .build()
+                )
+                .collect(Collectors.toList());
+
+            examSegmentService.update(examSegments.toArray(new ExamSegment[examSegments.size()]));
```
jtreuting: I'm following the pattern i've been seeing, but seems like we should probably have a `ExamSegmentService.updateSegmentPermeability(bool)` (that is a terrible name, but you get the idea), that just does a single SQL statement that takes the last exam_segment_event rows that need to change and inserts new ones based on those, with the new permeable flag.

I'm happy to do it, but didn't want to have it be the one place it was doing it differently.

Thoughts?

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -51,15 +54,20 @@ public void accept(final Exam oldExam, final Exam newExam) {
...
 
-        if (maybeExamSegment.isPresent()) {
-            examSegmentService.update(ExamSegment.Builder
-                .fromSegment(maybeExamSegment.get())
-                .withPermeable(false)
-                .build());
+        if (examSegmentsResponse.getData().isPresent()) {
+            List<ExamSegment> examSegments = examSegmentsResponse.getData().get().stream()
+                .filter(examSegment -> examSegment.isPermeable())
+                .map(examSegment -> ExamSegment.Builder
```
jtreuting: Filtering to only get the ones that will actually change from true to false

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
millerts: I'm fine with it, but I'd want to ensure that we're going to get a performance benefit to deviate from consistency.  I'm curious if we rewrote our insert statements to do bulk inserts rather than batch inserts if we'd see similar benefits.

I also wouldn't want to add lots of different update methods that update n number of fields whenever we have a new update group.

jeffjohnson9046: I think it depends: how often will we be updating a segment's permeability?  The only other place I can think of is when a student exits a segment; TDS _might_ mark the segment as impermeable so the student can't go back to the previous segment again (I'm not sure off the top of my head).  If we find ourselves updating a segment's permeability more often than we update other fields, it might be worth having a special update method to update just permeability.

jtreuting: Well we will get a performance improvement because right now it does a select, and then a batch insert or multiple inserts (depends on connection string setting).  So it would essentially remove the select part of that.  I did it this way for consistency for sure and I'd say we hold off on doing it a different way until we get more into specific optimizations.  but something that popped into my head while writing it.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -51,15 +51,20 @@ public void accept(final Exam oldExam, final Exam newExam) {
         }
 
         // CommonDLL#_OnStatus_Completed_SP, line 1425: Update the exam to indicate this segment is not permeable,
-        // meaning the segment cannot be accessed/visited again.  Legacy code sets isPermeable to -1
-        Optional<ExamSegment> maybeExamSegment = examSegmentService.findByExamIdAndSegmentPosition(newExam.getId(),
-            newExam.getCurrentSegmentPosition());
+        // meaning the segment cannot be accessed/visited again.  Legacy code sets isPermeable to -1 for all segments
+        Response<List<ExamSegment>> examSegmentsResponse = examSegmentService.findExamSegments(newExam.getId(), newExam.getSessionId(), newExam.getBrowserId());
```
jtreuting: Yeah I like that.  I honestly didn't look in the method itself, but it makes sense as to why those other fields are passed in.  I'll create a new method that can be used by this existing one and not have the verify in it.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
millerts: Semi related, we aren't writing our inserts properly.  I think we should be doing bulk inserts rather than batching a bunch of insert statements because they are faster.  

It doesn't change your argument for not doing the select.  However, I'd like to see numbers to see if the performance gains are worth the deviation.  

Bulk insert example:
```
insert (a, b, c)
values
(1, 2, 3),
(4, 5, 6),
(7, 8, 9);
```

jtreuting: @mtrupkin did that test and it showed some big improvements.  As far as I know with JDBC Template you can't do that and you need to rely on the driver connection string parameter `rewriteBatchedStatements=true` (https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html).

Also, I tried setting that flag to address TDS-729 (https://jira.fairwaytech.com/browse/TDS-729) but it produced SQL that had an error in it when accommodations were inserted.  I didn't get a chance to look into it too much but that is obviously concerning that it could produce bad SQL with that setting.

millerts: We can do batch inserts (example [here](https://gist.github.com/millerts/770bf7ec5f1e0f39123135c6de6a3121)), but it requires rewriting our queries and populating the parameters differently than we are today.

service/src/main/java/tds/exam/utils/VerifyAccessInterceptor.java
```diff
@@ -0,0 +1,118 @@
...
+    @Autowired
+    private ExamApprovalService examApprovalService;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+//    @Autowired
+//    public VerifyAccessInterceptor(final ExamApprovalService examApprovalService) {
+//        this.examApprovalService = examApprovalService;
+//    }
```
wtritch: Interesting, why not use constructor injection?  You wouldn't even need to \@Autowire them in here, just add them as autowired parameters to your verifyAccessInterceptor() \@Bean method above.

service/src/main/java/tds/exam/configuration/web/InterceptorConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Configure additional interceptors for Spring
+ */
+@Configuration
+public class InterceptorConfiguration extends WebMvcConfigurerAdapter {
+    private final ExamApprovalService examApprovalService;
+
+    @Autowired
+    public InterceptorConfiguration(ExamApprovalService examApprovalService) {
```
wtritch: nit: finals =)

service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
```diff
@@ -28,15 +27,12 @@ public ExamItemController(ExamItemService examItemService) {
         this.examItemService = examItemService;
     }
 
-    //@RequestMapping(value = "/{id}/page/{position}/responses", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
     @PostMapping("/{id}/page/{position}/responses")
```
wtritch: Do you still need the "produces" value here?  Or are we going to allow Spring to handle that via Accept request headers?

service/src/main/java/tds/exam/web/endpoints/ExamPageController.java
```diff
@@ -27,13 +26,11 @@ public ExamPageController(ExamPageService examPageService) {
         this.examPageService = examPageService;
     }
 
+    @VerifyAccess
     @RequestMapping(value = "/{id}/page/{position}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
wtritch: Do we want to switch these to GetMapping or leave it as personal preference?

service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
```diff
@@ -7,21 +7,20 @@
...
-import tds.common.Response;
 import tds.common.ValidationError;
 import tds.common.web.resources.NoContentResponseResource;
 import tds.exam.ExamSegment;
 import tds.exam.services.ExamSegmentService;
+import tds.exam.web.annotations.VerifyAccess;
 
 @RestController
-@RequestMapping("/exam/segments")
+@RequestMapping("/exam")
```
wtritch: I believe you could do "/exam/{examId}/segments" here if you want to reduce duplication and keep the controller focused on a defined subset of endpoints.

service/src/main/java/tds/exam/web/exceptions/ValidationException.java
```diff
@@ -0,0 +1,38 @@
+package tds.exam.web.exceptions;
+
+import tds.common.ValidationError;
+
+public class ValidationException extends Exception {
```
wtritch: nit: javadocs?

service/src/main/java/tds/exam/web/interceptors/VerifyAccessInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Verifies access for a user based on the ExamId, BrowserId and SessionId.
+ * This is used when a controller endpoint uses the {@link tds.exam.web.annotations.VerifyAccess} annotation
+ * The controller URL must follow the convention /exam/{examId}/[path]?sessionId=xxx&browserId=yyy
+ */
+public class VerifyAccessInterceptor extends HandlerInterceptorAdapter {
+    private ExamApprovalService examApprovalService;
+
+    @Autowired
```
wtritch: Since you're manually constructing this, I think you can remove the @Autowired from here.

service/src/main/java/tds/exam/web/interceptors/VerifyAccessInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+    public VerifyAccessInterceptor(final ExamApprovalService examApprovalService) {
+        this.examApprovalService = examApprovalService;
+    }
+
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
```
wtritch: nit: finals =)

wtritch: Also, recommend changing the throws to ValidationException and/or whatever UUID parsing throws if the @Override will let you...

service/src/main/java/tds/exam/web/interceptors/VerifyAccessInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+
+        if (verifyAccessAnnotation == null) {
+            return true;
+        }
+
+        // @VerifyAccess will only work on URLs that follow the convention /exam/{examId}/...., otherwise the examId won't be found
+        String[] pathParts = request.getRequestURI().split("/");
+        if (!pathParts[1].equals("exam")) {
+            throw new IllegalArgumentException(String.format("VerifyAccess: Exam ID could not be found for url %s.", request.getRequestURI()));
+        }
```
wtritch: Hrm... if we ever start versioning our apis (/api/v2/exams/{examId}/...) I suppose we could revisit this.  The alternative would be to march down the pathParts looking for idx["exam"]+1

service/src/main/java/tds/exam/web/interceptors/VerifyAccessInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+
+        Map<String, String[]> parameters = request.getParameterMap();
+
+        if (parameters.containsKey(verifyAccessAnnotation.sessionParamName())) {
+            sessionId = UUID.fromString(parameters.get(verifyAccessAnnotation.sessionParamName())[0]);
+        }
+
+        if (parameters.containsKey(verifyAccessAnnotation.browserParamName())) {
+            browserId = UUID.fromString(parameters.get(verifyAccessAnnotation.browserParamName())[0]);
+        }
```
wtritch: If sessionId or browserId are null/missing/invalid UUIDs should we return true or false without doing a db lookup?

jtreuting: We don't, but I didn't want to duplicate the error messages from here: https://github.com/SmarterApp/TDS_ExamService/blob/bugfix/completestatus-permeable-update/service/src/main/java/tds/exam/services/impl/ExamApprovalServiceImpl.java#L61

But I could rework the ExamApprovalService.getApproval to do those checks before the Exam lookup so it doesn't hit the DB and I don't have to duplciate the error messages in here

service/src/main/java/tds/exam/web/interceptors/VerifyAccessInterceptor.java
jtreuting: finals, will do.

The method itself can throw ValidationException, but ExamApprovalService can throw I can do `throws ValidationException, IllegalArgumentException`

service/src/main/java/tds/exam/web/exceptions/ValidationException.java
jtreuting: doh, missed one

service/src/main/java/tds/exam/web/endpoints/ExamItemController.java
jtreuting: I didn't change this, and everything was working.  I just got rid of the commented out code since it shouldn't be left around

service/src/main/java/tds/exam/web/endpoints/ExamPageController.java
jtreuting: I think we want to switch, but probably a later refactor across the board as a PR would be appropriate for that

PR #108, opened 2017-03-14
TDS-671: Feature/remove timestamp default

jeffjohnson9046:
[TDS-671](https://jira.fairwaytech.com/browse/TDS-671)
* Identify all tables in `exam` database that have a column w/a `DEFAULT` constraint that uses a `TIMESTAMP`:
```
SELECT 
	* 
FROM 
	INFORMATION_SCHEMA.COLUMNS
WHERE
	TABLE_SCHEMA = 'exam'
	AND COLUMN_DEFAULT LIKE '%TIMESTAMP%';
``` 
* Add database migration to remove all `DEFAULT %TIMESTAMP%` constraints
  * Only affected `created_at` columns, which are still `NOT NULL`
  * Omitted the `schema_version` table which is owned by flyway and not used by the application
* Update `insert`, `update`, `delete` methods to set the `created_at` to `Instant.now()`
* Update unit/integration tests

### DIFF
--- a/client/src/main/java/tds/exam/ExamAccommodation.java
+++ b/client/src/main/java/tds/exam/ExamAccommodation.java
@@ -87,7 +87,7 @@ public class ExamAccommodation {
         }
 
         public Builder withCreatedAt(Instant createdAt) {
-            this.createdAt = createdAt;
+            this.createdAt = checkNotNull(createdAt, "createdAt cannot be null");
             return this;
         }
 
--- a/client/src/main/java/tds/exam/ExamAccommodation.java
+++ b/client/src/main/java/tds/exam/ExamAccommodation.java
@@ -87,7 +87,7 @@ public class ExamAccommodation {
         }
 
         public Builder withCreatedAt(Instant createdAt) {
-            this.createdAt = checkNotNull(createdAt, "createdAt cannot be null");
+            this.createdAt = createdAt;
             return this;
         }
 
--- a/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
@@ -8,9 +8,9 @@ import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.stereotype.Repository;
 
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
 import tds.exam.ExamAccommodation;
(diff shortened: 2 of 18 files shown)


client/src/main/java/tds/exam/ExamAccommodation.java
```diff
@@ -87,7 +87,7 @@ public Builder withDeniedAt(Instant deniedAt) {
         }
 
         public Builder withCreatedAt(Instant createdAt) {
-            this.createdAt = checkNotNull(createdAt, "createdAt cannot be null");
```
millerts: Shouldn't this stay?  When is setting a null `createdAt` ever valid?

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
```diff
@@ -59,14 +71,16 @@ private void updateEvent(final ExamAccommodation... examAccommodations) {
             "deleted_at, " +
             "selectable," +
             "custom," +
-            "total_type_count) \n" +
+            "total_type_count," +
```
millerts: I think this private method should take in an Instant that you set when you created the initial record.  The public update method can pass in the Instant.now().  There are two benefits, the records will have the same created at when initial records are created (examAccommodation and event).  And all the created records will have the same created at date.  For example, all 5 accommodations being created/updated through this flow will all have the same createdAt date.

service/src/main/java/tds/exam/repositories/impl/ExamCommandRepositoryImpl.java
```diff
@@ -107,7 +110,8 @@ public void update(final Exam... exams) {
                 .addValue("waitingForSegmentApprovalPosition", exam.getWaitingForSegmentApprovalPosition())
                 .addValue("currentSegmentPosition", exam.getCurrentSegmentPosition())
                 .addValue("customAccommodations", exam.isCustomAccommodations())
-                .addValue("startedAt", mapJodaInstantToTimestamp(exam.getStartedAt())))
+                .addValue("startedAt", mapJodaInstantToTimestamp(exam.getStartedAt()))
+                .addValue("createdAt", mapJodaInstantToTimestamp(Instant.now())))
```
millerts: I think the Instant should be created outside of this loop and you may as well do the mapJodaInstantToTimestamp when it is created since the Instant is never outside of passing it into this static mapping class.

client/src/main/java/tds/exam/ExamAccommodation.java
jeffjohnson9046: Longer term, I think the right thing to do is remove the `createdAt` fields from our entities altogether.  With this PR, we won't be "honoring" whatever the caller sets the `createdAt` field to; we'll be generating the `createdAt` date in the repository layer during the `insert`/`update`/`delete` methods.

However, for now you're right; if someone wants to set the `createdAt` date, we should enforce the rule that it cannot be set to null.

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
jeffjohnson9046: Good catch.

jeffjohnson9046:
I agree with @millerts : we should probably have private `updateEvent` methods (an example can be seen [here]()).  I like the idea that the `createdAt` timestamp for the `%_event` record matches the `createdAt` for the "parent" record.  I don't think this would be too difficult to implement; I can do it as part of this PR or we can create a separate story.

PR #113, opened 2017-03-17
TDS-765: Add exam/item comments

jeffjohnson9046:
[TDS-765](https://jira.fairwaytech.com/browse/TDS-765):  Allow for persisting/fetching comments for `Exam`s or `ExamItem`s via the notepad tool in the student UI.

### Notes
* Referred to as "comments" in legacy code
* In the legacy Student application, Item-level notes are not retrievable.
  * If a user creates a comment on an item then refreshes the page, the item comment will no longer be available
  * The item comment will be available if the user uses the provided navigation controls to switch between pages
  * This PR will also store item-level notes, but does not provide an endpoint to retrieve them
* Item-level and Exam-level notes are an optional element within the TRT
* There can only be one Exam-level note
  * In the event there are multiple records for an exam-level note on a single exam, the most recent record is returned
* The TRT spec says that comments are a data element that Smarter Balanced isn't interested in ([TRT spec](http://www.smarterapp.org/documents/TestResults-DataModel.pdf), appendix B, p. 21), but there are `<comment>` elements in their [sample TRT output](http://www.smarterapp.org/documents/TestResultsTransmissionFormat_Sample.xml)
* Legacy `session.testeecomment` acts like our `_event` tables; records in `session.testeecomment` are never updated; new records are inserted when a user changes a note
* QA view included for comparing `exam.examinee_note` to `session.testeecomment`

### DIFF
--- a/client/src/main/java/tds/exam/ExamineeNote.java
+++ b/client/src/main/java/tds/exam/ExamineeNote.java
@@ -74,13 +74,13 @@ public class ExamineeNote {
             return new ExamineeNote(this);
         }
 
-        public Builder fromExamineeNote(ExamineeNote examineeNote) {
-            id = examineeNote.id;
-            examId = examineeNote.examId;
-            context = examineeNote.context;
-            itemPosition = examineeNote.itemPosition;
-            note = examineeNote.note;
-            return this;
+        public static Builder fromExamineeNote(ExamineeNote examineeNote) {
+            return new ExamineeNote.Builder()
+                .withId(examineeNote.id)
+                .withExamId(examineeNote.examId)
+                .withContext(examineeNote.context)
+                .withItemPosition(examineeNote.itemPosition)
+                .withNote(examineeNote.note);
         }
     }
 
--- a/client/src/main/java/tds/exam/ExamineeNote.java
+++ b/client/src/main/java/tds/exam/ExamineeNote.java
@@ -1,0 +1,126 @@
+package tds.exam;
+
+import java.util.UUID;
+
+import tds.common.util.Preconditions;
+
+
+/**
+ * Represents a comment/note made via the Notepad tool in the Student user interface
+ */
+public class ExamineeNote {
+    private long id;
+    private UUID examId;
+    private ExamineeNoteContext context;
+    private int itemPosition;
+    private String note;
+
+    /**
+     * Private constructor for frameworks
+     */
+    private ExamineeNote() {
+    }
+
+    private ExamineeNote(Builder builder) {
+        this.id = builder.id;
+        this.examId = builder.examId;
+        this.context = builder.context;
+        this.itemPosition = builder.itemPosition;
+        this.note = builder.note;
+    }
+
+    public static class Builder {
+        private long id;
+        private UUID examId;
+        private ExamineeNoteContext context;
+        private int itemPosition;
+        private String note;
+
+        public Builder withId(final long id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder withExamId(final UUID examId) {
+            Preconditions.checkNotNull(examId, "examId cannot be null");
+            this.examId = examId;
+            return this;
+        }
+
+        public Builder withContext(final ExamineeNoteContext context) {
+            Preconditions.checkNotNull(context, "context cannot be null");
+            this.context = context;
+            return this;
+        }
+
+        public Builder withItemPosition(final int itemPosition) {
+            this.itemPosition = itemPosition;
+            return this;
+        }
+
+        public Builder withNote(final String note) {
+            Preconditions.checkNotNull(note, "note cannot be null");
+            this.note = note;
+            return this;
+        }
+
+        public ExamineeNote build() {
+            // RULE:  If the note is in the context of an exam, the position should be set to 0.  Item position is not
+            // relevant to Exam-level notes
+            if (this.context.equals(ExamineeNoteContext.EXAM) && this.itemPosition > 0) {
+                throw new IllegalStateException("itemPosition must be 0 for an ExamineeNote within an EXAM-level context");
+            }
+
+            return new ExamineeNote(this);
+        }
+
+        public Builder fromExamineeNote(ExamineeNote examineeNote) {
+            id = examineeNote.id;
+            examId = examineeNote.examId;
+            context = examineeNote.context;
+            itemPosition = examineeNote.itemPosition;
+            note = examineeNote.note;
+            return this;
+        }
+    }
+
+    /**
+     * @return The unique identifier of this {@link ExamineeNote}
+     */
+    public long getId() {
+        return id;
+    }
+
+    /**
+     * @return The unique identifier of the {@link tds.exam.Exam} to which this {@link ExamineeNote} belongs
+     */
+    public UUID getExamId() {
+        return examId;
+    }
+
+    /**
+     * @return The {@link ExamineeNoteContext} describing what this {@link tds.exam.ExamineeNote} is associated
+     * to
+     */
+    public ExamineeNoteContext getContext() {
+        return context;
+    }
+
+    /**
+     * @return The position of the {@link tds.exam.ExamItem} for which this {@link ExamineeNote} was created
+     * <p>
+     * This property is only relevant for an item-scoped {@link tds.exam.ExamineeNote}.  An exam-scoped
+     * {@link tds.exam.ExamineeNote} will always have an item position of 0.
+     * </p>
+     */
+    public int getItemPosition() {
+        return itemPosition;
+    }
+
+    /**
+     * @return The text for this {@link ExamineeNote}
+     */
+    public String getNote() {
+        return note;
+    }
+}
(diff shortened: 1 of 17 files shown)


client/src/main/java/tds/exam/ExamineeNote.java
```diff
@@ -0,0 +1,126 @@
...
+
+import java.util.UUID;
+
+import tds.common.util.Preconditions;
+
+
+/**
+ * Represents a comment/note made via the Notepad tool in the Student user interface
+ */
+public class ExamineeNote {
```
millerts: Suggest using `@AutoValue` on new code.  No need to go back and change it now.

client/src/main/java/tds/exam/ExamineeNote.java
```diff
@@ -0,0 +1,126 @@
...
+            // RULE:  If the note is in the context of an exam, the position should be set to 0.  Item position is not
+            // relevant to Exam-level notes
+            if (this.context.equals(ExamineeNoteContext.EXAM) && this.itemPosition > 0) {
+                throw new IllegalStateException("itemPosition must be 0 for an ExamineeNote within an EXAM-level context");
+            }
+
+            return new ExamineeNote(this);
+        }
+
+        public Builder fromExamineeNote(ExamineeNote examineeNote) {
```
millerts: Can you move this out of the builder and make it static?  After being done someone should be able to do `ExamineeNote.fromExamineeNote().withSomething()`

service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
```diff
@@ -0,0 +1,57 @@
...
+    @RequestMapping(value = "/{id}/note", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
+    @VerifyAccess
+    ResponseEntity<ExamineeNote> getNoteInExamContext(@PathVariable final UUID id) {
+        Optional<ExamineeNote> maybeExamineeNote = examineeNoteService.findNoteInExamContext(id);
+
+        // It is possible that there is no note for an exam.  When that is the case, return a NO CONTENT response,
+        // indicating the call was successful but there was nothing to get.
+        return maybeExamineeNote.isPresent()
+            ? ResponseEntity.ok(maybeExamineeNote.get())
+            : new ResponseEntity<>(HttpStatus.NO_CONTENT);
```
millerts: I see your comment, but I think this should be a 404 since the resource isn't found.  The client can decide how they want to respond if a 404 is returned.

service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @RequestMapping(value = "/{id}/note", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: I think this should be a `POST` not a `PUT` since you're always creating a new resource.  It is a bit odd since there is only single note ever for an exam.  I look at is as you are creating a new resource and the old resource is no more.

service/src/main/resources/db/migration/V1489696880__exam_create_examinee_note_table.sql
```diff
@@ -0,0 +1,37 @@
...
+***********************************************************************************************************************/
+
+USE exam;
+
+DROP TABLE IF EXISTS examinee_note;
+
+CREATE TABLE examinee_note(
+  id BIGINT NOT NULL AUTO_INCREMENT,
+  exam_id CHAR(36) NOT NULL,
+  context VARCHAR(11) NOT NULL,
```
millerts: We haven't been doing this yet, but do you think we should add a constraint on this column to only allow certain values since the application will blow up if you don't use the right ones (enum throws)?

service/src/test/java/tds/exam/repositories/impl/ExamineeNoteQueryRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,124 @@
...
+        assertThat(result.getId()).isGreaterThan(0L);
+        assertThat(result).isEqualToComparingOnlyGivenFields(note,
+            "examId",
+            "context",
+            "itemPosition",
+            "note");
+    }
+
+    @Test
+    public void shouldOnlyReturnExamScopedNoteWhenExamHasMultipleNotes() {
```
millerts: I think we should combine this test with the one above.  It seems both satisfy the same flow.  And maybe created a third that has a different exam id?

service/src/test/java/tds/exam/services/impl/ExamineeNoteServiceImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Before
+    public void setup() {
+        examineeNoteService = new ExamineeNoteServiceImpl(mockExamineeNoteCommandRepository,
+            mockExamineeNoteQueryRepository);
+    }
+
+    @Test
+    public void shouldGetAnExamineeNote() {
+        ExamineeNote note = new ExamineeNote.Builder()
```
millerts: No need to change it now, but don't forget that nice random value generator project @burzum619 showed us

service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
jeffjohnson9046: From a behavior perspective, I see it as creating a new exam-level note or updating an existing exam-level note.  True enough - the server always inserts a new record, even when updating an existing one.  Both PUT and POST can be used to create/update resources, but I think PUTs are defined to be idempotent where the same is not necessarily true of POSTs.  I flipped and flopped on the verb for awhile, but PUT seemed more natural to me given the behavior (as opposed to the actual mechanics).

jeffjohnson9046:
@millerts : In response to your general comment:  there is no way to retrieve item-level comments/notes.  Item-level notes are stored in the database, but there is no endpoint to retrieve them.  Display of the item-level notes is driven entirely by client-side javascript.

#### Note/Comment Use Cases
* If a user creates a note on an item and uses the provided navigation controls to switch between pages, then the item-level notes are visible
* If a user creates a note on an item then refreshes the browser, the item-level note will no longer be visible
* If a user creates a note on an item, pauses their exam and then resumes the same exam (all within the same session), the item-level notes will not be visible
* If a user creates a note on the exam and uses the provided navigation controls to switch between pages, the exam-level notes are visible
* If a user creates a note on the exam and refreshes the browser, the exam-level note will be visible
  * any item-level notes will no longer be visible
* If a user creates a note on the exam, pauses their exam and then resumes the same exam (all within the same session), the exam-level note will still be visible

millerts:
I didn't realize Student UI was caching the notes.

service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
wtritch: Hrm... I can see both arguments.  But if there were a list of /notes available for an exam, I'd say POST is more natural.  But this is closer to setting/updating a single attribute on an Exam, which feels more like an update (PUT.)  That being said, in a pure REST context, "/{id}/note" is still a Resource, even if it's a singleton resource on "/{id}/"  I think I agree with Todd that a POST to create, and a PUT/PATCH to update would be more consistent. (Yes, I know we'll instead keep POSTing rather than messing with PUT or JsonPATCH.)

service/src/main/resources/db/migration/V1489696880__exam_create_examinee_note_table.sql
wtritch: I feel like that's moving business logic into the DB layer...

service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
jeffjohnson9046: I stewed on this for a minute, and I agree.  If someone's looking for a note and it isn't there, then 404 is the correct response.  The caller can decide if that's an error condition or not.

service/src/main/resources/db/migration/V1489696880__exam_create_examinee_note_table.sql
jeffjohnson9046: It looks like `CHECK` constraints in MySQL don't work the way I would expect.  From [the documentation](https://dev.mysql.com/doc/refman/5.7/en/create-table.html):

> The CHECK clause is parsed but ignored by all storage engines.

This is the test I conducted:
```
ALTER TABLE exam.examinee_note
	ADD CONSTRAINT chk_examinee_note_context CHECK (`context` IN ('GlobalNotes', 'TESTITEM'));

-- Create parent record to satisfy foreign key constraint
INSERT INTO exam.exam(id, client_name, environment, subject, student_id, assessment_id, assessment_key, assessment_algorithm, created_at)
VALUES('835c0e17-4646-4d9d-9e85-e169b7b6f764', 'client_name', 'test', 'ELA', 1, 'assessment_id', 'assessment_key', 'adaptive2', UTC_TIMESTAMP());

-- Should fail, but doesn't; the record will be inserted even though the context value of 'bar' violates the CHECK constraint on examinee_note
INSERT INTO exam.examinee_note (exam_id, context, note)
VALUES('835c0e17-4646-4d9d-9e85-e169b7b6f764', 'bar', 'test');
```

millerts: Yeah, I think the proper way to do it would be to have a `note_context` table that would be a FK relationship to the `context` column.  I don't think we should do that for this work.

PR #114, opened 2017-03-20
Check if segments completed - exam side

burzum619:
Endpoint for checking if all exam_segments for an exam are completed/satisfied.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
@@ -200,4 +200,10 @@ public class ExamSegmentServiceImpl implements ExamSegmentService {
 
         return Optional.empty();
     }
+
+    @Override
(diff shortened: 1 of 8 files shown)


service/src/main/java/tds/exam/services/impl/ExamSegmentServiceImpl.java
```diff
@@ -200,4 +200,10 @@ public void update(final ExamSegment... examSegments) {
 
         return Optional.empty();
     }
+
+    @Override
+    public boolean checkIfSegmentsCompleted(final UUID examId) {
+        return !examSegmentQueryRepository.findByExamId(examId).stream()
```
millerts: Wouldn't it be faster to add a query to the `ExamSegementQueryRepository` to select the count of records with the isSatisfied set to false?  If the count is greater than 1 then return false.

burzum619: Sure, I can create a new repository method to do the count in SQL.

PR #115, opened 2017-03-21
Skip duplicate accommodation types

burzum619:
This bugfix is for https://jira.fairwaytech.com/browse/TDS-727.

Current logic in ExamAccommodationServiceImpl does a findFirst() based on segment position, exam accommodation type, and examId. However, some exam accommodations can contain multiple values for the same type (for example, Audio Playback Controls). Before "updating" (soft deleting and re-inserting) a changed an accommodation, we need to check that the same accommodation type/value combination doesn't already exist. If it does, we should skip to avoid duplicate entries.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
@@ -327,6 +327,16 @@ class ExamAccommodationServiceImpl implements ExamAccommodationService {
         List<ExamAccommodation> examAccommodationsToDelete = new ArrayList<>();
 
         for (ExamAccommodation examAccommodation : accommodationsToAdd) {
+            // Check if the same accommodation type/value combination exists. If so, we can skip adding/deleting:
+            Optional<ExamAccommodation> identicalExamAccommodation = existingExamAccommodations.stream()
+                .filter(acc -> acc.equals(examAccommodation))
+                .findFirst();
+
+            if (identicalExamAccommodation.isPresent()) {
+                // No need to do anything if the exact same accommodation already exists.
+                continue;
+            }
+
             // Find an existing accommodation for the same type
             //  This might have the same code or not, so can't use contains since it checks for equality
             Optional<ExamAccommodation> existingAccommodation = existingExamAccommodations.stream()
--- a/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
@@ -298,6 +298,7 @@ class ExamAccommodationServiceImpl implements ExamAccommodationService {
                     .withCreatedAt(now)
                     .build();
             })
+            .filter(examAccommodation -> !existingExamAccommodations.contains(examAccommodation))
             .distinct()
             .collect(Collectors.toSet());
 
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -327,6 +327,16 @@ private static String getOtherAccommodationValue(String formattedValue) {
         List<ExamAccommodation> examAccommodationsToDelete = new ArrayList<>();
 
         for (ExamAccommodation examAccommodation : accommodationsToAdd) {
+            // Check if the same accommodation type/value combination exists. If so, we can skip adding/deleting:
+            Optional<ExamAccommodation> identicalExamAccommodation = existingExamAccommodations.stream()
```
millerts: Can we filter this out as part of the stream to accommodations to add after the map (roughly line 300)?

So it would be `map(accommodation -> examAccommodation).filter(examAccommodation -> existingExamAccommodations.contains(examAccommodation))` or something close to that?

burzum619: If we do that, the conditional below will result in the same issue we're seeing now (a duplicate record being inserted)

burzum619:
`````
            // Find an existing accommodation for the same type
            //  This might have the same code or not, so can't use contains since it checks for equality
            Optional<ExamAccommodation> existingAccommodation = existingExamAccommodations.stream()
                .filter(acc ->
                    acc.getExamId().equals(examAccommodation.getExamId())
                        && acc.getSegmentPosition() == examAccommodation.getSegmentPosition()
                        && acc.getType().equals(examAccommodation.getType())
                ).findFirst();

            if (existingAccommodation.isPresent()) {
                // check to see if the ExamAccommodations are logically the same
                //  if they changed then the existing one is removed and replaced with the new code
                if (!isEquivalent(existingAccommodation.get(), examAccommodation)) {
                    examAccommodationsToDelete.add(existingAccommodation.get());
                    examAccommodationsToInsert.add(examAccommodation);
                }
            } else {
                examAccommodationsToInsert.add(examAccommodation);
            }
`````

Comment (jeffjohnson9046) [inline service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java]: I think we _might_ be able to simplify this a little by using `Collections` [removeAll](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeAll-java.util.Collection-), maybe something like this:

```
ArrayList<ExamAccommodation> accommodationsThatAreNotIdentical = accommodationsToAdd.removeAll(existingExamAccommodations);

for (ExamAccommodation examAccommodation : accommodationsThatAreNotIdentical) {
    // do stuff with the examAccommodation....
}
```

... that should eliminate the need to call `filter(predicate).findFirst()` on the `existingExamAccommodation`s in the `for` loop, saving some precious precious CPU cycles.  All of this is untested, but it might be worth a try.

burzum619: Good call! I like this approach better.

jtreuting: This check was originally done in the isEquivalent below.  Which now is not necessary?  NOTE: the isEqual on the object is different checks than isEquivalent.

PR #117, opened 2017-03-23
Insert “default” student “TDS_Other” exam accommodations

burzum619:
Default "Other" accommodations (as configured in ART) were not being added to exam_accommodations table as they did not match up with any default Assessment accommodation.

This is part of TDS-807. However, this bug will not be closed out until TDS-777 is complete. Once TDS-777 is complete, Proctor should have exam accommodation metadata included, and TDS_Other should not be filtered out like it currently is.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
@@ -93,6 +93,13 @@ class ExamAccommodationServiceImpl implements ExamAccommodationService {
             if (assessmentAccommodations.containsKey(code)) {
                 Accommodation accommodation = assessmentAccommodations.get(code);
                 studentAccommodations.put(accommodation.getType(), accommodation);
+            } else if (code.startsWith(OTHER_ACCOMMODATION_VALUE)) {
+                Accommodation otherAccommodation = new Accommodation.Builder()
+                    .withAccommodationType(OTHER_ACCOMMODATION_NAME)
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -93,6 +93,13 @@ public ExamAccommodationServiceImpl(final ExamAccommodationQueryRepository examA
             if (assessmentAccommodations.containsKey(code)) {
                 Accommodation accommodation = assessmentAccommodations.get(code);
                 studentAccommodations.put(accommodation.getType(), accommodation);
+            } else if (code.startsWith(OTHER_ACCOMMODATION_VALUE)) {
+                Accommodation otherAccommodation = new Accommodation.Builder()
```
millerts: Can we break this creation into a private static method that can be used both by this and the line on [311](https://github.com/SmarterApp/TDS_ExamService/blob/develop/service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java#L311)?  I know it is tiny but since the creation of `OTHER` is unique and the same between the two it'd be nice if it was not in two places.

burzum619: They're actually not the same. Here we are constructing a (very basic) assessment Accommodation vs. line 311 where we are constructing an ExamAccommodation. I did at some point attempt to move the "Other" accommodation creation logic later in the flow, but this seemed to be the cleanest approach.

PR #118, opened 2017-03-23
Adding additional properties to ExamAccommodation to get rid of the n…

burzum619:
…eed to fetch Assessment Accommodations on student/proctor

We will no longer need to fetch/match assessment accommodations in Student/Proctor as the exam accommodations will contain all the data we need.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
@@ -48,6 +48,13 @@ public class ExamAccommodationQueryRepositoryImpl implements ExamAccommodationQu
                 "   ea.allow_change, \n" +
                 "   ea.value, \n" +
                 "   ea.segment_position, \n" +
+                "   ea.visible, \n" +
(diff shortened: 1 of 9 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamAccommodationQueryRepositoryImpl.java
```diff
@@ -152,6 +166,9 @@ public ExamAccommodation mapRow(ResultSet rs, int rowNum) throws SQLException {
                 .withSegmentPosition(rs.getInt("segment_position"))
                 .withTotalTypeCount(rs.getInt("total_type_count"))
                 .withCustom(rs.getBoolean("custom"))
+                .withVisible(rs.getBoolean("visible"))
+                .withSortOrder(rs.getInt("sort_order"))
+                .withDependsOn(rs.getString("depends_on"))
```
millerts: There seems to be more columns in the select that what is mapped here?

burzum619: Woah, good catch. Adding additional test coverage (I thought this was being tested already, but apparently not).

PR #125, opened 2017-03-28
Deny exam accommodations when exam status is set to “denied”

burzum619:
Previously, exam accommodations were not set to denied when the entire exam was denied by a proctor (only when individual accommodations were denied). This fixes that.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/ExamAccommodationCommandRepository.java
+++ b/service/src/main/java/tds/exam/repositories/ExamAccommodationCommandRepository.java
@@ -1,6 +1,7 @@
 package tds.exam.repositories;
 
 import java.util.List;
+import java.util.UUID;
 
 import tds.exam.ExamAccommodation;
 
--- a/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
@@ -11,7 +11,6 @@ import org.springframework.stereotype.Repository;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 import java.util.stream.Collectors;
 
 import tds.exam.ExamAccommodation;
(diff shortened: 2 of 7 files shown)


service/src/main/java/tds/exam/repositories/ExamAccommodationCommandRepository.java
```diff
@@ -1,6 +1,7 @@
 package tds.exam.repositories;
 
 import java.util.List;
+import java.util.UUID;
```
millerts: This doesn't look to be used?

service/src/main/java/tds/exam/repositories/impl/ExamAccommodationCommandRepositoryImpl.java
```diff
@@ -11,6 +11,7 @@
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
```
millerts: unused?

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -229,6 +229,20 @@ public ExamAccommodationServiceImpl(final ExamAccommodationQueryRepository examA
         return Optional.empty();
     }
 
+    @Override
+    public void denyAccommodations(final UUID examId) {
+        final List<ExamAccommodation> pendingAccommodations = examAccommodationQueryRepository.findAccommodations(examId);
```
millerts: I think we should do one of the following:

1. Change the argument to an `Exam` instead of the id.
2. Add an Instant argument to this method

The reason being is that I think the `examAccommodation.deniedAt` should be the same as the `exam.statusChangedAt` value when an exam is denied.  So line 239 should use the passed in date or the status changed at date.

I tend to lean towards number 2 myself but went back and forth on it.

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -229,6 +229,20 @@ public ExamAccommodationServiceImpl(final ExamAccommodationQueryRepository examA
         return Optional.empty();
     }
 
+    @Override
```
millerts: Transactional

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
burzum619: Good point, I like the second approach the best.

service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
burzum619: Whats the point of making this transaction if the only write is the update (which is batched)?

PR #130, opened 2017-04-03
Expanding ExpandableExam to include data needed by TRT.

burzum619:
Adding additional data to ExpandableExams. This data will be used by the TDS_ExamResultsTransmitter project for inclusion in the TRT.

https://jira.fairwaytech.com/browse/TDS-755

A few notes:

- Items, Pages, Responses, Scores will be added in a later story for next sprint. I have stubbed out the mapper for the time being and added TODOs. 
- We are using the ExpandableExamMapper interface, which currently only includes a method for fetching a mapping of examIds -> ExpandableExam.Builder objects. You may notice that the mapper implementations included in this PR do a forEach() over the map. However, these mappers will in reality never be used to process more than one record at a time, this is done simply for compatibility with the interface. I figured this approach is sufficient for the time being, as opposed to creating a new interface method or updating each repository find() call to take in a variable amount of examIds. **If we find a need to use these mappers for processing a batch of exams in the future, this code could easily be refactored at that time**. However, I personally think doing that now is a bit of a pre-optimization.


service/src/main/java/tds/exam/services/mappers/impl/ExamSegmentExpandableExamMapper.java
```diff
@@ -0,0 +1,38 @@
...
+        this.examSegmentService = examSegmentService;
+    }
+
+    @Override
+    public void updateExpandableMapper(final Set<ExpandableExamParameters> expandableExamAttributes, final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId) {
+        if (!expandableExamAttributes.contains(ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_SEGMENTS)) {
+            return;
+        }
+
+        examBuilders.forEach((examId, examBuilder) -> {
```
burzum619: Please refer to the second note in my PR description if you have concerns about this forEach().

service/src/test/java/tds/exam/services/impl/mappers/impl/ExpandableExamMappersTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private final Set<ExpandableExamParameters> expandableExamParametersTrt = ImmutableSet.of(
+        ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_SEGMENTS,
+        ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_NOTES,
+        ExpandableExamParameters.EXPANDABLE_PARAMS_EXAMINEE_ATTRIBUTES_AND_RELATIONSHIPS,
+        ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_PAGE_AND_ITEMS,
+        ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_STATUS_DATES
+    );
+
+    @Mock
```
burzum619: Initially this test class only covered a small number of mappers, and it made a decent amount of sense to have them grouped into one class of unit tests as opposed to many tiny unit test classes (especially considering that there is not a whole lot of business logic to test within the mappers anyway). However, I realize this thing is getter a bit large now that we've added more mappers for the TRT. I don't anticipate it will grow beyond what is is below, and I didn't quite yet feel the need to break these out into 8 separate integration test classes, but if someone feels strongly about it I can break these out.

service/src/main/java/tds/exam/services/mappers/impl/ExamStatusExpandableExamMapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Autowired
+    public ExamStatusExpandableExamMapper(final ExamStatusService examStatusService) {
+        this.examStatusService = examStatusService;
+    }
+
+    @Override
+    public void updateExpandableMapper(final Set<ExpandableExamParameters> expandableExamAttributes,
+                                       final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId) {
+        if (!expandableExamAttributes.contains(ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_STATUS_DATES)) {
```
millerts: I'm not advocating that we make this change now, but another pattern we could follow that is a bit clearer (to me) is this:

```
public interface ExpandableMapper {
  public boolean accept(final Set<ExpandableExamParameters> expandableExamAttributes);

  public void updateExpandableMapper(final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId);
}
```

Then the implentation would be:

```
public class ExamAccommodationsExpandableExamMapper implements ExpandableExamMapper {

	@Override
	public boolean accept(final Set<ExpandableExamParameters> expandableExamAttributes) {
        return expandableExamAttributes.contains(ExpandableExamParameters.EXPANDABLE_PARAMS_EXAM_ACCOMMODATIONS))
	}

    @Override
    public void updateExpandableMapper(final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId) {
        //Do Stuff
    }
}
```

Then the caller could do something like this:

```
        expandableExamMappers.stream()
            .filter(mapper -> mapper.accept(expandableExamAttributes))
            .forEach(mapper -> mapper.updateExpandableMapper(expandableExamAttributes, examBuilders, sessionId));
```

It would clearly separate the two actions from accepting the request and acting on the request.  Not that anything is wrong with what we have.  Just another pattern that I've seen previously when doing something like what we're doing here.

service/src/main/java/tds/exam/services/mappers/impl/ExamStatusExpandableExamMapper.java
```diff
@@ -0,0 +1,52 @@
...
+    private final ExamStatusService examStatusService;
+
+    @Autowired
+    public ExamStatusExpandableExamMapper(final ExamStatusService examStatusService) {
+        this.examStatusService = examStatusService;
+    }
+
+    @Override
+    public void updateExpandableMapper(final Set<ExpandableExamParameters> expandableExamAttributes,
+                                       final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId) {
```
millerts: Not recommending any change, but one way we could remove the session id from this interface is to instead do:

```
UUID sessionId = examBuilders.entrySet().build().getSessionId();
```

That particular one expects them to have the same session.  But it seems like this is forced anyhow by the interface as currently written.

service/src/main/java/tds/exam/web/endpoints/ExpandableExamController.java
```diff
@@ -27,9 +30,18 @@ public ExpandableExamController(final ExpandableExamService expandableExamServic
...
-                                                               @RequestParam(required = false) final String... embed) {
+                                                               @RequestParam(required = false) final ExpandableExamParameters... embed) {
         final List<ExpandableExam> exams = expandableExamService.findExamsBySessionId(sessionId, statusNot, embed);
 
         return ResponseEntity.ok(exams);
     }
+
+    @RequestMapping(value = "/{examId}/expandable", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<ExpandableExam> findExpandableExam(@PathVariable final UUID examId,
+                                                      @RequestParam(required = false) final ExpandableExamParameters... embed) {
```
millerts: Whatever naming convention you decide to follow let's make this consistent as well.  Along with the method above. (embed vs. attributes vs. parameters vs. properties)

service/src/test/java/tds/exam/services/impl/ExamStatusServiceImplTest.java
```diff
@@ -0,0 +1,45 @@
...
+        final Instant now = Instant.now();
+        final UUID examId = UUID.randomUUID();
+        when(mockExamStatusQueryRepository.findDateLastTimeStatus(examId, ExamStatusCode.STATUS_STARTED)).thenReturn(Optional.of(now));
+        Optional<Instant> maybeStartDate = examStatusService.findDateLastTimeStatus(examId, ExamStatusCode.STATUS_STARTED);
+
+        verify(mockExamStatusQueryRepository).findDateLastTimeStatus(examId, ExamStatusCode.STATUS_STARTED);
+        assertThat(maybeStartDate.isPresent()).isTrue();
+        assertThat(maybeStartDate.get()).isEqualTo(now);
+    }
+
```
millerts: nit - extra line break

service/src/test/java/tds/exam/services/impl/mappers/impl/ExpandableExamMappersTest.java
millerts: I don't think it is necessary to do now, but I do prefer a bunch of small test classes rather than one uber test.  As you mentioned it quickly grows and becomes harder to maintain.  Especially if any complex logic is added to the mapper in the future which is entirely possible.

We have bigger fish to fry though right now

service/src/main/java/tds/exam/services/impl/ExamStatusServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+public class ExamStatusServiceImpl implements ExamStatusService {
+    private final ExamStatusQueryRepository examStatusQueryRepository;
+
+    @Autowired
+    public ExamStatusServiceImpl(final ExamStatusQueryRepository examStatusQueryRepository) {
+        this.examStatusQueryRepository = examStatusQueryRepository;
+    }
+
+    @Override
+    public Optional<Instant> findDateLastTimeStatus(final UUID examId, final String examStatus) {
```
burzum619: I struggled naming this one too (and didn't like what I ended up with). I like your suggestion better.

service/src/main/java/tds/exam/services/impl/ExamineeServiceImpl.java
```diff
@@ -5,29 +5,46 @@
...
     public ExamineeServiceImpl(final ExamineeCommandRepository examineeCommandRepository,
+                               final ExamineeQueryRepository examineeQueryRepository,
                                final StudentService studentService) {
         this.examineeCommandRepository = examineeCommandRepository;
+        this.examineeQueryRepository = examineeQueryRepository;
         this.studentService = studentService;
     }
 
     @Override
+    public List<ExamineeAttribute> findAllAttributes(final UUID examId, final ExamineeContext examineeContext) {
```
burzum619: You're absolutely right, but I didn't realize this until I started working on the TRT integration. Removing the context param and getting all notes.

service/src/main/java/tds/exam/services/mappers/impl/ExamStatusExpandableExamMapper.java
burzum619: I do really like like this suggestion, but agree its not essential to do this refactoring now. Perhaps in a few months while weeping/rage-programming over a televised Chargers meltdown and

PR #131, opened 2017-04-03
Feature/add scoring service

millerts:
This is a mid stream PR.  We tackled scoring first because it is called first when creating new items.  However, it is also dependent on items being selected.  So it is a catch 22 when it comes to testing.  I'm moving onto item selection but it will be separate from this code


client/src/main/java/tds/exam/ExamItem.java
```diff
@@ -223,4 +228,14 @@ public String getItemFilePath() {
     public Optional<ExamItemResponse> getResponse() {
         return Optional.fromNullable(response);
     }
+
+    @JsonIgnore
+    public boolean isRepsonsePresent() {
```
burzum619: typo

scoring/src/main/java/tds/score/services/ResponseService.java
```diff
@@ -0,0 +1,44 @@
...
+     * @param responseUpdate {@link tds.student.sql.data.IItemResponseUpdate} that contains the response information
+     * @param score          score for the item
+     * @param scoreStatus    the score status
+     * @param scoreRationale the score rationale
+     * @param scoreLatency   the latency
+     * @param pageDuration   the page duration
+     * @return a {@link TDS.Shared.Data.ReturnStatus} that contains whether the update was successful
+     * @throws ReturnStatusException if there is any type of error or validation condition failure
+     */
+    ReturnStatus updateScoredResponse(ExamInstance examInstance, IItemResponseUpdate responseUpdate, int score, String scoreStatus, String scoreRationale, long scoreLatency, long pageDuration) throws ReturnStatusException;
```
burzum619: final args

PR #133, opened 2017-04-05
TDS-832: Make sure Exam-level notes are not HTML escaped

jeffjohnson9046:
[TDS-832](https://jira.fairwaytech.com/browse/TDS-832): Only the item-level notes should be HTML-escaped when stored in the database.  The exam-level notes should not be HTML-escaped.


service/src/main/java/tds/exam/web/endpoints/ExamineeNoteController.java
```diff
@@ -46,9 +47,16 @@ public ExamineeNoteController(final ExamineeNoteService examineeNoteService) {
     @ResponseStatus(HttpStatus.NO_CONTENT)
     void insert(@PathVariable final UUID id,
                 @RequestBody final ExamineeNote note) {
+        // In legacy the student application, only item-level notes are HTML-escaped and exam-level notes are not
+        // (TestShellHandler#RecordItemComment, line 413 and TestShellHandler#RecordOppComment, line 436).  Since the
+        // note text is already escaped by the time it gets here, it must be un-escaped for Exam-level notes.
+        final String noteText = note.getContext().equals(ExamineeNoteContext.EXAM)
```
millerts: Can we modify the code in student to un-escape the notes and store our notes in a consistent fashion?  I'm thinking this code can live in the Student controller or service instead of in our microservice.  

I understand you were just carrying forward the existing logic.  It just seems wrong to me.

jeffjohnson9046: I agree this is probably a bug in their legacy code - why would one type of note be escaped but not the other?  Also, I see this in the legacy Student application that used to escape the exam-level note but was removed (`TestShellHandler`, line 429):

```
    // Following code commented in 2013 Dotnet version
    /*
     * //encode and trim comment if (comment != null) { // comment =
     * AntiXss.HtmlEncode(comment, 2000); comment =
     * UrlEncoderDecoderUtils.encode (comment); }
     */
```

I'll try to fix it in the legacy Student application so that notes are escaped consistently.  My concern is the display of an escaped exam-level note; they might have turned that off/commented it out for some reason.

PR #134, opened 2017-04-05
Remove unwanted appender.

trapnine:
This stanza was creating unwanted files. Removing for now (probably never needed, in fact).


trapnine:
This stanza was creating unwanted files. Removing for now (probably never needed, in fact).

PR #140, opened 2017-04-07
Refactoring approve accommodations. Adding check/exam update for cust…

burzum619:
…om accommodations

This PR is in regards to TDS-836. This bug is caused because the "custom_accommodations" flag is not being set on the Exam after exam_accommodation changes are approved (neither by the guest student changing them from default, or a proctor approving non-default accommodations). 

Unfortunately, it wasn't as simple as simply adding ExamService (which contains a method already to update the exam flag if a custom accommodation exists) dependency to ExamAccommodation. This would have caused a circular dependency. ExamServiceImpl is already fairly large and I was hoping to not have to add to it, but I think this is the cleanest solution as ExamServiceImpl is already dependent on many of the services that are utilized in the approval flow.  The benefit of this refactor is that ExamAccommodationService now has three less dependencies (including ExamQueryRepository).

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -310,7 +310,7 @@ class ExamServiceImpl implements ExamService {
         Exam exam = maybeExam.get();
 
         if (!request.isGuest()) {
-            /* line 11441 */
+            /* StudentDLL line 11441 */
             Optional<ValidationError> maybeError = examApprovalService.verifyAccess(new ExamInfo(examId, request.getSessionId(), request.getBrowserId()), exam);
 
             if (maybeError.isPresent()) {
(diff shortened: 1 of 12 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -296,6 +296,44 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        if (!request.isGuest()) {
+            /* line 11441 */
+            Optional<ValidationError> maybeError = examApprovalService.verifyAccess(new ExamInfo(examId, request.getSessionId(), request.getBrowserId()), exam);
+
+            if (maybeError.isPresent()) {
+                return maybeError;
+            }
+        }
+
+        /* lines 11465-11473 */
```
millerts: Can you reference the class here.   I assume it is StudentDLL but let's make it clear since I hope to not change this for a while :)

PR #146, opened 2017-04-14
Submit completed exams to a messaging channel for further processing/…

wtritch:
…reporting.

This PR includes a logging submission acknowledgement handler.  If an exam cannot be submitted, the failure will be logged using the standard SLF4J logging framework.
@mtrupkin @trapnine How do we get the logs into the long-term logging framework?  Because if an exam cannot be submitted, it requires action to re-submit it.

### DIFF
--- a/client/src/main/java/tds/exam/ExamTopics.java
+++ b/client/src/main/java/tds/exam/ExamTopics.java
@@ -1,0 +1,19 @@
+package tds.exam;
+
+/**
+ * This class holds AMQP topics used by the ExamService.
+ * Topic names are in the general for of "{application}.{scope}.{operation}"
+ */
+public class ExamTopics {
+
+    /**
+     * This is the topic exchange for all exam topics.
+     */
+    public static final String TOPIC_EXCHANGE = "tds.exam.exchange";
+
+    /**
+     * This topic is published to when an exam is completed.
+     * The message body is a String containing the exam id.
+     */
+    public static final String TOPIC_EXAM_COMPLETED = "exam.completed";
+}
--- a/service/src/main/java/tds/exam/configuration/messaging/ExamMessagingConfiguration.java
+++ b/service/src/main/java/tds/exam/configuration/messaging/ExamMessagingConfiguration.java
@@ -1,0 +1,35 @@
+package tds.exam.configuration.messaging;
+
+import org.springframework.amqp.core.TopicExchange;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import tds.exam.utils.LoggingAmqpConfirmationCallback;
+
+import javax.annotation.PostConstruct;
+
+import static tds.exam.ExamTopics.TOPIC_EXCHANGE;
+
+/**
+ * This configuration is responsible for initializing AMQP (RabbitMQ)
+ */
+@Configuration
+public class ExamMessagingConfiguration {
+
+    @Autowired
+    private RabbitTemplate rabbitTemplate;
+
+    @Bean
+    public TopicExchange exchange() {
+        return new TopicExchange(TOPIC_EXCHANGE, true, false);
+    }
+
+    /**
+     * Add a logging confirmation callback that will log errors if an exam could not be submitted.
+     */
+    @PostConstruct
+    public void rabbitTemplate() {
+        rabbitTemplate.setConfirmCallback(new LoggingAmqpConfirmationCallback());
+    }
+}
--- a/service/src/main/java/tds/exam/services/MessagingService.java
+++ b/service/src/main/java/tds/exam/services/MessagingService.java
@@ -1,0 +1,14 @@
+package tds.exam.services;
+
+/**
+ * Implementations of this interface are responsible for publishing messages to the ecosystem messaging framework.
+ */
+public interface MessagingService {
+
+    /**
+     * Send an exam completion message to the transmission topic.
+     *
+     * @param examId The completed exam's id
+     */
+    void sendExamCompletion(final String examId);
+}
--- a/service/src/main/java/tds/exam/services/MessagingService.java
+++ b/service/src/main/java/tds/exam/services/MessagingService.java
@@ -1,5 +1,7 @@
 package tds.exam.services;
 
+import java.util.UUID;
+
 /**
  * Implementations of this interface are responsible for publishing messages to the ecosystem messaging framework.
  */
@@ -10,5 +12,5 @@ public interface MessagingService {
      *
      * @param examId The completed exam's id
      */
-    void sendExamCompletion(final String examId);
+    void sendExamCompletion(final UUID examId);
 }
--- a/service/src/main/java/tds/exam/services/impl/MessagingServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/MessagingServiceImpl.java
@@ -1,0 +1,30 @@
+package tds.exam.services.impl;
+
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.amqp.rabbit.support.CorrelationData;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import tds.exam.services.MessagingService;
+
+import static tds.exam.ExamTopics.TOPIC_EXAM_COMPLETED;
+import static tds.exam.ExamTopics.TOPIC_EXCHANGE;
+
+/**
+ * Default implementation of a MessagingService using a RabbitTemplate.
+ */
+@Service
+public class MessagingServiceImpl implements MessagingService {
+
+    private final RabbitTemplate rabbitTemplate;
+
+    @Autowired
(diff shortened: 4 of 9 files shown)


client/src/main/java/tds/exam/ExamTopics.java
```diff
@@ -0,0 +1,19 @@
...
+     * This is the topic exchange for all exam topics.
+     */
+    public static final String TOPIC_EXCHANGE = "tds.exam.exchange";
+
+    /**
+     * This topic is published to when an exam is completed.
+     * The message body is a String containing the exam id.
+     */
+    public static final String TOPIC_EXAM_COMPLETED = "exam.completed";
+}
```
wtritch: After discussion with a few people, I decided that the service should advertise the exchanges/topics that it's publishing to in the "client" module, that way consumers both see the channels with available messages as well as have a way to deserialize more complex payloads.

service/src/main/java/tds/exam/configuration/messaging/ExamMessagingConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+        return new TopicExchange(TOPIC_EXCHANGE, true, false);
+    }
+
+    /**
+     * Add a logging confirmation callback that will log errors if an exam could not be submitted.
+     */
+    @PostConstruct
+    public void rabbitTemplate() {
+        rabbitTemplate.setConfirmCallback(new LoggingAmqpConfirmationCallback());
+    }
```
wtritch: I'm not a huge fan of @PostConstruct, but the alternative is to copy the RabbitTemplate bean definition logic from RabbitAutoConfiguration to roll our own...  Thoughts?

service/src/main/java/tds/exam/services/MessagingService.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+public interface MessagingService {
+
+    /**
+     * Send an exam completion message to the transmission topic.
+     *
+     * @param examId The completed exam's id
+     */
+    void sendExamCompletion(final String examId);
+}
```
wtritch: I used this service to abstract out the messaging implementation, in case we ever switch to another messaging framework.

service/src/main/java/tds/exam/services/impl/MessagingServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    @Autowired
+    public MessagingServiceImpl(final RabbitTemplate rabbitTemplate) {
+        this.rabbitTemplate = rabbitTemplate;
+    }
+
+    @Override
+    public void sendExamCompletion(final String examId) {
+        final CorrelationData correlationData = new CorrelationData("exam.completion-" + examId);
+        this.rabbitTemplate.convertAndSend(TOPIC_EXCHANGE, TOPIC_EXAM_COMPLETED, examId, correlationData);
+    }
```
wtritch: I have to use a RabbitTemplate rather than a more generic AmqpTemplate in order to support submission acknowledgement using CorrelationData.  (The RabbitTemplate calls the LoggingAmqpConfirmationCallback with the CorrelationData when a submission is either ack'd or nack'd)

service/src/main/java/tds/exam/utils/LoggingAmqpConfirmationCallback.java
```diff
@@ -0,0 +1,25 @@
...
+    private final static Logger LOG = LoggerFactory.getLogger(LoggingAmqpConfirmationCallback.class);
+
+    @Override
+    public void confirm(final CorrelationData correlationData,
+                        final boolean ack,
+                        final String cause) {
+        if (ack) {
+            LOG.debug("Message acknowledged as sent: {}", correlationData);
+        } else {
+            LOG.error("Message not acknowledged as sent: {} cause: {}", correlationData, cause);
```
wtritch: If this happens we have a problem that requires action.  The exam will not be reported.  (It should, of course, never happen. =)

service/src/main/resources/application.yml
```diff
@@ -20,6 +20,10 @@ spring:
 #Port defined in TDS_Build:docker-compose.yml
   rabbitmq:
     port: 32846
+    publisher-confirms: true
+    template:
+      retry:
+        enabled: true
```
wtritch: I turned on retries with an exponential fallback.  By default the RabbitTemplate will retry 3 times.

service/src/main/java/tds/exam/services/MessagingService.java
millerts: Can we change this to take in a UUID?  The implementation can convert it to String

service/src/main/resources/application.yml
wtritch: Note that this also means that the ExamService WILL NOT stand up without having a Rabbit service available, since it now provides required functionality.  I can try digging into a way to provide a mock RabbitTemplate implementation if the configured rabbitmq server cannot be reached, but we'd need to be very careful to make sure we never end up using a mock impl in production...  I'm tempted to leave it be and require that developers run the TDS_Build's docker-compose environment before running the exam service locally...

service/src/main/java/tds/exam/configuration/messaging/ExamMessagingConfiguration.java
millerts: I guess it depends how much we think we're going to configure the rabbitTemplate.  For this one off, I think post construct is fine.

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
```diff
@@ -75,15 +77,17 @@ public void accept(final Exam oldExam, final Exam newExam) {
         // tables/queries that are referenced in that method are only used by the loader stored procedures, schema
         // creation/modification scripts and/or SimDLL.java (which is related to the simulator).
 
+        // Publish the submitted exam to the Messaging backend, allowing other services to continue processing it.
+        // Submit for scoring (CommonDLL#_OnStatus_Completed_SP, line 1433 - 1434), which changes the exam's status to "submitted"
+        messagingService.sendExamCompletion(newExam.getId().toString());
```
millerts: I think this should be the last thing done in this flow.

service/src/test/java/tds/exam/services/impl/MessagingServiceImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.verify;
+import static tds.exam.ExamTopics.TOPIC_EXAM_COMPLETED;
+import static tds.exam.ExamTopics.TOPIC_EXCHANGE;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MessagingServiceImplTest {
+
+    @Mock
+    private RabbitTemplate rabbitTemplate;
```
millerts: nit, can we name this `mockRabbitTemplate`?  We've been following that pattern because it is easier to see what is a mock implementation in some of our larger tests.

service/src/test/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListenerTest.java
```diff
@@ -45,6 +47,9 @@
     @Mock
     private ExamineeService mockExamineeService;
 
+    @Mock
+    private MessagingService messagingService;
```
millerts: Same comment about prefix `mock`

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
wtritch: I started that way, except there's a fast-fail a little further down, and this more closely mirrors the flow from the original implementation.  I can make it work if you feel strongly, however?

service/src/test/java/tds/exam/services/impl/MessagingServiceImplTest.java
wtritch: done

service/src/test/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListenerTest.java
wtritch: done

service/src/main/resources/application.yml
millerts: I think that is fine.  We have to do that now anyway to not see the startup exceptions.  And I think normal local development requires standing up docker.

service/src/main/java/tds/exam/services/MessagingService.java
wtritch: done

service/src/main/java/tds/exam/utils/listeners/OnCompletedStatusExamChangeListener.java
millerts: I'd tweak the conditional below to update if there are field test items.  If not just ignore.  Then do this code.

service/src/main/java/tds/exam/utils/LoggingAmqpConfirmationCallback.java
wtritch: After discussing with @mtrupkin standard SLF4J logs should find their way to a persistent log store.

PR #148, opened 2017-04-15
Added the ability to get items in the expandable mappers.  Also clean…

millerts:
…ed up some of the code in the other mappers.


service/src/main/java/tds/exam/mappers/impl/ExamPageItemResponseExpandableExamMapper.java
```diff
@@ -32,15 +32,8 @@ public void updateExpandableMapper(final Set<ExpandableExamAttributes> expandabl
         if (expandableAttributes.contains(ExpandableExamAttributes.EXAM_PAGE_AND_ITEMS)) {
             examBuilders.forEach((examId, examBuilder) -> {
                 List<ExamPage> examPages = examPageService.findAllPages(examId);
-                // TODO: Populate items
-                ExpandableExam.Builder builder = examBuilders.get(examId);
-                builder.withExamPages(examPages);
-            });
-        }
-
-        if (expandableAttributes.contains(ExpandableExamAttributes.EXAM_PAGE_ITEMS_AND_RESPONSES)) {
```
millerts: This was a mistake on my part when i created the initial JIRA.  I don't see a valid reason to ever get the items and not the responses.

service/src/main/java/tds/exam/mappers/impl/ExamSegmentExpandableExamMapper.java
```diff
@@ -29,9 +29,8 @@ public void updateExpandableMapper(final Set<ExpandableExamAttributes> expandabl
             return;
         }
 
-        examBuilders.forEach((examId, examBuilder) -> {
+        examBuilders.forEach((examId, builder) -> {
             List<ExamSegment> examSegments = examSegmentService.findExamSegments(examId);
-            ExpandableExam.Builder builder = examBuilders.get(examId);
```
millerts: This isn't needed because we're iterating over the Map.Entry which has both the key and the builder.  Removed this pattern from the few classes that had it.

client/src/main/java/tds/exam/ExpandableExam.java
```diff
@@ -16,7 +16,6 @@
     private List<ExamSegment> examSegments;
     private List<ExamPage> examPages;
     private List<ExamItem> examItems;
-    private List<ExamItemResponse> examItemResponses;
```
millerts: The ExamItem class has an ExamItemResponse.  Would rather associate the responses with the item rather than contain two collections.  Usually you'd want the item and response not either/or

PR #149, opened 2017-04-18
Feature/cleanup item rendering

millerts:
This leverages the changes in https://github.com/SmarterApp/TDS_ItemRenderer/pull/28

### DIFF
--- a/scoring/pom.xml
+++ b/scoring/pom.xml
@@ -18,7 +18,7 @@
         <java.version>1.8</java.version>
         <assertj.version>3.4.1</assertj.version>
         <item-scoring.version>3.0.1</item-scoring.version>
-        <item-renderer.version>3.0.4</item-renderer.version>
+        <item-renderer.version>3.0.8-SNAPSHOT</item-renderer.version>
         <tds-student-common.version>3.0.0</tds-student-common.version>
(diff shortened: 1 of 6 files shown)


scoring/pom.xml
```diff
@@ -18,7 +18,7 @@
         <java.version>1.8</java.version>
         <assertj.version>3.4.1</assertj.version>
         <item-scoring.version>3.0.1</item-scoring.version>
-        <item-renderer.version>3.0.7</item-renderer.version>
+        <item-renderer.version>3.0.8-SNAPSHOT</item-renderer.version>
```
millerts: I will not leave this as snapshot.  There is an open PR in item renderer, and I'll make this correct after I release it.

service/src/main/resources/application.yml
```diff
@@ -75,6 +75,7 @@ exam:
       pythonScoringUrl: http://localhost:32845
       maxAttempts: 3
       sympyTimeoutMillis: 1000
+      encryptionEnabled: false
```
wtritch: Should we note that encryption cannot be enabled?

millerts: For now, yes.

PR #151, opened 2017-04-18
Expose an item scoring endpoint.

wtritch:
This PR exposes an item scoring endpoint at:
/exam/{examId}/scores/responses
that will score and update student responses.

### DIFF
--- a/scoring/pom.xml
+++ b/scoring/pom.xml
@@ -17,9 +17,10 @@
     <properties>
         <java.version>1.8</java.version>
         <assertj.version>3.4.1</assertj.version>
+        <ss-multijar.version>3.0.3</ss-multijar.version>
         <item-scoring.version>3.0.1</item-scoring.version>
         <item-renderer.version>3.0.8</item-renderer.version>
-        <tds-student-common.version>3.0.0</tds-student-common.version>
+        <tds-student-common.version>3.0.4-SNAPSHOT</tds-student-common.version>
     </properties>
 
     <dependencies>
@@ -77,6 +78,11 @@
                     <artifactId>el-api</artifactId>
                 </exclusion>
             </exclusions>            
+        </dependency>
+        <dependency>
+            <groupId>org.opentestsystem.shared</groupId>
+            <artifactId>shared-db</artifactId>
+            <version>${ss-multijar.version}</version>
         </dependency>
         <dependency>
             <groupId>org.opentestsystem.delivery</groupId>
--- a/scoring/pom.xml
+++ b/scoring/pom.xml
@@ -20,7 +20,7 @@
         <ss-multijar.version>3.0.3</ss-multijar.version>
         <item-scoring.version>3.0.1</item-scoring.version>
         <item-renderer.version>3.0.8</item-renderer.version>
-        <tds-student-common.version>3.0.4-SNAPSHOT</tds-student-common.version>
+        <tds-student-common.version>3.0.4</tds-student-common.version>
     </properties>
 
     <dependencies>
--- a/service/src/main/java/tds/exam/item/S3ItemReader.java
+++ b/service/src/main/java/tds/exam/item/S3ItemReader.java
@@ -4,7 +4,6 @@ import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.model.AmazonS3Exception;
 import com.amazonaws.services.s3.model.GetObjectRequest;
 import com.amazonaws.services.s3.model.S3Object;
-import net.logstash.logback.encoder.org.apache.commons.lang.ArrayUtils;
 import org.springframework.stereotype.Component;
 import tds.exam.configuration.scoring.ScoringS3Properties;
 import tds.itemrenderer.processing.ItemDataReader;
@@ -13,14 +12,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 
-import static org.apache.commons.lang3.StringUtils.join;
+import static org.apache.commons.io.FilenameUtils.getName;
+import static org.apache.commons.io.FilenameUtils.removeExtension;
+import static org.apache.commons.lang3.StringUtils.capitalize;
 
 @Component
 public class S3ItemReader implements ItemDataReader {
     private final AmazonS3 s3Client;
     private final ScoringS3Properties s3Properties;
 
-    public S3ItemReader(final AmazonS3 s3Client,
+    S3ItemReader(final AmazonS3 s3Client,
                         final ScoringS3Properties s3Properties) {
         this.s3Client = s3Client;
         this.s3Properties = s3Properties;
@@ -28,7 +29,7 @@ public class S3ItemReader implements ItemDataReader {
 
     @Override
     public InputStream readData(final URI uri) throws IOException {
-        final String itemLocation = s3Properties.getItemPrefix() + trimPath(uri);
+        final String itemLocation = s3Properties.getItemPrefix() + buildPath(uri);
 
         try {
             final S3Object item = s3Client.getObject(new GetObjectRequest(
@@ -50,18 +51,12 @@ public class S3ItemReader implements ItemDataReader {
      * to
      * items/Item-187-2501/item-187-2501.xml
      *
-     * TODO This should be updated and/or removed once we have full control over the paths coming in.
-     *
      * @param uri   A resource uri
      * @return The resource path relative to our S3 bucket and prefix
      */
-    private String trimPath(final URI uri) {
-        String[] pathElements = uri.getPath().split("/");
-        if (pathElements.length < 3) {
-            throw new IllegalArgumentException("Resource path must have at least 3 elements: " + uri.toString());
-        }
-        return join(
-            ArrayUtils.subarray(pathElements, pathElements.length-3, pathElements.length),
-            "/");
+    private String buildPath(final URI uri) {
+        final String itemName = getName(uri.toString());
+        final String dirName = capitalize(removeExtension(itemName));
+        return "items/" + dirName + "/" + itemName;
     }
 }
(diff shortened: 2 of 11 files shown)


scoring/pom.xml
```diff
@@ -17,9 +17,10 @@
     <properties>
         <java.version>1.8</java.version>
         <assertj.version>3.4.1</assertj.version>
+        <ss-multijar.version>3.0.3</ss-multijar.version>
         <item-scoring.version>3.0.1</item-scoring.version>
         <item-renderer.version>3.0.8</item-renderer.version>
-        <tds-student-common.version>3.0.0</tds-student-common.version>
+        <tds-student-common.version>3.0.4-SNAPSHOT</tds-student-common.version>
```
wtritch: I won't check this in using a snapshot,  this is dependent upon some minor student changes.

scoring/pom.xml
```diff
@@ -79,6 +80,11 @@
             </exclusions>            
         </dependency>
         <dependency>
+            <groupId>org.opentestsystem.shared</groupId>
+            <artifactId>shared-db</artifactId>
+            <version>${ss-multijar.version}</version>
+        </dependency>
```
wtritch: This was stripped from tds-student-common between version 3.0.0 and the latest.  We need it for the ReturnStatusException class.

service/src/main/java/tds/exam/item/S3ItemReader.java
```diff
@@ -25,14 +28,40 @@ public S3ItemReader(final AmazonS3 s3Client,
...
+     * @return The resource path relative to our S3 bucket and prefix
+     */
+    private String trimPath(final URI uri) {
+        String[] pathElements = uri.getPath().split("/");
+        if (pathElements.length < 3) {
+            throw new IllegalArgumentException("Resource path must have at least 3 elements: " + uri.toString());
+        }
+        return join(
+            ArrayUtils.subarray(pathElements, pathElements.length-3, pathElements.length),
+            "/");
```
wtritch: I don't care for this, or for demanding that all S3 items be exactly 3 paths below the prefix.  I suppose I could use regex to find "items" and "stimuli" but that feels just as fragile.  Thoughts?

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -117,6 +116,7 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
         }
 
         if (StringUtils.isNotEmpty(errorMessage)) {
+            LOG.warn("Problem accessing exam item: {}", errorMessage);
```
wtritch: The thrown ReturnStatusException doesn't have the fancy error message defined above, which means we don't know why the request failed without a logging statement.

service/src/main/java/tds/exam/services/scoring/ScoreConfigServiceImpl.java
```diff
@@ -36,7 +35,7 @@ public ScoreConfigServiceImpl(final RestTemplate restTemplate, final ExamService
         UriComponents uriComponents =
             UriComponentsBuilder
                 .fromHttpUrl(examServiceProperties.getConfigUrl())
-                .path("{clientName}/scoring")
+                .path("config/{clientName}/scoring")
```
wtritch: The path was off.

service/src/main/java/tds/exam/web/advice/ExamExceptionAdvice.java
```diff
@@ -16,12 +18,30 @@
...
+        HttpStatus status = HttpStatus.valueOf(ex.getReturnStatus().getHttpStatusCode());
+        // By default a ReturnStatusException has a response code of 200 (OK).
+        // Since we should never return an error as a 200, convert 200 responses
+        // to a default code of 500 (Internal Server Error).
+        status = status == HttpStatus.OK ? HttpStatus.INTERNAL_SERVER_ERROR : status;
+
+        return new ResponseEntity<>(
+            new ExceptionMessageResource(ex.getReturnStatus().getStatus(), ex.getReturnStatus().getReason()),
+            status);
+    }
```
wtritch: Added a handler for ReturnStatusExceptions that are thrown by the system (I think only the scoring system throws them currently).  It's annoying that the default exception HttpStatus code is 200 OK...

service/src/main/java/tds/exam/web/endpoints/ExamScoringController.java
```diff
@@ -0,0 +1,49 @@
...
+    @VerifyAccess
+    public ResponseEntity<List<ItemResponseUpdateStatus>> updateResponses(@PathVariable final UUID examId,
+                                                                          @RequestParam final UUID sessionId,
+                                                                          @RequestParam final UUID browserId,
+                                                                          @RequestParam final String clientName,
+                                                                          @RequestParam final Float pageDuration,
+                                                                          @RequestBody final List<ItemResponseUpdate> responseUpdates) throws ReturnStatusException {
+        final ExamInstance examInstance = ExamInstance.create(examId, sessionId, browserId, clientName);
+        final List<ItemResponseUpdateStatus> responses = itemScoringService.updateResponses(examInstance, responseUpdates, pageDuration);
+        return new ResponseEntity<>(responses, HttpStatus.OK);
```
wtritch: Unfortunately, there are a *lot* of params in this request.

service/src/main/resources/application.yml
```diff
@@ -80,6 +80,6 @@ exam:
       encryptionEnabled: false
     s3:
       bucketName: tds-resources
-      prefix: item-content/local/
       accessKey: ${s3.key}
-      secretKey: ${s3.secret}
\ No newline at end of file
+      secretKey: ${s3.secret}
+      itemPrefix: item-content/local/
```
wtritch: This was a small typo.  The configuration properties class has an "itemPrefix" property, rather than a "prefix" property.  I'll update the configuration repository as well.

service/src/main/java/tds/exam/item/S3ItemReader.java
millerts: Consider the path `/usr/local/tomcat/resources/tds/bank/items/Item-187-2501/item-187-2501.xml`.  I think we should do the following:

1. Use FilePathUtils to get the file only.  This would leave us with `item-187-2501.xml`. Call this string `itemFileName`
2. Get the file name without the extension and capitalize the first letter. Call this `itemDirectory` and the value would be `Item-187-2501`
3. Our prefix should be whatever we want on s3 down to that level.  So for our example let's say `prefix="tds-resouces/local/items"`
4. Then pull it all together `return prefix + "/" + itemDirectory + "/" + itemFile`

The reason I think this will always work is because the specification for items themselves insist that the item package is name `Item-###/item-###.xml`.  So we're guaranteed that the last part of the path will always be that.  If not, the system won't ever work.

service/src/main/java/tds/exam/web/advice/ExamExceptionAdvice.java
millerts: Yep.  The system currently never throws 500.  Another interesting decision made by previous maintainers

service/src/main/java/tds/exam/web/endpoints/ExamScoringController.java
millerts: Why parameters instead of making it part of the body?  Basically shove all the parameters into a more complex object.

```
{
  "sessionId":"id",
  "responseUpdates": [ {responseUpdateJson}]
}
```
If you don't think that should be in the body then sure, but if we're doing this just to make it easy on the Student side I think we should tweak the template code that calls this to build up a more complex object.

service/src/main/resources/application.yml
millerts: Thanks.  This was merged as a WIP because it was blocking some other people.  :/ - lesson learned

service/src/test/java/tds/exam/web/endpoints/ExamScoringControllerIntegrationTests.java
```diff
@@ -0,0 +1,87 @@
...
+            .thenReturn(responses);
+
+        final MvcResult result = http.perform(MockMvcRequestBuilders.put(new URI("/exam/" + examId.toString() + "/scores/responses"))
+            .param("sessionId", sessionId.toString())
+            .param("browserId", browserId.toString())
+            .param("clientName", clientName)
+            .param("pageDuration", pageDuration.toString())
+            .contentType(APPLICATION_JSON)
+            .content(objectMapper.writeValueAsString(updates)))
+//            .andDo(print())
```
millerts: remove?

service/src/main/java/tds/exam/item/S3ItemReader.java
wtritch: Good point.  I had forgotten this class is specifically for reading "Items" not general S3 resources.  So we don't need this class to read "stimuli" ?  Will we have an associated S3StimuliReader in the future?  And for item resources like {{items/Item-187-2501/item_2501_enu_contracted.brf}} will we have a separate resource reader?
I guess the question is if this *should* be an item-xml-only reader or if we want it to be able to read any Item S3 resource, or any S3 resource?

service/src/main/java/tds/exam/web/endpoints/ExamScoringController.java
wtritch: ...I'm not sure that I *want* to create another wrapper model object...  I'm tempted to leave this as-is for now.

service/src/test/java/tds/exam/web/endpoints/ExamScoringControllerIntegrationTests.java
wtritch: Ah, thanks for calling this out, removed.  I left this in to point it out as a way to see raw/serialized response values and request properties in Controller Integration tests in case people haven't seen it before.  Took me a while to find it the first time.

service/src/main/java/tds/exam/item/S3ItemReader.java
wtritch: After discussing offline, this class will ONLY ever read item xml.  Other resources will require other readers.  That allows this method to be cleaned up to generate the path based solely upon the item filename.

PR #154, opened 2017-04-20
TDS-853: Feature/get opp items

jeffjohnson9046:
[TDS-853](https://jira.fairwaytech.com/browse/TDS-853): Service endpoint for getting data necessary to build a `PageList` for the legacy student application.

**NOTE:** There will be a companion PR against **TDS_Student** to call this endpoint.


service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -45,6 +56,27 @@ public void deletePages(final UUID examId) {
     }
 
     @Override
+    public List<ExamPage> findAllPagesWithItems(final UUID examId) {
+        final List<ExamPage> examPages = examPageQueryRepository.findAll(examId);
+        final List<ExamItem> examItems = examItemQueryRepository.findExamItemAndResponses(examId);
+
+        // Associate each exam item to its parent page
+        return examPages.stream()
```
millerts: I would flip this.  I would iterate over the items and map them to a Map<UUID, List<ExamItems>>.  I would then iterate over the Pages fetching from the map.

The reason I would flip is because items will always be the larger collection so we should only iterate over it once instead iterating over that list for each page.

millerts:
```
        Map<UUID, List<ExamItem>> mappedItesm = items.stream()
            .collect(Collectors.groupingBy(ExamItem::getExamPageId));

        List<ExamPage> pageWithItems = pages.stream()
            .map(examPage -> {
                    List<ExamItem> pageItems = mappedItesm.containsKey(examPage.getId())
                        ? mappedItesm.get(examPage.getId())
                        : new ArrayList<>();
                    return ExamPage.Builder
                        .fromExamPage(examPage)
                        .withExamItems(pageItems)
                        .build();
                }
            ).collect(Collectors.toList());
```

PR #157, opened 2017-04-21
Bump TDS_Common dependency version, disable caching in local-developm…

wtritch:
…ent.

UPDATE:  This is more of a PR template.  I'll be doing the same to all of our microservices once this PR is approved.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>org.opentestsystem.delivery</groupId>
             <artifactId>tds-common-legacy</artifactId>
-            <version>${tds-common-legacy.version}</version>
+            <version>${tds-common.version}</version>
         </dependency>
         <!-- Validation -->
         <dependency>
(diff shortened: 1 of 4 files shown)


client/pom.xml
```diff
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>org.opentestsystem.delivery</groupId>
             <artifactId>tds-common-legacy</artifactId>
-            <version>${tds-common-legacy.version}</version>
+            <version>${tds-common.version}</version>
```
wtritch: There's no reason to specify tds-common-legacy and tds-common versions independently.  They're always released in lock-step.

service/src/main/resources/application.yml
```diff
@@ -28,6 +28,10 @@ spring:
       retry:
         enabled: true
 
+tds:
+  caching:
+    enabled: false
```
wtritch: Disabled caching in local-development

service/src/test/resources/application.yml
```diff
@@ -13,6 +13,10 @@ spring:
     bus:
       enabled: false
 
+tds:
+  caching:
+    enabled: false
```
wtritch: disabled caching during tests

PR #158, opened 2017-04-24
Bugfix/include all page fields

millerts:
Contains the following:

- Fetches all pertinent data for an `exam_page` that is contained in the exam_page table and `exam_page_event`
- Renamed `exam_segment_key` to `segment_key` because there is no `examSegmentKey`.  It is a composite key of `exam_id` and `segment_key`. 
- Removed the Thread.sleep()
- made non null fields in the DB non null when creating the object

### DIFF
--- a/client/src/main/java/tds/exam/ExamPage.java
+++ b/client/src/main/java/tds/exam/ExamPage.java
@@ -6,6 +6,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
+import static tds.common.util.Preconditions.checkNotNull;
+
 /**
  * Represents the page of an exam
  */
@@ -13,7 +15,7 @@ import java.util.UUID;
 public class ExamPage {
     private UUID id;
     private int pagePosition;
-    private String segmentKey;
+    private String examSegmentKey;
     private String segmentId;
     private int segmentPosition;
     private String itemGroupKey;
@@ -32,16 +34,16 @@ public class ExamPage {
     }
 
     private ExamPage(Builder builder) {
-        id = builder.id;
+        id = checkNotNull(builder.id);
         pagePosition = builder.pagePosition;
-        segmentKey = builder.segmentKey;
+        examSegmentKey = checkNotNull(builder.segmentKey);
         segmentId = builder.segmentId;
         segmentPosition = builder.segmentPosition;
-        itemGroupKey = builder.itemGroupKey;
+        itemGroupKey = checkNotNull(builder.itemGroupKey);
         groupItemsRequired = builder.groupItemsRequired;
-        examId = builder.examId;
+        examId = checkNotNull(builder.examId);
         examItems = builder.examItems == null ? new ArrayList<ExamItem>() : builder.examItems;
-        createdAt = builder.createdAt;
+        createdAt = checkNotNull(builder.createdAt);
         deletedAt = builder.deletedAt;
         startedAt = builder.startedAt;
     }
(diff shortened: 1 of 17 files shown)


client/src/main/java/tds/exam/ExamPage.java
```diff
@@ -169,14 +171,14 @@ public String getSegmentKey() {
     }
 
     /**
-     * @return The id of which segment owns this page
+     * @return The id of which segment owns this page. This is only populated if the query joins ExamSegment
```
millerts: I don't love that we have these properties that are only populated for certain calls, but I didn't want to go down the rabbit hole of changing it.

service/src/main/java/tds/exam/repositories/ExamAccommodationCommandRepository.java
```diff
@@ -1,7 +1,6 @@
 package tds.exam.repositories;
 
 import java.util.List;
-import java.util.UUID;
```
millerts: Unused import.  Couldn't help myself.

service/src/main/resources/db/migration/V1493073035__exam_replace_testeeresponse_view.sql
```diff
@@ -0,0 +1,87 @@
...
+    'not migrated' AS responsedurationinsecs
+  FROM
+    exam_page AS page
+    JOIN
+    qa_exam_most_recent_event_per_exam_page AS last_page_event
+      ON page.id = last_page_event.exam_page_id
+    JOIN
+    exam_segment AS segment
+      ON segment.exam_id = page.exam_id
+         AND segment.segment_key = page.segment_key
```
millerts: This is the only thing that changed.  We join against `segment_key` instead of `exam_segment_key`

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -203,28 +203,28 @@ public void shouldFindExamWithLatestLanguage() throws InterruptedException {
...
+            .addValue("createdAt", mapJodaInstantToTimestamp(Instant.now().minus(50000)));
 
-        examAccommodationCommandRepository.insert(Collections.singletonList(enuAccommodation));
+        jdbcTemplate.update(SQL, parameters);
 
         Optional<Exam> maybeExam = examQueryRepository.getExamById(exam.getId());
         assertThat(maybeExam).isPresent();
         assertThat(maybeExam.get().getLanguageCode()).isEqualTo("ENU");
 
-        Thread.sleep(5000);
```
millerts: I mentioned this in Slack, but we should never use the Thread.sleep.  It artificially slows down our tests.  I thought things were hanging as I waited five seconds for this to run.  Rather than use the sleep I just inserted the records in the state they needed to be to test the logic.

client/src/main/java/tds/exam/ExamPage.java
```diff
@@ -169,14 +171,14 @@ public String getSegmentKey() {
...
      */
     public String getSegmentId() {
         return segmentId;
     }
 
     /**
-     * @return The position of the segment owns this page
+     * @return The position of the segment owns this page.  This is only populated if the query joins ExamSegment
```
wtritch: This is a really odd doc on a getter.  As somebody without deep knowledge, I wouldn't know if/when this property would be populated.

millerts: Agreed.  I've discussed this with @jeffjohnson9046 , and he will address this domain object with https://jira.fairwaytech.com/browse/TDS-860

PR #159, opened 2017-04-25
Feature/item selection

millerts:
I'm still doing some testing, but I feel confident enough that this can be reviewed.  Any future changes should be as simple as find x value from y instead of z.

A lot of the internal interfaces being used are from ItemSelectionShell.  Feel free to look at the item-selection-components module for more context.

Couple general notes:
- Some classes have unimplemented implementations or have comments like `//Do we need this`.  Since I was focused on getting `FixedForm` assessments working I ignored the implementations needed for adaptive and msb, which i will tackle next.
- There may be some inefficiencies in the way we fetch data, but we're kinda stuck with it for now since we chose to use the existing item selection code rather than rewriting to a spec.


service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
```diff
@@ -0,0 +1,156 @@
...
+        this.examService = examService;
+        this.assessmentService = assessmentService;
+    }
+
+    @Transactional
+    @Override
+    public List<OpportunityItem> createNextPageGroup(final UUID examId, final int lastPage, final int lastPosition) {
+        Exam exam = examService.findExam(examId).orElseThrow(() -> new IllegalArgumentException("Invalid exam id"));
+        Assessment assessment = assessmentService.findAssessment(exam.getClientName(), exam.getAssessmentKey())
+            .orElseThrow(() -> new IllegalArgumentException("bad assessment"));
```
wtritch: nit: maybe include the invalid ids in the exception messages?

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
```diff
@@ -0,0 +1,156 @@
...
+                throw new IllegalStateException(String.format("Could not find an assessment %s item with id %s",
+                    assessment.getKey(),
+                    testItem.getItemID()));
+            }
+
+            //Item position is the item position in the exam.  The TestItem or Item position is the position
+            //within a segment.  So we take the last position by student and then increment.
+            examItemPosition++;
+
+            Item item = itemByKey.get(testItem.getItemID()).get(0);
```
wtritch: Do we always want to grab the "zeroth" or should we randomly grab one from the collection?

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
```diff
@@ -0,0 +1,156 @@
...
+        examPageCommandRepository.insert(page);
+        examItemCommandRepository.insert(examItems.toArray(new ExamItem[examItems.size()]));
+
+        final SimpleDateFormat simpleDateFormat = new SimpleDateFormat(CREATED_AT_FORMAT);
+        final String dateCreated = simpleDateFormat.format(new Date(Instant.now().toEpochMilli()));
+
+        return examItems.stream()
+            .map(examItem -> {
+                OpportunityItem oppItem = new OpportunityItem();
+                Item item = itemByKey.get(examItem.getItemKey()).get(0);
```
wtritch: Do we need to get the item again here?  Doesn't the examItem have the .getItemType() property?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+        List<SegmentHolder> unsatisfiedSegments = segments.stream()
+            .filter(segmentHolder -> !segmentHolder.examSegment.isSatisfied())
+            .collect(Collectors.toList());
+
+        final List<FieldTestItemGroup> fieldTestItemGroups = fieldTestService.findUsageInExam(examId);
+
+        //Only care about non deleted nor administered field tests
+        final Map<String, List<FieldTestItemGroup>> fieldTestItemsBySegmentKey = fieldTestItemGroups.stream()
+            .filter(fieldTestItemGroup -> !fieldTestItemGroup.isDeleted() && fieldTestItemGroup.getAdministeredAt() == null)
+            .collect(Collectors.groupingBy(FieldTestItemGroup::getSegmentKey));
```
wtritch: I'm checking my knowledge here, but I think by filtering out all deleted and administered FieldTestItemGroup instances, we won't be creating an empty value for the key.  Which means we won't have an empty collection value used in the fieldTestItemsSatisfied check on line 109?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+            boolean fieldTestItemsSatisfied = fieldTestItemsBySegmentKey.containsKey(examSegment.getSegmentKey())
+                && fieldTestItemsBySegmentKey.get(examSegment.getSegmentKey()).isEmpty();
+
+            if (examSegment.getAlgorithm().getType().contains(ADAPTIVE)
+                && examSegment.getExamItemCount() == nonFieldTestItemCount
+                && fieldTestItemsSatisfied) {
+                satisfiedSegments.add(examSegment);
+            } else if (examSegment.getAlgorithm().equals(Algorithm.FIXED_FORM)
+                && examSegment.getExamItemCount() == (nonFieldTestItemCount + fieldTestItemCount)) {
+                satisfiedSegments.add(examSegment);
```
wtritch: Is this saying that a segment is satisfied in FIXED_FORM iff the total examSegment item count (answered items?) is equal to the total number of field test + non-field test items in the segment?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+                .count();
+
+            final ExamSegment examSegment = segment.examSegment;
+            boolean fieldTestItemsSatisfied = fieldTestItemsBySegmentKey.containsKey(examSegment.getSegmentKey())
+                && fieldTestItemsBySegmentKey.get(examSegment.getSegmentKey()).isEmpty();
+
+            if (examSegment.getAlgorithm().getType().contains(ADAPTIVE)
+                && examSegment.getExamItemCount() == nonFieldTestItemCount
+                && fieldTestItemsSatisfied) {
+                satisfiedSegments.add(examSegment);
```
wtritch: Is this saying that a segment is satisfied in ADAPTIVE iff the total examSegment item count (answered items?) is equal to the total number of non-field test items in the segment AND all field test items in the segment have been either deleted or administered?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+                .filter(examItem -> !examItem.isFieldTest())
+                .count();
+
+            long fieldTestItemCount = segment.items.stream()
+                .filter(ExamItem::isFieldTest)
+                .count();
+
+            final ExamSegment examSegment = segment.examSegment;
+            boolean fieldTestItemsSatisfied = fieldTestItemsBySegmentKey.containsKey(examSegment.getSegmentKey())
+                && fieldTestItemsBySegmentKey.get(examSegment.getSegmentKey()).isEmpty();
```
wtritch: Should this be:
```
boolean fieldTestItemsSatisfied = !fieldTestItemsBySegmentKey.containsKey(examSegment.getSegmentKey())
    || fieldTestItemsBySegmentKey.get(examSegment.getSegmentKey()).isEmpty();
```
?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+        if (!satisfiedSegments.isEmpty()) {
+            List<ExamSegment> segmentsToUpdate = satisfiedSegments.stream()
+                .map(examSegment -> ExamSegment.Builder
+                    .fromSegment(examSegment)
+                    .withSatisfied(true)
+                    .build())
+                .collect(Collectors.toList());
+
+            examSegmentService.update(segmentsToUpdate.toArray(new ExamSegment[segmentsToUpdate.size()]));
+        }
```
wtritch: This feels like tertiary behavior.  It feels confusing to me when a getter performs cleanup like this... But I don't know enough about the system to suggest an alternative...

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+                    examSegment.getSegmentPosition(),
+                    groupBlock.group,
+                    groupBlock.block,
+                    exam.getExam().getSessionId(),
+                    false,
+                    //TODO - Find out where isActive is coming from in legacy code
+                    null);
+            })
+            .collect(Collectors.toList());
+    }
```
wtritch: Recommend breaking the above method into smaller chunks.

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
millerts: The code that I ported in student always takes the first result.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+        HashSet<String> ftFieldGroups = new HashSet<>();
+        ftFieldGroups.addAll(fieldTestItemGroups);
+        history.set_previousFieldTestItemGroups(ftFieldGroups);
+
+        List<SegmentHolder> segmentHolders = mapSegmentToItems(exam);
+
+        ArrayList<ItemResponse> itemResponses = new ArrayList<>();
+        for(final SegmentHolder segmentHolder : segmentHolders) {
+            List<ItemResponse> responses = segmentHolder.items.stream()
+                .filter(examItem -> examItem.getItemKey() != null)
```
wtritch: Interesting, when do we have examItems without an associated item key?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+                    if(examItem.getResponse().isPresent()) {
+                        ExamItemResponse examItemResponse = examItem.getResponse().get();
+
+                        if(examItemResponse.getScore().isPresent()) {
+                            ExamItemResponseScore score = examItemResponse.getScore().get();
+                            response.setScore(score.getScore());
+                            response.setScoreDimensions(score.getScoringDimensions());
+                            response.loadDimensionScores(score.getScoringDimensions());
+                        }
+                    }
```
wtritch: maybe flip these to guard code?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+                    return response;
+                }).collect(Collectors.toList());
+
+            itemResponses.addAll(responses);
+        }
+
+        history.set_previousResponses(itemResponses);
+
+        return history;
+    }
```
wtritch: Recommend breaking the above method into smaller chunks.  (Like maybe the mapping method?)

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
millerts: Good catch.  Will remove.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+
+    @Override
+    public boolean setSegmentSatisfied(UUID examId, Integer segmentPosition, String reason) throws ReturnStatusException {
```
wtritch: super-nit: finals

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+        List<ExamSegment> segments = expandableExam.getExamSegments();
+        List<ExamItem> items = expandableExam.getExamItems();
+
+        List<SegmentHolder> segmentHolders = segments.stream()
+            .map(SegmentHolder::new)
+            .collect(Collectors.toList());
+
+        for (ExamPage page : expandableExam.getExamPages()) {
+            for (SegmentHolder holder : segmentHolders) {
+                if (holder.isPageInSegment(page)) {
```
wtritch: The holder itself has logic?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -0,0 +1,397 @@
...
+            if (maybeFieldTestItem.isPresent()) {
+                groupId = maybeFieldTestItem.get().getGroupId();
+                blockId = maybeFieldTestItem.get().getBlockId();
+            }
+        }
+
+        return new GroupBlock(groupId, blockId);
+    }
+
+    private class SegmentHolder {
```
wtritch: recommend static

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
millerts: Oh, sorry, I responded a bit to fast here.  What this is doing is that I'm using the groupBy function with the stream on line 73.  There should never be more than 1 item.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
millerts: Good catch for here and your comment above.  Fixed as you suggest and added a test.  I hadn't hit up against this particular logic yet because I've been so focused on Fixed Form

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
millerts: It is true if:

1. Algorithm type for the exam segment is `ADAPTIVE`
2. The number of items for the segment (ignoring field test items) == the number of items expected for a segment.
3. All field test items required for the item group have been satisfied

Sorry if this is information you know, but field test items are items that will be presented and answered by students but are not included in scoring nor satisfy the required number of items for a segment.  FieldTestItems are the way SBAC verify that potential new items fit within their statistical model for student scoring.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
millerts: Yes, that is what the conditional is.  It is a bit odd to me, and I think we could remove the fieldTestItemCount all together, but I'm not sure.

Here is the logic in the legacy code which this is ported from - https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/develop/tds-dll-mysql/src/main/java/tds/dll/mysql/ItemSelectionDLL.java#L339

They use temporary tables so I may have it wrong.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
millerts: Yeah, I don't love this solution either.  However, this is an implementation that is used within the bowels of ItemSelection.  I guess we could do similar logic that we have in this class up in the ExamItemSelectionService layer.  This class still needs to only return the non satisfied segments (ItemCandidatesData representing those unsatisfied segments).

Considering how this is still leveraging legacy flow I'm tempted to leave it as is for now.

PR #164, opened 2017-04-27
Feature/get eligible exam assessments

burzum619:
This endpoint is for retrieving exam/assessment information to determine which assessments are available/unavailable for a student/session, which assessments have exams that can be resumed, and which assessments have previously been taken.

### DIFF
--- a/client/src/main/java/tds/exam/ExamAssessmentInfo.java
+++ b/client/src/main/java/tds/exam/ExamAssessmentInfo.java
@@ -1,0 +1,157 @@
+package tds.exam;
+
+/**
+ * An object containing recent exam data for an assessment
+ */
+public class ExamAssessmentInfo {
+    private String assessmentKey;
+    private String assessmentId;
+    private int attempt;
+    private int maxAttempts;
+    private String assessmentLabel;
+    private String status;
+    private String subject;
+    private String grade;
+    private String deniedReason;
+
+    /**
+     * Private constructor for frameworks
+     */
+    private ExamAssessmentInfo() {}
+
+    private ExamAssessmentInfo(Builder builder) {
+        this.subject = builder.subject;
+        this.assessmentKey = builder.assessmentKey;
+        this.assessmentLabel = builder.assessmentLabel;
+        this.status = builder.status;
+        this.grade = builder.grade;
+        this.attempt = builder.attempt;
+        this.maxAttempts = builder.maxAttempts;
+        this.assessmentId = builder.assessmentId;
+        this.deniedReason = builder.deniedReason;
+    }
+
+    public static class Builder {
+        private String assessmentKey;
+        private String assessmentId;
+        private int attempt;
+        private int maxAttempts;
+        private String assessmentLabel;
+        private String status;
+        private String subject;
+        private String grade;
+        private String deniedReason;
+
+        public Builder withAssessmentKey(String assessmentKey) {
+            this.assessmentKey = assessmentKey;
+            return this;
+        }
+
+        public Builder withAssessmentId(String assessmentId) {
+            this.assessmentId = assessmentId;
+            return this;
+        }
+
+        public Builder withAttempt(int attempt) {
+            this.attempt = attempt;
+            return this;
+        }
+
+        public Builder withMaxAttempts(int maxAttempts) {
+            this.maxAttempts = maxAttempts;
+            return this;
+        }
+
+        public Builder withAssessmentLabel(String assessmentLabel) {
+            this.assessmentLabel = assessmentLabel;
+            return this;
+        }
+
+        public Builder withStatus(String status) {
+            this.status = status;
+            return this;
+        }
+
+        public Builder withSubject(String subject) {
+            this.subject = subject;
+            return this;
+        }
+
+        public Builder withGrade(String grade) {
+            this.grade = grade;
+            return this;
+        }
+
+        public Builder withDeniedReason(String deniedReason) {
+            this.deniedReason = deniedReason;
+            return this;
+        }
+
+        public ExamAssessmentInfo build() {
+            return new ExamAssessmentInfo(this);
+        }
+    }
+
+    /**
+     * @return The key of the assessment for the exam
+     */
+    public String getAssessmentKey() {
+        return assessmentKey;
+    }
+
+    /**
+     * @return The id of the assessment for the exam
+     */
+    public String getAssessmentId() {
+        return assessmentId;
+    }
+
+    /**
+     * @return The current attempt number for the assessment
+     */
+    public int getAttempt() {
+        return attempt;
+    }
+
+    /**
+     * @return The maximum number of attempts for an assessment
+     */
+    public int getMaxAttempts() {
+        return maxAttempts;
+    }
+
+    /**
+     * @return The label for the assessment
+     */
+    public String getAssessmentLabel() {
+        return assessmentLabel;
+    }
+
+    /**
+     * @return The status of exam assessment
+     */
+    public String getStatus() {
+        return status;
+    }
+
+    /**
+     * @return The subject of the assessment
+     */
+    public String getSubject() {
+        return subject;
+    }
+
+    /**
+     * @return The grade of the assessment
+     */
+    public String getGrade() {
+        return grade;
+    }
+
+    /**
+     * @return The reason why the assessment is unavailable
+     */
+    public String getDeniedReason() {
+        return deniedReason;
+    }
+}
--- a/service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
@@ -17,6 +17,7 @@ import java.util.Optional;
 
 import tds.accommodation.Accommodation;
 import tds.assessment.Assessment;
+import tds.assessment.AssessmentInfo;
 import tds.assessment.AssessmentWindow;
 import tds.assessment.SegmentItemInformation;
 import tds.common.cache.CacheType;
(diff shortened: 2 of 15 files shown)


service/src/main/java/tds/exam/services/impl/SessionServiceImpl.java
```diff
@@ -123,4 +128,44 @@ public SessionServiceImpl(final RestTemplate restTemplate, final ExamServiceProp
...
+            builder.build().toUri(),
+            HttpMethod.GET,
+            null,
+            new ParameterizedTypeReference<List<SessionAssessment>>() {});
+
+        return sessionAssessmentsResponse.getBody();
+    }
+
+    @Override
+    public List<Session> findSessionsByIds(final List<UUID> sessionIds) {
```
millerts: Cacheable `SHORT_TERM`

service/src/main/java/tds/exam/services/impl/SessionServiceImpl.java
```diff
@@ -123,4 +128,44 @@ public SessionServiceImpl(final RestTemplate restTemplate, final ExamServiceProp
...
+        final UriComponentsBuilder builder =
+            UriComponentsBuilder
+                .fromHttpUrl(String.format("%s/%s",
+                    examServiceProperties.getSessionUrl(),
+                    SESSION_APP_CONTEXT));
+
+        if (sessionIds.isEmpty()) {
```
millerts: Suggest moving to the first line in this method since nothing else matters if session ids are empty

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -156,4 +157,12 @@ void pauseExamsInSession(@PathVariable final UUID sessionId) {
 
         return new ResponseEntity<>(headers, HttpStatus.NO_CONTENT);
     }
+
+    @RequestMapping(value = "/eligible/{clientName}/{studentId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: What do you think about having this be `exam/eligible?clientName="SBAC"&studentId=1`?  Our URLs aren't the best right now across all the apps, 20/20 hindsight and all, but in my mind the resource is an ExamAssessmentInfo and the other parts are just filters on those resources.  We can make them all required.

Also, what do you think about the `eligible` being changed to `metadata` or `info`?

client/src/main/java/tds/exam/ExamAssessmentInfo.java
```diff
@@ -0,0 +1,157 @@
...
+    private String grade;
+    private String deniedReason;
+
+    /**
+     * Private constructor for frameworks
+     */
+    private ExamAssessmentInfo() {}
+
+    private ExamAssessmentInfo(Builder builder) {
+        this.subject = builder.subject;
```
millerts: If any of these fields cannot be null in practice can you add a checkNotNull around them?

service/src/test/java/tds/exam/web/endpoints/ExamControllerIntegrationTests.java
```diff
@@ -256,4 +257,28 @@ public void shouldReturnUnprocessableEntityWithError() throws Exception {
...
+
+        http.perform(get(new URI(String.format("/exam/eligible/%s/%s", clientName, studentId)))
+            .contentType(MediaType.APPLICATION_JSON)
+            .param("sessionId", sessionId.toString())
+            .param("grade", grade))
+            .andExpect(status().isOk())
+            .andExpect(jsonPath("data.[0].assessmentId", is(examAssessmentInfo1.getAssessmentId().toString())))
+            .andExpect(jsonPath("data.[1].assessmentId", is(examAssessmentInfo2.getAssessmentId().toString())));
+
+        verify(mockExamService).findExamAssessmentInfo(clientName, studentId, sessionId, grade);
```
millerts: Can you add the objectmapper check to make sure the POJO can be serialized/deserialized?

client/src/main/java/tds/exam/ExamAssessmentInfo.java
```diff
@@ -0,0 +1,157 @@
+package tds.exam;
+
+/**
+ * An object containing recent exam data for an assessment
+ */
+public class ExamAssessmentInfo {
```
millerts: What do you think about `ExamAssessmentMetadata` instead of info?  I know assessment is already named info.  Something about Info just isn't sitting well with me.  Not a big deal either way.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -334,6 +344,141 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+
+                        maybeError = canCreateNewExam(clientName, request, recentExam);
+                        isNewOpportunity = !maybeError.isPresent();
+                        currentAttempt = isNewOpportunity
+                            ? recentExam.getAttempts() + 1
+                            : recentExam.getAttempts();
+                    }
+                } else { // If theres no previous exam, only need to validate the opportunity count
+                        /* StudentDLL Lines [10649-10660] */
+                    if (assessment.getMaxAttempts() <= 0) {
```
wtritch: Where do we test if currentAttempt >= maxAttempt?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -776,4 +915,45 @@ private void updateExam(final Exam exam, final Exam updatedExam) {
 
         examStatusChangeListeners.forEach(listener -> listener.accept(exam, updatedExam));
     }
+
+    private Optional<Exam> getMostRecentExamForAssessment(final String assessmentKey, final Map<String, List<Exam>> assessmentExams) {
```
wtritch: This feels like a brittle question.  If we have 100,000 students simultaneously taking the assessment, wouldn't this change 100,000 times in a minute or two?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -334,6 +344,141 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        Set<String> sessionAssessmentIds = sessionService.findSessionAssessments(sessionId).stream()
+            .map(SessionAssessment::getAssessmentId)
+            .collect(Collectors.toSet());
+        /* StudentDLL - 10439 and 10535 - get both student package attributes at once */
+        List<RtsStudentPackageAttribute> attributes = studentService.findStudentPackageAttributes(studentId, clientName, ELIGIBLE_ASSESSMENTS, BLOCKED_SUBJECT);
+        /* StudentDLL.java - 10451 - for _GetCurrentTests, we need assessment metadata for each eligible assessment */
+        List<AssessmentInfo> eligibleAssessments = assessmentService.findAssessmentInfosForAssessments(clientName,
+            getEligibleAssessmentKeysFromAttributes(attributes));
+
+        /* StudentDLL [10520] - _GetOpportunityInfo - Get the exam and session data for all exams taken by this student */
```
wtritch: Ah.  Well, this should be bounded to what? dozens of exams? hundreds?  I'm trying to keep an eye out for things like the ART startup logic where all students in the system are loaded into memory.

service/src/main/java/tds/exam/services/impl/AssessmentServiceImpl.java
```diff
@@ -146,4 +147,40 @@ public AssessmentServiceImpl(final RestTemplate restTemplate, final ExamServiceP
 
         return maybeSegmentItemInfo;
     }
+
+    public List<AssessmentInfo> findAssessmentInfosForAssessments(final String clientName, final String... assessmentKeys) {
```
burzum619: Yes, good point. Certainly cacheable.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Yes, this call could fetch potentially dozens (or even hundreds, during development) of Exams. Fortunately the Exam class isn't quite as thick as our beefy Assessment class.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -334,6 +344,141 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        Map<String, List<Exam>> assessmentExams = studentExams.stream().collect(Collectors.groupingBy(Exam::getAssessmentKey));
+
+        for (AssessmentInfo assessment : eligibleAssessments) {
+            int currentAttempt = 0;
+            String deniedReason = null;
+            String status = null;
+            Optional<Exam> maybeMostRecentExam = getMostRecentExamForAssessment(assessment.getKey(), assessmentExams);
+            boolean isNewOpportunity = false;
+
+            if (sessionId != null && !sessionAssessmentIds.contains(assessment.getId())) {
```
burzum619: In the legacy code, this is corresponds to the stored procedure `session._getopportunityinfo`. I think your explanation is accurate. However, in this particular line you are commenting on (410), sessionAssessmentIds refer to the assessment ids for assessments that are enabled in the current session by the proctor.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: Good point @will, looks like I might have missed that. I'll add a unit test.

burzum619: @millerts In reality, I don't think this is ever the case unless someone messes directly with the configs database tables. The loader script currently defaults maxopportunity to 3 in `client_testproperties`. But I believe this is just a simple check against bad data, and the correct way to disable an assessment would be to set "isselectable = 0" in configs.client_testproperties.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -776,4 +915,45 @@ private void updateExam(final Exam exam, final Exam updatedExam) {
...
+
+        return assessmentExams.get(assessmentKey).stream()
+            .sorted(Comparator.comparing(Exam::getCreatedAt).reversed())
+            .findFirst();
+    }
+
+    private String[] getEligibleAssessmentKeysFromAttributes(final List<RtsStudentPackageAttribute> attributes) {
+        final Optional<String> eligibleAssessmentStr = attributes.stream()
+            .filter(attribute -> ELIGIBLE_ASSESSMENTS.equals(attribute.getName()))
+            .map(RtsStudentPackageAttribute::getValue)
```
burzum619: No, I believe the student package returns all assessmentkeys in one attribute, delimited by a semicolon.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
will: glad I could help

service/src/main/java/tds/exam/web/endpoints/ExamController.java
burzum619: I like both suggestions, thanks.

PR #168, opened 2017-05-01
Support a SATISFIED response when requesting the next ItemGroup.

wtritch:
This PR adds the ability to accept a SATISFIED response from ItemSelectionService#getNextItemGroup.
A SATISFIED response indicates that the exam has been completed, and there IS no next item group to retrieve.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
         <tds-student-client.version>0.0.2</tds-student-client.version>
         <tds-config-client.version>0.0.2</tds-config-client.version>
         <tds-student-library.version>3.0.5</tds-student-library.version>
-        <tds-item-selection.version>3.0.1</tds-item-selection.version>
+        <tds-item-selection.version>3.0.2</tds-item-selection.version>
 
         <maven.compiler.plugin.version>3.5.1</maven.compiler.plugin.version>
         <maven.release.plugin.version>2.5.3</maven.release.plugin.version>
--- a/service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
@@ -3,16 +3,6 @@ package tds.exam.services.impl;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
-
-import java.text.SimpleDateFormat;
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -33,7 +33,7 @@
         <tds-student-client.version>0.0.2</tds-student-client.version>
         <tds-config-client.version>0.0.2</tds-config-client.version>
         <tds-student-library.version>3.0.5</tds-student-library.version>
-        <tds-item-selection.version>3.0.1</tds-item-selection.version>
+        <tds-item-selection.version>3.0.2</tds-item-selection.version>
```
wtritch: This is a bump in the ItemSelectionShell dependency that adds a State enum to the ItemResponse returned from ItemSelectionService#getNextItemGroup.

If the status is SATISFIED, then all item groups have been served, and there is no "next" item group.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -126,6 +121,10 @@ public ItemCandidatesData getItemCandidates(UUID examId) throws ReturnStatusExce
             updateSatisfiedSegments(satisfiedSegments);
         }
 
+        if (nonSatisfiedSegments.isEmpty()) {
+            return Collections.singletonList(new ItemCandidatesData(examId, IItemSelectionDLL.SATISFIED));
+        }
```
wtritch: If we have no non-satisfied segments, return a special ItemCandidatesData response with a special SATISFIED algorithm.  (This is based on legacy code, "special" values usually cause hard-to-diagnose subtle bugs.)

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
```diff
@@ -61,7 +63,9 @@ public ExamItemSelectionServiceImpl(final ItemSelectionService itemSelectionServ
 
         ItemResponse<ItemGroup> response = itemSelectionService.getNextItemGroup(exam.getId(), assessment.isMultiStageBraille());
 
-        if (response.getErrorMessage().isPresent()) {
+        if (response.getResponseStatus() == SATISFIED) {
+            return Collections.emptyList();
+        } else if (response.getErrorMessage().isPresent()) {
```
wtritch: If all item groups have already been returned (the exam is SATISFIED) then return an empty list.

PR #169, opened 2017-05-01
Feature/cleanup exam page

millerts:


service/src/main/java/tds/exam/services/impl/ExamPageServiceImpl.java
```diff
@@ -78,7 +77,7 @@ public void deletePages(final UUID examId) {
 
 
     @Override
-    public Response<ExamPage> getPage(final UUID examId, final int pageNumber) {
+    public ExamPage getPage(final UUID examId, final int pageNumber) {
```
millerts: This is the key change.  A `Response` object is to be used when we want to return data along with the failure condition.  For example, if you have validation errors.  However, as you can see this either returns an ExamPage or throws.

I have changes that accommodate this on the student side.  I will merge the two projects at the same time.

millerts:
Closing the PR because after seeing usage in Student I need to make more extensive changes.

PR #170, opened 2017-05-01
TDS-519: Feature/review exam

jeffjohnson9046:
[TDS-519](https://jira.fairwaytech.com/browse/TDS-519): Endpoint for review exam
This PR will also cover [TDS-536](https://jira.fairwaytech.com/browse/TDS-536): Complete Exam: Exam service should set status to 'review'


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -590,6 +593,20 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         return new Response<>(examConfig);
     }
 
+    @VerifyAccess
+    @Override
+    public Optional<ValidationError> reviewExam(final UUID examId) {
+        boolean isComplete = examSegmentService.checkIfSegmentsCompleted(examId);
+        if (!isComplete) {
```
jtreuting: seems like the if statement could include the call to `checkIfSegmentsCompleted` directly

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -590,6 +593,20 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+            ValidationError error = new ValidationError(ValidationErrorCode.EXAM_INCOMPLETE, "Review Test: Cannot end test because test length is not met.");
```
jtreuting: I'm not sure the message is clear (at least to me).  is it really "Cannot end test because the test segments are not all complete"

millerts: However, is this message used directly on the student side @jeffjohnson9046 ?  If not, then I agree with @jtreuting

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -84,6 +84,22 @@ public ExamController(ExamService examService) {
         return ResponseEntity.ok(examConfiguration);
     }
 
+    @RequestMapping(value ="/{examId}/review", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: Surprised you didn't use `@PutMapping`

millerts:
@burzum619 maybe we should go over this in person after today's standup.  I'm not following the validation that is missing that you've commented on in both PRs related to this functionality.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -84,6 +84,22 @@ public ExamController(ExamService examService) {
...
+    ResponseEntity<NoContentResponseResource> reviewExam(@PathVariable final UUID examId) {
+        final Optional<ValidationError> maybeReviewExamFailure = examService.reviewExam(examId);
+
+        if (maybeReviewExamFailure.isPresent()) {
+            NoContentResponseResource response = new NoContentResponseResource(maybeReviewExamFailure.get());
+            return new ResponseEntity<>(response, HttpStatus.UNPROCESSABLE_ENTITY);
```
wtritch: Should we return the reason for the failure to the caller?  Right now it looks like we never log or respond with the ValidationError's message, or am I missing something?

service/src/main/java/tds/exam/web/endpoints/ExamController.java
```diff
@@ -84,6 +84,22 @@ public ExamController(ExamService examService) {
...
+        }
+
+        Link link = linkTo(methodOn(ExamController.class).getExamById(examId)).withSelfRel();
+        final HttpHeaders headers = new HttpHeaders();
+        headers.add("Location", link.getHref());
+
+        return new ResponseEntity<>(headers, HttpStatus.NO_CONTENT);
```
wtritch: Huh... I don't think I've seen a redirect response from a RESTful endpoint before.  Do we need to do this?
Option 1: the caller receives a 200 OK and goes and fetches the entity
Option 2: we inject the EntityService in here and serve it up if validation succeeds
My vote is Option 1 =)

service/src/test/java/tds/exam/web/endpoints/ExamControllerTest.java
```diff
@@ -227,6 +228,36 @@ public void shouldApproveAccommodationsAndReturnNoErrors() {
...
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.NO_CONTENT);
+        assertThat(response.getHeaders().containsKey("Location")).isTrue();
+    }
+
+    @Test
+    public void shouldReturnUnprocessableEntityWhenExamCannotBeReviewedBecauseItIsIncomplete() {
+        final UUID examId = UUID.randomUUID();
+        final ValidationError error = new ValidationError(ValidationErrorCode.EXAM_INCOMPLETE,
+            "Review Test: Cannot end test because test length is not met.");
```
wtritch: As the caller, I'd love to see this message if validation failed for any reason.

service/src/main/java/tds/exam/web/endpoints/ExamController.java
millerts: We return the ValidationError object.  It has a message and the error code.  We probably don't need that object anymore (20/20 hindsight and all) because we have to construct the error messages that eventually show on screen due to Student's design.  Anyhow, the ValidationError has a code and a message.  The idea was that student would see the code and get the translated message.  

That being said seeing how this is used in student I also suggest just sending the message back in the body.

wtritch: So a NoContentResponseResource actually sends content to the caller?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
jeffjohnson9046: This message is the exact text from the legacy student application.  To the best of my knowledge, this message is not actually displayed anywhere; it's returned to the caller and ignored.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -246,9 +250,11 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         Exam exam = examQueryRepository.getExamById(examId)
             .orElseThrow(() -> new NotFoundException(String.format("Exam could not be found for id %s", examId)));
 
-        if (!StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), newStatus.getCode())) {
+        if (!statusChangeValidators.stream().allMatch(v -> v.validate(exam, newStatus))) {
             return Optional.of(new ValidationError(ValidationErrorCode.EXAM_STATUS_TRANSITION_FAILURE,
-                String.format("Transitioning exam status from %s to %s is not allowed", exam.getStatus().getCode(), newStatus.getCode())));
+                String.format("Transitioning exam status from %s to %s is not allowed",
+                    exam.getStatus().getCode(),
+                    newStatus.getCode())));
```
wtritch: Is there any way to allow the individual validators to add a validation failure message?  e.g. I'd like to know *why* I can't go from x -> y

service/src/main/java/tds/exam/utils/statusvalidators/DefaultStatusChangeValidator.java
```diff
@@ -0,0 +1,16 @@
...
+import tds.exam.utils.ExamStatusChangeValidator;
+import tds.exam.utils.StatusTransitionValidator;
+
+@Component
+public class DefaultStatusChangeValidator implements ExamStatusChangeValidator {
+    @Override
+    public boolean validate(final Exam exam, final ExamStatusCode intendedStatus) {
+        return StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), intendedStatus.getCode());
+    }
+}
```
wtritch: This class feels like an odd wrapper around StatusTransitionValidator.  Is this now the only thing using StatusTransitionValidator?  And, if so, could we just change StatusTransitionValidator to implement ExamStatusChangeValidator?

service/src/main/java/tds/exam/utils/statusvalidators/ReviewStatusChangeValidator.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    public boolean validate(final Exam exam, final ExamStatusCode intendedStatus) {
+        if (!intendedStatus.getCode().equals(ExamStatusCode.STATUS_REVIEW)) {
+            return true;
+        }
+
+        return examSegmentService.checkIfSegmentsCompleted(exam.getId())
+            && StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), intendedStatus.getCode());
```
wtritch: I think the change validators should be orthogonal.  The StatusTransitionValidator.isValidTransition should have already been checked by a different ExamStatusChangeValidator, right?

jeffjohnson9046:
Pursuant to our in-person code review this morning, I am closing this PR.  A new PR will be opened containing the changes we discussed.

PR #171, opened 2017-05-02
Creating exam reported message listener to update exam status

burzum619:


service/src/main/java/tds/exam/configuration/messaging/ExamMessagingConfiguration.java
```diff
@@ -25,6 +36,26 @@ public TopicExchange exchange() {
...
     }
 
+    @Bean
+    public Queue examReportedQueue() {
+        return new Queue(QUEUE_EXAM_REPORTED, true);
+    }
+
+    @Bean
+    public Binding examReportedBinding(@Qualifier("examReportedQueue") final Queue queue,
+                                         @Qualifier("exchange") final TopicExchange exchange) {
```
millerts: It looks to me like maybe you need to run formatter on this class?

service/src/main/java/tds/exam/messaging/ExamReportedMessageListener.java
```diff
@@ -0,0 +1,37 @@
...
+     * message from our message broker.
+     *
+     * @param examId The completed exam id
+     */
+    public void handleMessage(final String examId) {
+        log.debug("Received reported exam notification for id: {}", examId);
+        try {
+            examService.updateExamStatus(UUID.fromString(examId), new ExamStatusCode(ExamStatusCode.STATUS_REPORTED));
+        } catch (NotFoundException e) {
+            log.error("Unable to update exam status for the exam: " + e.getMessage());
```
millerts: Can you add a comment here explaining why we're logging rather than handling the Exception.  Basically, what we discussed in the parking lot today.

PR #172, opened 2017-05-02
Feature/exam segment wrapper

millerts:
This is a pretty big change.  We have a need in student to get segment, page, and item information in a single call.  After some discussion with @wtritch and @jeffjohnson9046 this was the design we came up with.

Major changes:
- ExamPage no longer has a list of Items or Segment data that isn't in the table that backs it.  This additional data isn't always populated and led to quite a bit of confusion as you'll note by the way it spread throughout the code base.
- ExamPageWrapper is a new class that contains the ExamPage and its ExamItems
- ExamSegmentWrapper contains its ExamSegment and assocaited ExamPages


service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -25,13 +29,83 @@
...
 import tds.exam.repositories.ExamPageQueryRepository;
+import tds.exam.wrapper.ExamPageWrapper;
 
 @Repository
 public class ExamPageQueryRepositoryImpl implements ExamPageQueryRepository {
     private final NamedParameterJdbcTemplate jdbcTemplate;
     private static final ExamPageRowMapper examPageRowMapper = new ExamPageRowMapper();
     private static final ExamPageResultSetExtractor examPageResultExtractor = new ExamPageResultSetExtractor();
 
+    private static final String EXAM_PAGE_WRAPPER_WITH_ITEM_SELECT = "SELECT \n" +
```
millerts: Extracted this out since this is common amongst multiple queries.

service/src/main/java/tds/exam/web/endpoints/ExamPageController.java
```diff
@@ -1,50 +0,0 @@
...
-import java.util.UUID;
-
-import tds.common.Response;
-import tds.exam.ExamPage;
-import tds.exam.services.ExamPageService;
-import tds.exam.web.annotations.VerifyAccess;
-
-@RestController
-@RequestMapping("/exam")
-public class ExamPageController {
```
millerts: This no longer is needed based on the changes.  It is used by student but I will be fixing student.

service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
```diff
@@ -54,4 +56,32 @@ public ExamSegmentController(final ExamSegmentService examSegmentService) {
 
         return new ResponseEntity<>(HttpStatus.NO_CONTENT);
     }
+
+    @GetMapping(value = "{examId}/segmentWrappers", produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<List<ExamSegmentWrapper>> findExamSegmentWrappersForExam(@PathVariable final UUID examId) {
+        List<ExamSegmentWrapper> wrappers = examSegmentService.findAllExamSegments(examId);
+
+        if (wrappers.isEmpty()) {
```
millerts: If we don't get any exam segments it means that there are no exam segments for an exam.  This really isn't possible if the exam id is valid.  Return not found rather than an empty list.  Could go either way.  @wtritch ?

service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
```diff
@@ -54,4 +56,32 @@ public ExamSegmentController(final ExamSegmentService examSegmentService) {
...
+        }
+
+        return ResponseEntity.ok(wrappers);
+    }
+
+    @GetMapping(value = "{examId}/segmentWrappers/{segmentPosition}", produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<ExamSegmentWrapper> findExamSegmentWrapperForExamAndSegmentPosition(@PathVariable final UUID examId,
+                                                                                       @PathVariable final int segmentPosition,
+                                                                                       @RequestParam(value = "pagePosition", required = false) Integer pagePosition) {
+        if (pagePosition == null) {
```
millerts: I went back and forth here if this should be in the service layer or here.  I decided to leave this here to keep the exam segment service interface clear.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -116,85 +190,54 @@ public ExamPageQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final NamedPa
     }
 
     @Override
-    public Optional<ExamPage> findPageWithItems(final UUID examId, final int position) {
+    public Optional<ExamPageWrapper> findPageWithItems(final UUID examId, final int position) {
```
wtritch: It feels like this belongs in an ExamPageWrapperService rather than having the ExamPageService dealing with multiple resources.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
```diff
@@ -25,13 +29,83 @@
...
+        "       exam_item_id, \n" +
+        "       MAX(id) AS id\n" +
+        "   FROM \n" +
+        "       exam_item_response \n" +
+        "   GROUP BY \n" +
+        "       exam_item_id) most_recent_response \n" +
+        "   ON item.id = most_recent_response.exam_item_id \n" +
+        "LEFT JOIN \n" +
+        "   exam_item_response response \n" +
+        "   ON most_recent_response.id = response.id \n";
```
wtritch: Ah.  This makes sense, I just didn't realize we were going to use custom SQL to saturate an ExamPageWrapper, I pictured an ExamPageWrapperService that had an ExamService, ItemService, etc injected into it.

millerts: I thought about doing that but this already existed and fetches all the data.  Figure might as well let the database do the work and parse it.

service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
wtritch: Age-old question.  I've seen it done both ways.  I think a 404 is reasonable.

service/src/main/java/tds/exam/repositories/impl/ExamPageQueryRepositoryImpl.java
millerts: Yeah, I thought about the same thing while in the middle of doing it.  Will separate all resources into separate controller/service/repository

service/src/main/java/tds/exam/web/endpoints/ExamSegmentController.java
jeffjohnson9046: I think the 404 response is correct, especially with the "This really isn't possible if the exam id is valid" statement.

millerts:
Comments addressed.

PR #174, opened 2017-05-03
TDS-519: Feature/review exam refactor

jeffjohnson9046:
[TDS-519](https://jira.fairwaytech.com/browse/TDS-519):  Exam service side of review exam based on code review feedback.

**NOTE:** The [TDS_Student PR](https://github.com/SmarterApp/TDS_Student/pull/70) depends on the code in this PR.


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -364,7 +374,7 @@ public void pauseAllExamsInSession(final UUID sessionId) {
             examAssessmentMetadatas = findEligibleExamAssessmentsForGuest(session.getClientName(), grade);
         }
 
-        return new Response(examAssessmentMetadatas);
+        return new Response<>(examAssessmentMetadatas);
```
jeffjohnson9046: "unchecked" warning fix.

service/src/main/java/tds/exam/utils/statusvalidators/ReviewStatusChangeValidator.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    public boolean validate(final Exam exam, final ExamStatusCode intendedStatus) {
+        if (!intendedStatus.getCode().equals(ExamStatusCode.STATUS_REVIEW)) {
+            return true;
+        }
+
+        return examSegmentService.checkIfSegmentsCompleted(exam.getId())
+            && StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), intendedStatus.getCode());
```
millerts: Doesn't this result in this check being done multiple times?  Once here and once in the other status change validator?

service/src/test/java/tds/exam/web/endpoints/ExamControllerTest.java
```diff
@@ -43,6 +43,7 @@
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.verifyZeroInteractions;
```
millerts: Is this actually used?

service/src/main/java/tds/exam/utils/statusvalidators/ReviewStatusChangeValidator.java
jeffjohnson9046: True enough - I'll remove it.

service/src/test/java/tds/exam/web/endpoints/ExamControllerTest.java
jeffjohnson9046: Sure it is - just not here.  I'll pull this out before merging to `develop`.

service/src/main/java/tds/exam/utils/ExamStatusChangeValidator.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface ExamStatusChangeValidator {
+    /**
+     * Validate that an {@link tds.exam.Exam} can transition from one state to another.
+     *
+     * @param exam The {@link tds.exam.Exam} containing the current {@link tds.exam.ExamStatusCode}
+     * @param intendedStatus The {@link tds.exam.ExamStatusCode} to which the {@link tds.exam.Exam} wants to change
+     * @return True if the status transition is valid; otherwise false
+     */
+    boolean validate(final Exam exam, final ExamStatusCode intendedStatus);
```
jeffjohnson9046: @wtritch : In the last PR, you mentioned something about returning a custom error message or some sort of detail on _why_ the validator fails.  While I think returning `boolean` to indicate whether it worked or didn't, I don't love it; when you call these in a loop you can't (easily) determine which `ExamStatusChangeValidator` failed; only that one of them did.

I considered having the `validate` method return an `Optional<ValidationError>` (or de facto object for handling ... well... validation failures).  When I tried it, I didn't really like the way the code felt so I backed out.  Any thoughts?

wtritch: I would model it after an existing validation framework know to work well like JSR 303 which does something like this:

Set<constraintviolation<member>> violations = validator.validate(member)

millerts: For the sake of time I think we could change the validators to return a `Set<ConstraintViolation>` where `ConstraintViolation` is our own class.  I'm not sure the type stuff in JSR303 is necessary for this particular change, but we could still add it in the future if we want.

```
public class ConstraintViolation {
   String getMessage();
}
```

jeffjohnson9046: So I actually just updated this a few minutes ago to jive with what we've typically been doing:

* The `ExamStatusChangeValidator#validate` method returns an `Optional<ValidationError>`
* The `ExamServiceImpl#updateStatus` method loops through each `ExamStatusChangeValidator`.  If it gets a `ValidationError`, it returns it to the caller.

... That should at least give us the ability to return a custom error message from an `ExamStatusChangeValidator` implementation.

This isn't the most elegant solution and I'm certain there's a much cleaner/better way to do it.  However, I think it's consistent with how we're handling validation failures now.  I'll write up a story to revisit this at a later date with a more comprehensive fix.

service/src/main/java/tds/exam/utils/statusvalidators/DefaultStatusChangeValidator.java
```diff
@@ -0,0 +1,27 @@
...
+
+import tds.common.ValidationError;
+import tds.exam.Exam;
+import tds.exam.ExamStatusCode;
+import tds.exam.error.ValidationErrorCode;
+import tds.exam.utils.ExamStatusChangeValidator;
+import tds.exam.utils.StatusTransitionValidator;
+
+@Component
+public class DefaultStatusChangeValidator implements ExamStatusChangeValidator {
```
millerts: Can we rename this from `Default` to `StatusTransitionStatusChangeValidator` or something similar?

service/src/main/java/tds/exam/utils/statusvalidators/StatusTransitionChangeValidator.java
```diff
@@ -0,0 +1,27 @@
...
+import tds.exam.ExamStatusCode;
+import tds.exam.error.ValidationErrorCode;
+import tds.exam.utils.ExamStatusChangeValidator;
+import tds.exam.utils.StatusTransitionValidator;
+
+@Component
+public class StatusTransitionChangeValidator implements ExamStatusChangeValidator {
+    @Override
+    public Optional<ValidationError> validate(final Exam exam, final ExamStatusCode intendedStatus) {
+        if (!StatusTransitionValidator.isValidTransition(exam.getStatus().getCode(), intendedStatus.getCode())) {
```
wtritch: If this is the only place the StatusTransitionValidator is used, I'd recommend consolidating the two classes into here.

PR #178, opened 2017-05-04
TDS-875:  Bugfix/qa view update

jeffjohnson9046:
[TDS-875](https://jira.fairwaytech.com/browse/TDS-875):  Additional data for the QA views to support the testing effort.


service/src/main/resources/db/migration/V1493865701__exam_update_qa_views.sql
```diff
@@ -0,0 +1,127 @@
...
+  exam.browser_id AS _fk_browser,
+  response.is_valid AS isvalid,
+  response.score_latency AS scorelatency,
+  CASE page.are_group_items_required
+    WHEN 1 THEN -1
+    ELSE 0
+  END AS groupitemsrequired,
+  response.scoring_status AS scorestatus,
+  response.scored_at AS scoringdate,
+  'not migrated' AS scoremark,
```
millerts: exam_response has this column.

service/src/main/resources/db/migration/V1493865701__exam_update_qa_views.sql
```diff
@@ -0,0 +1,127 @@
...
+  response.scoring_rationale AS scorerationale,
+  'not migrated' AS scoreattempts,
```
millerts: I believe this can be calculated by taking the number of records that have a distinct score_sent_at

service/src/main/resources/db/migration/V1493865701__exam_update_qa_views.sql
```diff
@@ -0,0 +1,127 @@
...
+  item.item_key AS _efk_itemkey,
+  'not migrated' AS _fk_responsesession,
```
millerts: Not sure if this is necessary but I would assume this can be fetched by getting the joined exam's session.

PR #181, opened 2017-05-06
Fixing a bug where marking an item that did not yet have a response f…

burzum619:
…or review was returning an error (422)

https://jira.fairwaytech.com/browse/TDS-893


burzum619:
Yeah, that was my goal, but it looks like the ExamItemResponse object throws an exception if the sequence is < 1

PR #186, opened 2017-05-10
Adding LIMIT 1 to query to prevent multiple records from being returned

burzum619:
https://jira.fairwaytech.com/browse/TDS-885

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -179,7 +179,8 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
                 "   AND e.client_name = :clientName \n" +
                 "   AND lang.type = 'Language'" +
                 "ORDER BY \n" +
-                "   e.created_at DESC";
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -179,7 +179,8 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 "   AND e.client_name = :clientName \n" +
                 "   AND lang.type = 'Language'" +
                 "ORDER BY \n" +
-                "   e.created_at DESC";
+                "   e.created_at DESC \n" +
+                "LIMIT 1";
```
millerts: What is bringing back multiple records?  Or was this always happening?

burzum619: This query, when attempting to take an exam for a third time, was returning two records instead of the most recent single exam.

burzum619: We just were never able to finish an exam before, so we never hit this bug early on.

PR #189, opened 2017-05-12
Adding missing fields needed for the TRT/ERT

burzum619:
Adding fields to ExpandableExam needed by the ExamResultsTransmitter (TRT). The integration into ERT will be a separate PR in that repository.


client/src/main/java/tds/exam/ExpandableExam.java
```diff
@@ -19,8 +23,11 @@
     private List<ExamineeNote> examineeNotes;
     private List<ExamineeAttribute> examineeAttributes;
     private List<ExamineeRelationship> examineeRelationships;
+    private List<ExamSegmentWrapper> examSegmentWrappers;
```
millerts: A note here for other reviews.  @burzum619 and I chatted about this yesterday, and I asked him to make an additive change here.  I'm going to modify item selection as I do adaptive to use this new construct, and we can then delete the ExamSegment, ExamPage, and ExamItem properties on this class.

The nice thing about the ExpandableExam is that the code @burzum619 wrote allows us to pass in what is populated so we're not fetching extra data even with this change.

service/src/main/java/tds/exam/mappers/impl/WindowAttemptsExpandableExamMapper.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void updateExpandableMapper(final Set<ExpandableExamAttributes> expandableAttributes, final Map<UUID, ExpandableExam.Builder> examBuilders, final UUID sessionId) {
+        if (!expandableAttributes.contains(ExpandableExamAttributes.WINDOW_ATTEMPTS)) {
+            return;
+        }
+
+        examBuilders.forEach((examId, builder) -> {
+            Exam exam = examService.findExam(examId)
```
millerts: Rather than fetch the exam can't you just do builder.build() which will give you an Exam?

service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
```diff
@@ -123,6 +123,48 @@ public int getCurrentExamItemPosition(final UUID examId) {
...
     }
 
+    @Override
+    public Map<UUID, Integer> getResponseUpdateCounts(final UUID examId) {
+        // Needs to toString() each UUID individually
+        final SqlParameterSource params = new MapSqlParameterSource("examId", examId.toString());
+        final String SQL =
+            "SELECT \n" +
+                "   exam_item_id, \n" +
+                "   COUNT(exam_item_response_id) AS response_count \n" +
```
millerts: Do you need to look if the response has changed?  Maybe if you did a distinct you'd get the data you want?  Or maybe now that we don't do async scoring this isn't an issue.  I just worry that there could be multiple changes to the response table (since it is effectively an event table) that have nothing to do with the student responding.

service/src/main/java/tds/exam/mappers/impl/WindowAttemptsExpandableExamMapper.java
burzum619: I like that way more. Thanks

service/src/main/java/tds/exam/repositories/impl/ExamItemQueryRepositoryImpl.java
burzum619: That's a good point. I suppose like you mentioned, doing a DISTINCT and including the response in my selector (even if I'm not actually mapping it to anything) should work for us.

burzum619: Come to think of it, I'll replace the "IR.id" with "IR.response", since ID will always be distinct for any response/event

PR #190, opened 2017-05-12
add group id to exam item

millerts:
This is going to be leveraged in the adaptive code.


service/src/main/resources/db/migration/V1494607731__exam_item_group_id.sql
```diff
@@ -0,0 +1,14 @@
...
+
+  Desc: Adds a group id column to the exam item
+
+***********************************************************************************************************************/
+
+USE exam;
+
+ALTER TABLE exam_item ADD COLUMN group_id VARCHAR(100);
+
+UPDATE exam_item set group_id = '' where group_id is null;
```
millerts: As discussed in the general channel this won't happen in practice moving forward.  Since we have legacy data prior to this change I just figured "" is good enough for now.

PR #198, opened 2017-05-16
Persist positive item scores.

wtritch:


service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -118,70 +117,71 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
...
         ExamItemResponseScore.Builder updatedScoreBuilder = new ExamItemResponseScore.Builder();
+
+        if (responseUpdate.getValue() == null) {
+            updatedResponseBuilder.withScore(updatedScoreBuilder.build());
+            examItemCommandRepository.insertResponses(updatedResponseBuilder.build());
+            return new ReturnStatus("updated");
+        }
+
         final Instant now = Instant.now();
-        if (responseUpdate.getValue() != null || DbComparator.lessThan(score, 0)) {
```
wtritch: The '||' here was the root problem.  If responseUpdate.getValue was ever non-null, we saved a score of -1.  As long as I was in here I did a little restructuring to prevent dipping in and out of 
```
if (responseUpdate.getValue() != null)
```
conditional blocks.

service/src/test/java/tds/exam/services/scoring/ResponseServiceImplTest.java
```diff
@@ -80,4 +87,57 @@ public void itShouldThrowIfExamNotInCorrectStatus() throws ReturnStatusException
...
+
+        responseService.updateScoredResponse(examInstance, itemResponseUpdate,
+            1, "Scored", "scoreRationale", 123L, 456L);
+
+        final ArgumentCaptor<ExamItemResponse> responseCaptor = ArgumentCaptor.forClass(ExamItemResponse.class);
+        verify(mockExamItemCommandRepository).insertResponses(responseCaptor.capture());
+
+        final ExamItemResponse persisted = responseCaptor.getValue();
+        assertThat(persisted.getScore().get().getScore()).isEqualTo(1);
+    }
```
wtritch: Wrote a little test to make sure we persist scores > 0

client/src/main/java/tds/exam/ExamItem.java
```diff
@@ -139,6 +144,24 @@ public Builder withGroupId(final String groupId){
...
+            withPosition(examItem.getPosition());
+            withRequired(examItem.isRequired());
+            withFieldTest(examItem.isFieldTest());
+            withResponse(examItem.getResponse().orNull());
+            withItemFilePath(examItem.getItemFilePath());
+            withStimulusFilePath(examItem.getStimulusFilePath().orNull());
+            withCreatedAt(examItem.getCreatedAt());
+            withGroupId(examItem.getGroupId());
+            return this;
+        }
```
wtritch: While writing my test, I was missing a "copy" method on the builder.  The idea was to use random(ExamItem.class) to generate a base, copy it into the builder, then set the values I cared about.  Unfortunately, the builder performs validation on the setters rather than in the final build() method... so I was unable to use random() without restructuring the Builder.  Still, once built I decided to leave this in.

millerts: Yeah.  I went back and forth on whether it is better to throw as soon as possible or within the build.  The check ensures that you can never set a value that will then throw when you try to build the object.

client/src/main/java/tds/exam/ExamItem.java
```diff
@@ -139,6 +144,24 @@ public Builder withGroupId(final String groupId){
             return this;
         }
 
+        public Builder fromExamItem(final ExamItem examItem) {
```
millerts: Our pattern has been to make these static so you'd do `ExamItem.Builder.fromExamItem(examItem).with...`

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
millerts: @burzum619 , looks like you were correct in your earlier analysis 👍

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -118,70 +117,71 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
...
-
-        if (responseUpdate.getValue() != null) {
-            updatedResponseBuilder.withResponse(
-                StringUtils.replace(responseUpdate.getValue(), "\\", "\\\\"));
-            updatedScoreBuilder.withScoreLatency(scoreLatency);
-            updatedScoreBuilder.withScoringStatus(scoreStatus == null ? null : ExamScoringStatus.fromType(scoreStatus));
-            updatedScoreBuilder.withScoringRationale(scoreRationale);
-            updatedScoreBuilder.withScoringDimensions(buildScoreInfoNode(score, "overall", scoreStatus));
+        } else if (DbComparator.greaterOrEqual(scoreLatency, 0)) {
+//            updatedScoreBuilder.withScoreMark(null); // any previous scoremark is now obsolete
```
millerts: I'm not sure why student does this but can you uncomment out this line?  Once scored the system expects score mark to be null.  I think we can probably remove score mark long term, but short term this would make it align with legacy data.

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -118,70 +117,71 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
...
+        examItemCommandRepository.insertResponses(updatedResponseBuilder.build());
 
-            ExamPage examPage = ExamPage.Builder.fromExamPage(maybeExamPage.get())
-                .withDuration(pageDuration)
-                .build();
+        final Optional<ExamPage> maybeExamPage = examPageService.find(existingExamItem.getExamPageId());
 
-            examPageService.update(examPage);
+        //Something is misconfigured if this is ever true
+        if (!maybeExamPage.isPresent()) {
```
millerts: Can you do `ExamPage exampage = examPageService.find(existingExamItem.getExamPageId()).orElseThrow(() -> new ReturnStatusException("fail"))` or does ReturnStatusException not jive with that interface?

client/src/main/java/tds/exam/ExamItem.java
wtritch: fixed

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
wtritch: fixed

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
wtritch: Fixed.  Changed the other non-optional optionals to the same pattern.

PR #200, opened 2017-05-17
Continue processing when rubric is not found

mtrupkin:
Fixes TDS-949 POST_updateResponses error on Perf Grade 4 MATH


scoring/src/main/java/tds/score/services/impl/ItemScoringServiceImpl.java
```diff
@@ -366,7 +366,9 @@ public ItemScore scoreItem(UUID oppKey, IItemResponseScorable responseScorable,
                     machineRubric.setData(rubricService.findOne(machineRubric.getData()));
                 } catch (final IOException e) {
                     LOG.error("Failed to load scoring rubric for item: {}", itemID, e);
```
millerts: Let's change this log message to warn instead of error since it is a handled error.

PR #201, opened 2017-05-17
Remove caching around SessionService calls.

wtritch:
I've removed all caching around calls to the SessionService.  I've left in caching around the AssessmentService, StudentService, and ConfigService.

### DIFF
--- a/scoring/pom.xml
+++ b/scoring/pom.xml
@@ -20,14 +20,13 @@
         <ss-multijar.version>3.0.6</ss-multijar.version>
         <item-scoring.version>3.0.1</item-scoring.version>
         <item-renderer.version>3.0.9</item-renderer.version>
-        <tds-student-common.version>3.0.4</tds-student-common.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>org.opentestsystem.delivery</groupId>
             <artifactId>student-library</artifactId>
-            <version>${tds-student-common.version}</version>
+            <version>${tds-student-library.version}</version>
             <exclusions>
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/SessionServiceImpl.java
```diff
@@ -61,7 +57,6 @@ public SessionServiceImpl(final RestTemplate restTemplate, final ExamServiceProp
     }
 
     @Override
-    @Cacheable(CacheType.LONG_TERM)
```
millerts: This can stay long term.  I believe this should probably be in ConfigsService long term, but for now it resides in Session service.  The configuration is static config data that doesn't change once configured.  However, that lives in session schema :/ which is why it is in the session service.

TLDR - Leave it long term for now

wtritch: done

scoring/pom.xml
```diff
@@ -20,14 +20,13 @@
...
         <item-renderer.version>3.0.9</item-renderer.version>
-        <tds-student-common.version>3.0.4</tds-student-common.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>org.opentestsystem.delivery</groupId>
             <artifactId>student-library</artifactId>
-            <version>${tds-student-common.version}</version>
+            <version>${tds-student-library.version}</version>
```
wtritch: Fixed a place where we were double-declaring the student-common version in ExamService.

PR #202, opened 2017-05-18
TDS-241: Feature/exam db tuning

jeffjohnson9046:
[TDS-241](https://jira.fairwaytech.com/browse/TDS-241):  This is a first pass at optimizing _some_ of the queries and indexes in the `exam` database.  This PR affects most queries in the `ExamQueryRepositoryImpl`.  There will be additional PRs for addressing this story; the intent is to iterate quickly on these tuning changes and get them into the load test environment ASAP.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -146,42 +146,36 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
         final SqlParameterSource parameters = new MapSqlParameterSource(queryParameters);
 
         String query =
-            "SELECT " +
+            "SELECT \n" +
                 EXAM_QUERY_COLUMN_LIST +
-                "FROM exam.exam e\n" +
-                "JOIN ( \n" +
-                "   SELECT \n" +
-                "       exam_id, \n" +
-                "       MAX(id) AS id \n" +
-                "   FROM \n" +
-                "       exam.exam_event \n" +
-                "   WHERE deleted_at IS NULL \n" +
-                "   GROUP BY exam_id \n" +
-                ") last_event \n" +
-                "  ON e.id = last_event.exam_id \n" +
-                "JOIN exam.exam_event ee \n" +
-                "  ON last_event.exam_id = ee.exam_id AND \n" +
-                "     last_event.id = ee.id \n" +
-                "JOIN exam.exam_status_codes esc \n" +
-                "  ON esc.status = ee.status \n" +
-                "LEFT JOIN exam.exam_accommodation lang \n" +
-                "  ON lang.exam_id = e.id \n" +
-                "  AND lang.created_at = \n" +
-                "  ( \n" +
+                "FROM \n" +
+                "   exam.exam e \n" +
+                "JOIN \n" +
+                "   exam.exam_event ee \n" +
+                "   ON ee.exam_id = e.id \n" +
+                "   AND ee.deleted_at IS NULL \n" +
+                "JOIN \n" +
+                "   exam.exam_status_codes esc \n" +
+                "   ON esc.status = ee.status \n" +
+                "LEFT JOIN \n" +
+                "   exam.exam_accommodation lang \n" +
+                "   ON lang.exam_id = e.id \n" +
+                "   AND lang.type = 'Language' \n" +
+                "   AND lang.created_at = ( \n" +
                 "       SELECT \n" +
                 "           MAX(eacc.created_at) \n" +
                 "       FROM \n" +
-                "          exam.exam_accommodation eacc \n" +
-                "          WHERE \n" +
-                "          eacc.exam_id = e.id \n" +
-                "          AND eacc.type = 'Language'\n" +
-                "  ) \n" +
+                "           exam.exam_accommodation eacc \n" +
+                "       WHERE \n" +
+                "           eacc.exam_id = e.id \n" +
+                "           AND eacc.type = 'Language') \n" +
                 "WHERE \n" +
                 "   e.student_id = :studentId \n" +
                 "   AND e.assessment_id = :assessmentId \n" +
                 "   AND e.client_name = :clientName \n" +
-                "   AND lang.type = 'Language'" +
+                "   AND lang.type = 'Language' \n" +
                 "ORDER BY \n" +
+                "   ee.id DESC, \n" +
                 "   e.created_at DESC \n" +
                 "LIMIT 1";
 
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -96,12 +96,13 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
                 "FROM exam.exam e\n" +
                 "JOIN ( \n" +
                 "   SELECT \n" +
-                "       exam_id, \n" +
-                "       MAX(id) AS id \n" +
+                "       id, \n" +
+                "       exam_id \n" +
                 "   FROM \n" +
                 "       exam.exam_event \n" +
                 "   WHERE exam_id = :examId\n" +
-                "   GROUP BY exam_id \n" +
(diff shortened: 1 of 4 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -93,15 +93,16 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
         String querySQL =
             "SELECT \n" +
                 EXAM_QUERY_COLUMN_LIST +
-                "FROM exam.exam e\n" +
+                "FROM exam.exam e \n" +
```
jeffjohnson9046: Formatting

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -111,16 +112,14 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 "  ON esc.status = ee.status \n" +
                 "LEFT JOIN exam.exam_accommodation lang \n" +
                 "  ON lang.exam_id = e.id \n" +
-                "  AND lang.created_at = \n" +
-                "  ( \n" +
+                "  AND lang.id = ( \n" +
                 "       SELECT \n" +
-                "           MAX(eacc.created_at) \n" +
+                "           MAX(eacc.id) \n" +
```
jeffjohnson9046: Changing from `created_at` to `id` causes the query to look at the primary key instead of an indexed column.  Since the `id` is auto-incrementing, the theory here is getting the max `id` is the same as getting the most recent `created_at` date.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -145,42 +144,36 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
         final SqlParameterSource parameters = new MapSqlParameterSource(queryParameters);
 
         String query =
-            "SELECT " +
+            "SELECT \n" +
```
jeffjohnson9046: formatting

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -145,42 +144,36 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
-                "JOIN exam.exam_event ee \n" +
-                "  ON last_event.exam_id = ee.exam_id AND \n" +
-                "     last_event.id = ee.id \n" +
-                "JOIN exam.exam_status_codes esc \n" +
-                "  ON esc.status = ee.status \n" +
-                "LEFT JOIN exam.exam_accommodation lang \n" +
-                "  ON lang.exam_id = e.id \n" +
-                "  AND lang.created_at = \n" +
-                "  ( \n" +
+                "FROM \n" +
```
jeffjohnson9046: In this case, we only want the most recent `exam` record, so we can join to `exam_event` directly instead of going through the usual process of finding the most recent `exam_event` record.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -201,52 +194,47 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
-                "       exam_page P\n" +
-                "   JOIN \n" +
-                "       exam_page_event PE\n" +
-                "   ON\n" +
-                "       P.id = PE.exam_page_id\n" +
-                "   WHERE \n" +
-                "       P.exam_id = :examId AND\n" +
-                "       PE.deleted_at IS NULL \n" +
-                ") AS lastStudentActivityTime";
+                "   MAX(ee.changed_at) AS lastStudentActivityTime \n" +
```
jeffjohnson9046: Removed the "derived table" code; you can order by the result of the `UNION ALL` in descending order and pick the most recent date.

Since there are several places we need to look for identifying recent activity within an exam, this query is difficult to optimize.  In a way, we don't know _exactly_ what we're looking for, but we know where to look.  The query to look at `exam_item_response` is particularly difficult; it's hard to narrow down what we're actually looking for.  If there's another way to detect recent activity, we should consider it.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -285,26 +273,27 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 "       MAX(id) AS id \n" +
                 "   FROM \n" +
                 "       exam.exam_event \n" +
-                "   GROUP BY exam_id \n" +
-                ") last_event \n" +
-                "  ON e.id = last_event.exam_id \n" +
+                "   WHERE \n" +
+                "       session_id = :sessionId" +
```
jeffjohnson9046: Adding the `session_id` here reduced the index scan to build this derived table by an order of magnitude in my test (from 10k to 999 rows).

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -285,26 +273,27 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
...
+                "   GROUP BY \n" +
+                "       exam_id \n" +
+                ")  last_event \n" +
+                "   ON e.id = last_event.exam_id \n" +
                 "JOIN exam.exam_event ee \n" +
-                "  ON last_event.exam_id = ee.exam_id AND \n" +
-                "     last_event.id = ee.id\n" +
+                "   ON last_event.exam_id = ee.exam_id \n" +
```
jeffjohnson9046: formatting

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -338,8 +327,8 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 ") last_event \n" +
                 "  ON exam.id = last_event.exam_id \n" +
                 "JOIN exam.exam_event ee \n" +
-                "  ON last_event.exam_id = ee.exam_id AND \n" +
-                "     last_event.id = ee.id \n" +
+                "  ON last_event.exam_id = ee.exam_id \n" +
+                "  AND last_event.id = ee.id \n" +
```
jeffjohnson9046: formatting

service/src/test/java/tds/exam/repositories/impl/ExamDataGenerator.java
```diff
@@ -0,0 +1,159 @@
+package tds.exam.repositories.impl;
```
jeffjohnson9046: Feel free to ignore this; I was using it to generate load up my local database w/some dummy exams.  Then I got the bright idea of getting a dump of the load test database, which was much better.  This class _may_ still come in handy, so I don't want to remove it just yet.  However, this class _will_ be removed prior to the final merge to **develop**.

service/src/test/java/tds/exam/repositories/impl/ExamDataGenerator.java
```diff
@@ -0,0 +1,159 @@
...
+import tds.exam.builder.ExamPageBuilder;
+import tds.exam.builder.ExamSegmentBuilder;
+import tds.exam.repositories.ExamAccommodationCommandRepository;
+import tds.exam.repositories.ExamCommandRepository;
+import tds.exam.repositories.ExamItemCommandRepository;
+import tds.exam.repositories.ExamPageCommandRepository;
+import tds.exam.repositories.ExamSegmentCommandRepository;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
```
millerts: Can you add` @Ignore`at a class level so it doesn't run?

PR #205, opened 2017-05-19
Fix language selection

wtritch:
This PR fixes two issues:  A hand-jammed SQL insert into an associated table in unit tests and ordering on a UUID column.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -9,6 +9,13 @@ import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.stereotype.Repository;
+import tds.common.data.mapping.ResultSetMapperUtility;
+import tds.common.data.mysql.UuidAdapter;
+import tds.exam.Exam;
+import tds.exam.ExamStatusCode;
+import tds.exam.ExamStatusStage;
+import tds.exam.models.Ability;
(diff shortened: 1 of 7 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -114,12 +113,14 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
                 "  ON lang.exam_id = e.id \n" +
                 "  AND lang.id = ( \n" +
                 "       SELECT \n" +
-                "           MAX(eacc.id) \n" +
```
wtritch: The "eacc.id" column is a UUID, so this was more-or-less randomly picking one of the language exam_accommodations.  I changed below to order by "created_date."

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -218,35 +218,34 @@ public void shouldReturnLastPausedDate() {
 
     @Test
     public void shouldFindExamWithLatestLanguage() throws InterruptedException {
-        Exam exam = new ExamBuilder().build();
+        final Exam exam = new ExamBuilder().build();
         examCommandRepository.insert(exam);
 
-        String SQL = "insert into exam_accommodation(id, exam_id, type, code, description, value, segment_position, default_accommodation, allow_combine, sort_order, created_at)\n" +
-            "values (:id, :examId, :type, :code, 'description', 'value', 1, 0, 0, 1, :createdAt);";
```
wtritch: This is generally a bad idea.  I'd added an exam_id field to the exam_accommodation_event table that wasn't being populated by this SQL.  It's generally better to go through the associated handlers so we don't end up breaking unassociated tests.

PR #206, opened 2017-05-19
Fixing issues with exam resumes before and after grace period

burzum619:
https://jira.fairwaytech.com/browse/TDS-983

This PR includes a number of bugfixes/missing features from the legacy system. With these fixes:
1. A student resuming an exam within the grace period should resume from the position of the last item that was answered. The should be able to access previous pages that were answered previously before the exam was paused.
2. A student resuming an exam outside of the grace period will be placed on the first unanswered item. They will not be able to access previously answered pages.

Some notes:
- Initially while writing /startExam, I was under the impression that resuming outside of the grace period would make the student start the exam from scratch. This is not the case. Instead the exam resumes from the same position it normally would, but prior pages become inaccessible.
- I noticed page duration wasn't being carried over into the database and I fixed this issue while addressing resumes.
- ExamPageWrapper now has an "isVisible" flag that will be used by the student application to determine whether or not the page can be accessed by an exam after a resume. If the exam is restarted outside of the grace period, this value will be "false" for all previously answered pages.
- Added a column to exam_page_event for the attempt restart number (exam_restarts_and_resumptions). This is the total number of restarts/gracePeriodRestarts. We can consider renaming "restarts" and "restartsAndResumptions" to simply "resumptions"/"gracePeriodResumptions" as there is essentially no such thing as restarting an exam from scratch in TDS (unless the exam is somehow submitted/completed), but I did not make this change part of this PR.

### DIFF
--- a/service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
@@ -42,6 +42,22 @@ public class ExamPageWrapperQueryRepositoryImpl implements ExamPageWrapperQueryR
         "   page.created_at, \n" +
         "   page.segment_key, \n" +
         "   page_event.started_at, \n" +
+        "   ( \n" +
+        "       CASE WHEN page_event.exam_restarts_and_resumptions = \n" +
+        "       ( \n" +
+        "           SELECT \n" +
+        "               restarts_and_resumptions\n" +
+        "           FROM \n" +
+        "               exam_event\n" +
+        "           WHERE \n" +
+        "               exam_id = :examId AND deleted_at IS NULL \n" +
+        "           ORDER BY \n" +
+        "               created_at\n" +
+        "           DESC LIMIT 1\n" +
+        "       ) THEN TRUE \n" +
+        "       ELSE FALSE \n" +
+        "       END \n" +
+        "   ) AS is_visible, " +
         "   item.id AS item_id, \n" +
         "   item.item_key, \n" +
         "   item.assessment_item_bank_key, \n" +
@@ -165,11 +181,11 @@ public class ExamPageWrapperQueryRepositoryImpl implements ExamPageWrapperQueryR
         @Override
         public List<ExamPageWrapper> extractData(ResultSet resultExtractor) throws SQLException, DataAccessException {
             Map<UUID, List<ExamItem>> itemsForPage = new HashMap<>();
-            Map<UUID, ExamPage> examPages = new HashMap<>();
+            Map<UUID, ExamPageVisiblityWrapper> examPageWrappers = new HashMap<>();
 
             while (resultExtractor.next()) {
                 UUID pageId = UUID.fromString(resultExtractor.getString("page_id"));
-                if (!examPages.containsKey(pageId)) {
+                if (!examPageWrappers.containsKey(pageId)) {
                     ExamPage page = new ExamPage.Builder()
                         .withId(UUID.fromString(resultExtractor.getString("page_id")))
                         .withPagePosition(resultExtractor.getInt("page_position"))
@@ -181,7 +197,7 @@ public class ExamPageWrapperQueryRepositoryImpl implements ExamPageWrapperQueryR
                         .withStartedAt(ResultSetMapperUtility.mapTimestampToJodaInstant(resultExtractor, "started_at"))
                         .build();
 
-                    examPages.put(pageId, page);
+                    examPageWrappers.put(pageId, new ExamPageVisiblityWrapper(page, resultExtractor.getBoolean("is_visible")));
                 }
 
                 // An item might not have a response
@@ -230,19 +246,38 @@ public class ExamPageWrapperQueryRepositoryImpl implements ExamPageWrapperQueryR
                     .build());
             }
 
-            return examPages
+            return examPageWrappers
                 .values()
                 .stream()
-                .map(examPage -> {
+                .map(visiblityWrapper -> {
+                    ExamPage examPage = visiblityWrapper.getExamPage();
                     List<ExamItem> items = itemsForPage.get(examPage.getId())
                         .stream()
                         .sorted(Comparator.comparingInt(ExamItem::getPosition))
                         .collect(Collectors.toList());
 
-                    return new ExamPageWrapper(examPage, items);
+                    return new ExamPageWrapper(examPage, items, visiblityWrapper.isVisible());
                 })
                 .sorted(Comparator.comparingInt(o -> o.getExamPage().getPagePosition()))
                 .collect(Collectors.toList());
+        }
+
+        private class ExamPageVisiblityWrapper {
+            private ExamPage examPage;
+            private boolean visible;
+
+            public ExamPageVisiblityWrapper(final ExamPage examPage, boolean visible) {
+                this.examPage = examPage;
+                this.visible = visible;
+            }
+
+            public ExamPage getExamPage() {
+                return examPage;
+            }
+
+            public boolean isVisible() {
+                return visible;
+            }
         }
     }
 }
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -40,6 +40,9 @@ import tds.exam.ExamApproval;
 import tds.exam.ExamAssessmentMetadata;
 import tds.exam.ExamConfiguration;
 import tds.exam.ExamInfo;
+import tds.exam.ExamItem;
+import tds.exam.ExamPage;
+import tds.exam.ExamSegment;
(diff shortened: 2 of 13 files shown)


service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
```diff
@@ -230,19 +246,38 @@ public ExamPageWrapperQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final
...
                         .collect(Collectors.toList());
 
-                    return new ExamPageWrapper(examPage, items);
+                    return new ExamPageWrapper(examPage, items, visiblityWrapper.isVisible());
                 })
                 .sorted(Comparator.comparingInt(o -> o.getExamPage().getPagePosition()))
                 .collect(Collectors.toList());
         }
+
+        private class ExamPageVisiblityWrapper {
```
burzum619: I created this private inner class simply to couple the isVisible flag with the exam page, and to prevent from having/accessing two separate hashmaps. Please feel free to suggest a better solution if there is one.

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
```diff
@@ -42,6 +42,22 @@
...
+        "               exam_event\n" +
+        "           WHERE \n" +
+        "               exam_id = :examId AND deleted_at IS NULL \n" +
+        "           ORDER BY \n" +
+        "               created_at\n" +
+        "           DESC LIMIT 1\n" +
+        "       ) THEN TRUE \n" +
+        "       ELSE FALSE \n" +
+        "       END \n" +
+        "   ) AS is_visible, " +
```
wtritch: Could I get a breakdown on what exactly this is doing?
the current page event is_visible IFF the current page event has the same number of restarts and resumptions as the last non-deleted exam_event?
I understand all the words I just typed, but I'm having difficulty wrapping my head around them.  Maybe a couple of example flows would help?

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
wtritch: Do we need this?  Doesn't ExamPage already have an isVisible property?

burzum619: Understandable. Yes, the original idea was to simply include the "restart count" of the exam page, and pass that on to the Student application. the "visibility" condition is based on the exam page's restart # matched up with the exam's current restart #. When these items are fetched in the /updateResponses flow in TDS_Student, we do not have the exam nor the current restart number, so we have to figure out if the exam page is visible on the ExamService side, as I'm doing here.

The alternative would be either to fetch the exam in another API call to ExamService and make the comparison on the student side, or we could see if we could persist the restart number somewhere on the student side... perhaps in the student cookie?

burzum619: But it would be easier to simply rename this "VisibilityWrapper" to something else :)

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
burzum619: You and me both! Essentially this comparison is done to determine if the exam page is one that should be visible/accessible in the UI for a student. In the startExam flow (before this query is ever called), the exam page's exam restart # is updated to the current exam's restart number so that they match. This update occurs if

1. If its a resume within a grace period, than all of the exam pages that were part of the last "restart" should be accessible.
2. If the exam is resumed after the grace period, then we only increment the exam pages that have not yet been answered/seen by the student

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
wtritch: I think this is a pretty big sign that we're trying to fetch too much in one query.  We shouldn't have two hash-maps going that we join at the end.
I think it's out-of-scope for this PR, but how would you feel about adding exam_id columns to the tables in here, selecting all the items/responses, then doing a second query to fetch the list of page_id's that they're referencing?  Joining across 3 models gives us a whooooole lotta duplication.  (Not to mention there are two full index scans in the sub-selects in this query.)

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
wtritch: Thanks for the explanation, that helps a ton!

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -606,6 +612,133 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+                        .anyMatch(item -> !item.getResponse().isPresent()
+                            || (item.getResponse().isPresent() && !item.getResponse().get().isValid()))
+                )
+                .map(pageWrapper ->
+                    ExamPage.Builder
+                        .fromExamPage(pageWrapper.getExamPage())
+                        .withExamRestartsAndResumptions(updateExamRestartsAndResumptions)
+                        .build()
+                )
+                .collect(Collectors.toList());
```
wtritch: That's a pretty big ternary.  Personal preference, but I might break that into an if/else for clarity.

service/src/main/java/tds/exam/repositories/impl/ExamPageWrapperQueryRepositoryImpl.java
burzum619: Agreed this code could use some refactoring if we see performance issues surrounding this logic, also agree we should probably do it outside of this PR though.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -606,6 +612,133 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+
+        examPageService.update(examPages.toArray(new ExamPage[examPages.size()]));
```
wtritch: This is why I generally dislike varargs.  They're super-convenient.  Until they're not. =P

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -606,6 +612,133 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+
+        // Find the first segment position where dateExited is not null or that has a satisfied permeable condition
+        /* StudentDLL.ResumeItemPosition_FN [5156] */
+        Optional<ExamSegment> maybeResumeExamSegment = examSegmentWrappers.stream()
+            .map(wrapper -> wrapper.getExamSegment())
+            .filter(examSegment -> examSegment.getExitedAt() == null || (examSegment.isPermeable() && examSegment.getRestorePermeableCondition() != null))
+            .min(Comparator.comparingInt(ExamSegment::getSegmentPosition));
+
+        // If there was a segment found with the above query
+        if (maybeResumeExamSegment.isPresent()) {
```
wtritch: maybe break this conditional block into a method for clarity?
```
if (maybeResumeExamSegment.isPresent()) {
    maybeResumeExamItem = findResumeExamItem(maybeResumeExamSegment.get());
    ...
}
```
See below for another potential option.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -606,6 +612,133 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        }
+
+        /* StudentDLL.ResumeItemPosition_FN [5193] */
+        // If all items had a response, find the highest item position regardless of whether it has a response or not
+        maybeResumeExamItem = examItemsInAssessment.stream()
+            .max(Comparator.comparingInt(ExamItem::getPosition));
+
+        if (maybeResumeExamItem.isPresent()) {
+            return maybeResumeExamItem.get().getPosition();
+        }
```
wtritch: The 704-732 block looks real similar to the 671-700 block.  Any way to consolidate?  Maybe by pulling out a method?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
burzum619: I tend to go overboard when I start lambda-ing.

millerts:
I've talked with @burzum619 about this offline, but let's leave this PR as is for now.  We'll discuss a bit offline on Monday and someone can finish it up while he's on vacation.

millerts:
Closing this PR since it is outdated and we replaced by a subsequent PR leveraging this code.

PR #207, opened 2017-05-19
TDS-999: Feature/exam page wrapper tuning

jeffjohnson9046:
[TDS-999](https://jira.fairwaytech.com/browse/TDS-999):  Query performance tuning for the `ExamPageWrapperQueryRepositoryImpl`.
* Added `exam_id` to the `exam_page_event` and `exam_item_response` tables
* Added indexes to help with the grouping of event records
* Removed `ORDER BY` in queries; we're sorting the `ExamPage` and `ExamItem`s in code after we hydrate the objects

With these changes, the execution plan for the `EXAM_PAGE_WRAPPER_WITH_ITEM_SELECT` query using the various predicates is all green.

### DIFF
--- a/client/src/main/java/tds/exam/ExamItemResponse.java
+++ b/client/src/main/java/tds/exam/ExamItemResponse.java
@@ -14,6 +14,7 @@ import static tds.common.util.Preconditions.checkNotNull;
 public class ExamItemResponse {
     private long id;
     private UUID examItemId;
+    private UUID examId;
     private String response;
     private int sequence;
     private boolean valid;
@@ -31,6 +32,7 @@ public class ExamItemResponse {
     private ExamItemResponse(Builder builder) {
         id = builder.id;
         examItemId = checkNotNull(builder.examItemId);
+        examId = checkNotNull(builder.examId);
         response = checkNotNull(builder.response);
         sequence = builder.sequence;
         valid = builder.valid;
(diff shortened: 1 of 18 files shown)


client/src/main/java/tds/exam/ExamItemResponse.java
```diff
@@ -65,22 +69,27 @@ public static Builder fromExamItemResponse(ExamItemResponse examItemResponse) {
...
 
 
-        public Builder withId(long id) {
+        public Builder withId(final long id) {
             this.id = id;
             return this;
         }
 
-        public Builder withExamItemId(UUID examItemId) {
+        public Builder withExamItemId(final UUID examItemId) {
```
jeffjohnson9046: Add `final` to builder parameters.  Only interesting add here is the `examId`

service/src/test/java/tds/exam/builder/ExamItemResponseBuilder.java
```diff
@@ -35,42 +37,47 @@ public ExamItemResponse build() {
             .build();
     }
 
-    public ExamItemResponseBuilder withId(long id) {
+    public ExamItemResponseBuilder withId(final long id) {
```
jeffjohnson9046: Added `final` to all builder methods

PR #208, opened 2017-05-22
Feature/populate student history

millerts:


service/src/main/java/tds/exam/configuration/item/selection/ItemSelectionConfiguration.java
```diff
@@ -15,7 +15,7 @@
 public class ItemSelectionConfiguration {
     @Bean
     @Qualifier("adaptiveSelector")
-    public ItemSelector getAdaptiveItemSelector(final SegmentService segmentService, final ItemCandidatesService itemCandidatesService){
+    public MsbItemSelector getAdaptiveItemSelector(final SegmentService segmentService, final ItemCandidatesService itemCandidatesService){
```
millerts: This was a change that happened in item selection.  MsbItemSelector is a class that has an extra method to allow MSB calculations.  The AdaptiveSelector2013 implements that interface.

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
```diff
@@ -54,4 +68,80 @@ public HistoryQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final NamedPar
...
+
+    @Override
+    public List<ItemGroupHistory> findPreviousItemGroups(final long studentId, final UUID excludedExamId, final String assessmentId) {
+        SqlParameterSource parameters = new MapSqlParameterSource("examId", excludedExamId)
+            .addValue("studentId", studentId)
+            .addValue("assessmentId", assessmentId);
+
+        final String SQL = "SELECT e.id as examId, item.group_id as groupId FROM exam_item\n" +
+            "JOIN exam e\n" +
+            "JOIN ( \n" +
```
millerts: @jeffjohnson9046 , should I add the `exam_id <> :examId` to the places where appropriate?

service/src/main/java/tds/exam/services/impl/ExamItemSelectionServiceImpl.java
```diff
@@ -73,7 +73,7 @@ public ExamItemSelectionServiceImpl(final ItemSelectionService itemSelectionServ
         if (response.getResponseStatus() == SATISFIED) {
             return Collections.emptyList();
         } else if (response.getErrorMessage().isPresent()) {
-            throw new RuntimeException("Failed to create item group: " + response.getErrorMessage());
+            throw new RuntimeException("Failed to create item group: " + response.getErrorMessage().get());
```
millerts: We needed to include the actual error message when things fail.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -239,6 +243,11 @@ public StudentHistory2013 loadOppHistory(UUID examId, String segmentKey) throws
                         }
                     }
 
+                    //Legacy doesn't handle null and expects non scored items to be -1 score
+                    if(response.getScore() == null) {
+                        response.setScore(-1);
```
millerts: In the legacy code score can't ever be null.  They default it to -1 which is what we now do here.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -268,8 +294,8 @@ public boolean setSegmentSatisfied(UUID examId, Integer segmentPosition, String
     @Override
     public OffGradeResponse addOffGradeItems(UUID examId, String designation, String segmentKey) throws ReturnStatusException {
         //AA_AddOffgradeItems_SP
-        //TODO - figure out if this is really necessary
-        return new OffGradeResponse(OffGradeResponse.SUCCESS, "");
+        //TODO - We are going to verify with SBAC if this is actually used
+        return new OffGradeResponse(OffGradeResponse.FAILED, "offgrade accommodation not exists");
```
millerts: This will be implemented in a follow up PR.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -980,4 +981,8 @@ private static boolean isSubjectBlocked(final List<RtsStudentPackageAttribute> b
 
     }
 
+    @Override
+    public List<ItemGroupHistory> findPreviousItemGroups(final long studentId, final UUID currentExamId, final String assessmentId) {
```
millerts: This will be removed because I decided to move this to a history service

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
wtritch: Shouldn't this have already been done by the scoring process?

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -250,6 +259,23 @@ public StudentHistory2013 loadOppHistory(UUID examId, String segmentKey) throws
         return history;
     }
 
+    /**
+     * Legacy implementation conversion.  The legacy objects use implementation classes rather than interfaces
+     */
+    private ArrayList<HashSet<String>> getPreviousItemGroups(final UUID examId, final long studentId, final String assessmentId) {
+        List<ItemGroupHistory> histories = examHistoryService.findPreviousItemGroups(studentId, examId, assessmentId);
+
+        ArrayList<HashSet<String>> itemGroups = new ArrayList<>();
```
wtritch: I know a lot of this is legacy stuff, but recommend using Set<> and List<> in variables, parameters and interfaces rather than ArrayList<> and HashSet<>.  (And preferring Collection<> over either unless order or uniqueness is part of the interface.)

millerts: I do too, but the legacy objects that we're using require the concrete classes.  Modifying them to use collections would ripple throughout the legacy code.

Trying to balance how much rewrite I have to do based on time remaining.

service/src/main/java/tds/exam/repositories/impl/HistoryQueryRepositoryImpl.java
jeffjohnson9046: I suggest adding the `exam_id <> :examId` where-ever we can.  That will still likely result in a scan of the index, but the "not equals" clause will cut down on the number of records that have to be scanned.

There is another story in the sprint () to review/tune the queries in the history repository, (TDS-1003), so adding the `exam_id <> :examId` will also help that story along.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
millerts: Not for freshly created items.  This is when we initially create the items so there is no score yet.  However, student/legacy code expects -1.

PR #210, opened 2017-05-23
Make sure the item response has an exam id

millerts:


service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -113,7 +113,7 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
 
         if (StringUtils.isNotEmpty(errorMessage)) {
             LOG.warn("Problem accessing exam item: {}", errorMessage);
-            final String message = configService.getFormattedMessage(examInstance.getClientName(), null, "T_UpdateScoredResponse", null);
+            final String message = configService.getFormattedMessage(examInstance.getClientName(), null, "T_UpdateScoredResponse");
```
millerts: This is removing a warning.

service/src/main/java/tds/exam/services/scoring/ResponseServiceImpl.java
```diff
@@ -126,6 +126,7 @@ public ReturnStatus updateScoredResponse(final ExamInstance examInstance,
         } else {
             updatedResponseBuilder = new ExamItemResponse.Builder()
                 .withExamItemId(existingExamItem.getId())
+                .withExamId(exam.getId())
```
millerts: examId was made required but this response wasn't setting it.  So any time we tried to save a response it would blow up.

service/src/test/java/tds/exam/services/scoring/ResponseServiceImplTest.java
```diff
@@ -141,4 +141,48 @@ public void itShouldSaveAPositiveScore() throws Exception {
         final ExamItemResponse persisted = responseCaptor.getValue();
         assertThat(persisted.getScore().get().getScore()).isEqualTo(1);
     }
+
+    @Test
+    public void itShouldSaveAnInitialPositiveScore() throws Exception {
+        final ExamInstance examInstance = ExamInstance.create(UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID(), "clientName");
```
millerts: The existing test covered the if an ExamItem had a response.  This tests if a new ExamItem that doesn't have a response is able to be committed.  This is the normal flow when an item is being answered for the first time.

PR #211, opened 2017-05-23
TDS-1001: Feature/exam segment tuning

jeffjohnson9046:
[TDS-1001](https://jira.fairwaytech.com/browse/TDS-1001):  Optimization for `ExamSegmentQueryRepositoryImpl`:
* Removed `ORDER BY`; the primary key will handle returning the data in the desired order
* Drop unused index
* Rename foreign key constraint


service/src/main/java/tds/exam/repositories/impl/ExamSegmentQueryRepositoryImpl.java
```diff
@@ -85,9 +85,7 @@ public ExamSegmentQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") final Name
                 "   exam_segment_event se \n" +
                 "ON \n" +
                 "   last_event.exam_id = se.exam_id AND \n" +
-                "   last_event.id = se.id \n" +
-                "ORDER BY \n" +
-                "   segment_position \n";
+                "   last_event.id = se.id";
```
jeffjohnson9046: The primary key will guarantee the order; no need for `ORDER BY`.

jtreuting: Meaning it will default sort by primary key?  Is that guaranteed?

jeffjohnson9046: That is guaranteed; the primary key dictates how the data is physically stored for the table.  If you just did `SELECT * FROM exam_segment`, it'd be sorted by `exam_id` then `segment_position`.

jtreuting: That is not what I am seeing here: https://dba.stackexchange.com/questions/6051/what-is-the-default-order-of-records-for-a-select-statement-in-mysql

Though admittedly most answers I'm finding via google are older.  I always thought there was no guarantee of order on disk or returned results because the DB engine can optimize as it sees fit and each DB engine might do things differently.

jeffjohnson9046: Here are my thoughts after reading those links (which raise some questions) and additional research I've done:
* SQL as a language does not guarantee sort order.  A `SELECT` statement without an accompanying `ORDER BY` clause does not in self guarantee any sort order.
  * This is especially true when when querying against a heap table (i.e. a table without a primary key)
* The primary key on an InnoDB table is a [clustered index](https://dev.mysql.com/doc/refman/5.7/en/glossary.html#glos_clustered_index), which means the data is "organized based on the values of the primary key columns".  Presumably, this organization is based on whatever collation the table/columns/server is configured for.
* InnoDB treats tables with clustered indexes as [index-organized tables](http://use-the-index-luke.com/sql/clustering/index-organized-clustered-index), meaning the primary key and the table data are [stored in the same structure](http://www.ovaistariq.net/521/understanding-innodb-clustered-indexes/#.WSiRdhPytuY)
  * To me, this implies that the data is physically stored by how the primary key is organized
* In order to keep the clustered index's btree optimized, data must be sorted correctly (again, presumably according to the collation settings).
  * this is why inserts on non-sequential primary keys can be expensive; the btree must be re-organized if a record is inserted "in the middle".
  * Non-sequential primary keys can result in excessive [page splitting](https://www.percona.com/blog/2017/04/10/innodb-page-merging-and-page-splitting/), where the pages of data on disk must be re-organized to accommodate the new record
  * This behavior further leads me to believe the data in the primary key is stored in a particular order

**To summarize:**  I'm willing to acknowledge that my statement was somewhat inaccurate; MySQL itself may not guarantee a sort order.  However, I'm confident that given how the data is stored within the primary key in combination with the collation settings will return the data in the correct order.  At least, I've never seen otherwise.

Having said all that, if we want to add the `ORDER BY` back in, we can.  

PS:  I'm also reasonably confident that the code consuming this data does not need the `exam_segment` records to be sorted by `segment_position`.

PR #214, opened 2017-05-26
Add caching around content service

millerts:


scoring/pom.xml
```diff
@@ -19,10 +19,19 @@
...
         <ss-multijar.version>3.0.6</ss-multijar.version>
         <item-scoring.version>3.0.1</item-scoring.version>
-        <item-renderer.version>3.0.9</item-renderer.version>
+        <item-renderer.version>3.0.10</item-renderer.version>
+        <tds-common.version>0.0.21</tds-common.version>
     </properties>
 
     <dependencies>
+        <!-- TDS -->
+        <dependency>
```
millerts: This is the only addition.  I just grouped dependencies together in the other changes below

wtritch: works for me.  But isn't the tds-common version declared higher up in the pom hierarchy?  Should it be?  We need the exam service and the scoring service to be using the same version of tds-common

millerts: Good catch.

PR #218, opened 2017-05-30
Bugfix/resumptions exam page visibility

millerts:
Leveraging the work @burzum619 did initially this is a slightly altered and what I'm hoping a little simpler implementation.

Instead of relying on resumptions instead the exam service will mark pages as not visible if the grace period has passed and the page has been answered in full.


service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -110,14 +109,4 @@
...
-
-    /**
-     * Finds the previous item groups administered in previous exams for the student and assessment id
-     *
-     * @param studentId     the student id
-     * @param currentExamId the current exam id
-     * @param assessmentId  the assessment id for the exam
-     * @return a list of {@link tds.exam.models.ItemGroupHistory} for previous exams taken by the student for the assessment id
-     */
-    List<ItemGroupHistory> findPreviousItemGroups(final long studentId, final UUID currentExamId, final String assessmentId);
```
millerts: This is left over and not actually implemented.  The real implementation lives in the history service.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -407,105 +414,6 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         return examAssessmentMetadatas;
     }
 
-    private List<ExamAssessmentMetadata> findEligibleExamAssessmentsForStudent(final long studentId, final Session session, final String grade) {
```
millerts: This was just moved so that the public methods are together.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+
+    /*
+        This method emulated StudentDLL.ResumeItemPosition_FN [5151]
+     */
+    private int findExamStartPosition(final List<ExamSegmentWrapper> examSegmentWrappers) {
+        Optional<ExamItem> maybeResumeExamItem;
+
+        // Find the first segment position where dateExited is not null or that has a satisfied permeable condition
+        /* StudentDLL.ResumeItemPosition_FN [5156] */
+        maybeResumeExamItem = getExamItemFromResumableSegment(examSegmentWrappers);
```
millerts: I'm not 100% sure this is necessary, but I left it in because it is in the legacy implementation.  It is very similar to the logic below but handles segment exits.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -945,6 +965,105 @@ private void updateExam(final Exam exam, final Exam updatedExam) {
         examStatusChangeListeners.forEach(listener -> listener.accept(exam, updatedExam));
     }
 
+    private List<ExamAssessmentMetadata> findEligibleExamAssessmentsForStudent(final long studentId, final Session session, final String grade) {
```
millerts: This is the moved method from above.

client/src/main/java/tds/exam/ExamPage.java
```diff
@@ -38,6 +39,8 @@ private ExamPage(Builder builder) {
         createdAt = builder.createdAt;
         deletedAt = builder.deletedAt;
         startedAt = builder.startedAt;
+        duration = builder.duration;
+        visible = builder.visible;
```
wtritch: Interesting, was duration just missing here before?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+    //If all items are answered and not in a grace period resume mark the pages as not visible
+    private void updateFinishedPages(final List<ExamSegmentWrapper> examSegmentWrappers, final boolean isGracePeriodResume) {
+        if(isGracePeriodResume) {
+            return;
+        }
+
+        List<ExamPage> examPages = examSegmentWrappers.stream()
+            .flatMap(segmentWrapper -> segmentWrapper.getExamPages().stream())
+            .filter(pageWrapper ->
+                // Find any pages that have items that have not yet been answered.
```
wtritch: nit: maybe change comment to "Filter out any incomplete pages" ?

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+                pageWrapper.getExamItems().stream()
+                    .noneMatch(item -> !item.getResponse().isPresent()
+                        || (item.getResponse().isPresent() && !item.getResponse().get().isValid()))
```
wtritch: I think this can be simplified to:
```
item -> item.getResponse().isPresent() && item.getResponse().get().isValid()
```

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+            || (examSegment.getExamSegment().isPermeable()
+            && examSegment.getExamSegment().getRestorePermeableCondition() != null);
+    }
+
+    private static boolean isItemResponseValid(final ExamItem examItem) {
+        return !examItem.getResponse().isPresent()
+            || (examItem.getResponse().isPresent()
+            && examItem.getResponse().get().getResponse().equals("")
+            && !examItem.getResponse().get().isValid());
+    }
```
wtritch: Why does this method return true when examItem.getResponse().get().isValid() returns false??

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        // If all items had a response, find the highest item position regardless of whether it has a response or not
+        ExamSegmentWrapper lastSegment = examSegmentWrappers.get(examSegmentWrappers.size() - 1);
+        ExamPageWrapper lastPage = lastSegment.getExamPages().get(lastSegment.getExamPages().size() - 1);
+        return lastPage.getExamItems().get(lastPage.getExamItems().size() - 1).getPosition();
+    }
+
+    private Optional<ExamItem> getExamItemFromResumableSegment(final List<ExamSegmentWrapper> examSegmentWrappers) {
+        Optional<ExamSegmentWrapper> maybeResumeExamSegment = examSegmentWrappers.stream()
+            .filter(this::isExamSegmentResumable)
+            .min(Comparator.comparingInt(o -> o.getExamSegment().getSegmentPosition()));
```
wtritch: If the segmentWrappers are already sorted, I don't think you need this min comparator, you could just grab the first.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        }
+
+        // flat map all exam pages, items for this assessment/restart number
+        List<ExamItem> examItemsWithoutResponse = examSegmentWrappers.stream()
+            .flatMap(wrapper -> wrapper.getExamPages().stream()
+                .filter(pageWrapper -> pageWrapper.getExamPage().isVisible())
+                .flatMap(examPage -> examPage.getExamItems().stream()
+                    .filter(ExamServiceImpl::isItemResponseValid)
+                ))
+            .collect(Collectors.toList());
```
wtritch: If we're only ever using the "first" of these, you don't have to stream the whole collection, just grab the first.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -607,6 +513,120 @@ public void pauseAllExamsInSession(final UUID sessionId) {
...
+        if(isGracePeriodResume) {
+            return;
+        }
+
+        List<ExamPage> examPages = examSegmentWrappers.stream()
+            .flatMap(segmentWrapper -> segmentWrapper.getExamPages().stream())
+            .filter(pageWrapper ->
+                // Find any pages that have items that have not yet been answered.
+                pageWrapper.getExamItems().stream()
+                    .noneMatch(item -> !item.getResponse().isPresent()
```
millerts: I'll change this to use the private static method isItemResponseValid like the other methods that do this item check.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Poorly named.

client/src/main/java/tds/exam/ExamPage.java
millerts: Yes, I think it was added because we knew we needed it but was not implemented.  Honestly, it was probably me and I forgot because duration is stored a bit differently on our side than on legacy.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: I've simplified it to the following:

```
    private static boolean isItemUnanswered(final ExamItem examItem) {
        return !examItem.getResponse().isPresent()
            || (examItem.getResponse().get().getResponse().equals("")
            && !examItem.getResponse().get().isValid());
    }
```

@jeffjohnson9046 I'm a bit confused by isValid.  Currently a exam_item_response` can have a populated `response` and still be invalid.  This logic (ported from legacy) seems to imply that the isValid matters for whether a question is answered.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
millerts: Good call.  This was a late modification of the previous work.

PR #220, opened 2017-05-31
Bugifx/irp fixing

millerts:


service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -264,6 +262,9 @@ public StudentHistory2013 loadOppHistory(UUID examId, String segmentKey) throws
                             response.setScore(score.getScore());
                             response.setScoreDimensions(score.getScoringDimensions());
                             response.loadDimensionScores(score.getScoringDimensions());
+                        } else {
+                            //Legacy scores are -1 not null
+                            response.setScore(-1);
```
millerts: Scores are never null in the legacy system.  They have a -1.  Added this for consistency since this flows through legacy item selection

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -508,8 +509,7 @@ private ItemCandidatesData convertSegmentHolderToItemCandidateData(final Expanda
             groupBlock.getBlock(),
             exam.getExam().getSessionId(),
             false,
-            //TODO - Find out where isActive is coming from in legacy code
-            null);
+            !examSegment.isSatisfied());
```
millerts: This is the big fix.  I didn't realize what isActive meant, but after looking at the item selection DLL isActive just means it hasn't been satisfied...

jtreuting: oh, yeah i can see that being confusing

PR #223, opened 2017-06-01
Fixed the issue with being able to resume paused exams.

millerts:
There were a few issues in the code that didn't allow for properly restarting an exam.  This is some really complicated code with multiple nesting and conditionals.  I definitely think if we have more time post release we should revisit and see if we can simplify it.

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -211,30 +211,22 @@ class ExamServiceImpl implements ExamService {
         Optional<Exam> maybePreviousExam = examQueryRepository.getLastAvailableExam(openExamRequest.getStudentId(), assessment.getAssessmentId(), currentSession.getClientName());
 
         if (maybePreviousExam.isPresent()) {
-            Optional<ValidationError> canOpenPreviousExamError = canOpenPreviousExam(maybePreviousExam.get(), currentSession);
+            PreviousExamStatus previousExamStatus = canOpenPreviousExam(maybePreviousExam.get(), currentSession);
             Exam previousExam = maybePreviousExam.get();
-            if (canOpenPreviousExamError.isPresent()) {
-                return new Response<>(canOpenPreviousExamError.get());
-            } else {
-                Optional<ValidationError> maybeOpenNewExamValidationError = canCreateNewExam(currentSession.getClientName(),
-                    openExamRequest, maybePreviousExam.get(), assessment.getMaxOpportunities());
-
-                // If we are able to open a previous, exam, but not a new one, we should resume our current exam
-                if (maybeOpenNewExamValidationError.isPresent()) {
-                    return openPreviousExam(currentSession.getClientName(), openExamRequest, maybePreviousExam.get(), assessment);
-                }
-                // Otherwise, open a fresh exam
-                return createExam(currentSession.getClientName(), openExamRequest, currentSession, assessment, externalSessionConfiguration, previousExam, student);
+            if (previousExamStatus.getValidationError().isPresent()) {
+                return new Response<>(previousExamStatus.getValidationError().get());
+            } else if(previousExamStatus.isResumable()) {
+                return openPreviousExam(currentSession.getClientName(), openExamRequest, previousExam, assessment);
             }
-        } else {
-            Exam previousExam = maybePreviousExam.orElse(null);
-            Optional<ValidationError> maybeOpenNewExamValidationError = canCreateNewExam(currentSession.getClientName(),
-                openExamRequest, previousExam, assessment.getMaxOpportunities());
-            if (maybeOpenNewExamValidationError.isPresent()) {
-                return new Response<>(maybeOpenNewExamValidationError.get());
-            }
-            return createExam(currentSession.getClientName(), openExamRequest, currentSession, assessment, externalSessionConfiguration, previousExam, student);
         }
+
+        Exam previousExam = maybePreviousExam.orElse(null);
+        Optional<ValidationError> maybeOpenNewExamValidationError = canCreateNewExam(currentSession.getClientName(),
+            openExamRequest, previousExam, assessment.getMaxOpportunities());
+        if (maybeOpenNewExamValidationError.isPresent()) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -211,30 +211,22 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
         Optional<Exam> maybePreviousExam = examQueryRepository.getLastAvailableExam(openExamRequest.getStudentId(), assessment.getAssessmentId(), currentSession.getClientName());
 
         if (maybePreviousExam.isPresent()) {
-            Optional<ValidationError> canOpenPreviousExamError = canOpenPreviousExam(maybePreviousExam.get(), currentSession);
+            PreviousExamStatus previousExamStatus = canOpenPreviousExam(maybePreviousExam.get(), currentSession);
```
millerts: After  walking through with @jeffjohnson9046 there are two results that can happen when verifying we can open a previous exam.  

1. A validation type error that the data isn't in a good state to restart an exam which breaks out of the flow entirely.
2. a boolean telling us if the previous exam can be reopened.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -820,24 +812,23 @@ private static ExamConfiguration getExamConfiguration(final Exam exam, final Ass
         return new Response<>(currentExam);
     }
 
-    private Optional<ValidationError> canOpenPreviousExam(final Exam previousExam,
-                                                          final Session currentSession) {
+    private PreviousExamStatus canOpenPreviousExam(final Exam previousExam, final Session currentSession) {
```
millerts: Lots of changes here, and this better reflects the code [here](https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/master/tds-dll-mysql/src/main/java/tds/dll/mysql/StudentDLL.java#L5488).

If you're comparing this to the legacy code then consider the boolean to be the equivalent to the ocnt (or numberRef) in the legacy code linked above.  The validation error is like the msgRef.

In the legacy code if the numberRef.value == 0 that would equate to false in the code below.  numberRef.value > 0 would equal true in our code.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -211,30 +211,22 @@ public ExamServiceImpl(ExamQueryRepository examQueryRepository,
...
-                // Otherwise, open a fresh exam
-                return createExam(currentSession.getClientName(), openExamRequest, currentSession, assessment, externalSessionConfiguration, previousExam, student);
-            }
-        } else {
-            Exam previousExam = maybePreviousExam.orElse(null);
-            Optional<ValidationError> maybeOpenNewExamValidationError = canCreateNewExam(currentSession.getClientName(),
-                openExamRequest, previousExam, assessment.getMaxOpportunities());
-            if (maybeOpenNewExamValidationError.isPresent()) {
-                return new Response<>(maybeOpenNewExamValidationError.get());
+            if (previousExamStatus.getValidationError().isPresent()) {
```
millerts: This code is much easier to understand.  The pink code above worked in a pretty hacky fashion.  Basically, it would check to see if it could open a previous exam.  If it didn't get an error it would then check if it could open a new exam.  If it couldn't open a new exam it would try to open the previous one otherwise it would open a new one.

So it worked but was not efficient.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -997,10 +988,12 @@ private void updateExam(final Exam exam, final Exam updatedExam) {
                     status = ExamStatusCode.STATUS_DENIED;
                 } else if (maybeMostRecentExam.isPresent()) {
                     Exam recentExam = maybeMostRecentExam.get();
-                    Optional<ValidationError> maybeError = canOpenPreviousExam(recentExam, examSessions.get(recentExam.getSessionId()));
+                    PreviousExamStatus previousExamStatus = canOpenPreviousExam(recentExam, session);
 
-                    if (maybeError.isPresent()) {
-                        deniedReason = maybeError.get().getMessage();
+                    if (previousExamStatus.getValidationError().isPresent()) {
```
millerts: And all the other changes above led to us being able to make this fix.  The denied reason is correct and @mtrupkin fixed that yesterday.  The next part is if the previous exam is resumable we need to set the current attempt to the resumed exam's attempt number.  That way it'll say "Resume 2 of 3 opportunities"

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -997,10 +988,12 @@ private void updateExam(final Exam exam, final Exam updatedExam) {
                     status = ExamStatusCode.STATUS_DENIED;
                 } else if (maybeMostRecentExam.isPresent()) {
                     Exam recentExam = maybeMostRecentExam.get();
-                    Optional<ValidationError> maybeError = canOpenPreviousExam(recentExam, examSessions.get(recentExam.getSessionId()));
+                    PreviousExamStatus previousExamStatus = canOpenPreviousExam(recentExam, session);
```
millerts: We got lucky with this one.  The previous exam check needs to be done using the current session not the previous one.  Double checked the legacy code where this was ported.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1669,6 +1669,7 @@ public void shouldFindEligibleExamsForStudent() {
         // Assessment that is enabled and student has already started
         AssessmentInfo assessmentInfo3 = AssessmentInfo.Builder
             .fromAssessmentInfo(random(AssessmentInfo.class))
+            .withId("id")
```
millerts: I would like to break apart this very large unit test because it is really hard to understand and maintain.  I may do that as a follow up merge, but for right now I left it as is adding a new validation for the fix above.

wtritch:
Thanks for all the comments, they really helped follow the flow.

PR #224, opened 2017-06-04
Bugfix/accommodation ignore case

millerts:


service/src/main/java/tds/exam/services/impl/ExamAccommodationServiceImpl.java
```diff
@@ -339,7 +339,7 @@ private static String getOtherAccommodationValue(String formattedValue) {
                 .filter(acc ->
                     acc.getExamId().equals(examAccommodation.getExamId())
                         && acc.getSegmentPosition() == examAccommodation.getSegmentPosition()
-                        && acc.getType().equals(examAccommodation.getType())
+                        && acc.getType().equalsIgnoreCase(examAccommodation.getType())
```
millerts: These were all the places I saw in Exam that we are doing the type check.

PR #225, opened 2017-06-05
Feature/enable msb

millerts:


service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -150,7 +150,7 @@ public ItemCandidatesData getItemCandidates(UUID examId) throws ReturnStatusExce
         }
 
         if (nonSatisfiedSegments.isEmpty()) {
-            return Collections.singletonList(new ItemCandidatesData(examId, IItemSelectionDLL.SATISFIED));
+            return Collections.singletonList(new ItemCandidatesData(examId, AlgorithmType.SATISFIED.getType()));
```
millerts: I was using the wrong satisfied value so [this](https://github.com/SmarterApp/TDS_ItemSelectionShell/blob/develop/tds-item-selection-components/src/main/java/tds/itemselection/services/impl/ItemSelectionServiceImpl.java#L67) was always failing.

millerts: And once we have the right value it flows to [this](https://github.com/SmarterApp/TDS_ItemSelectionShell/blob/develop/tds-item-selection-components/src/main/java/tds/itemselection/services/impl/ItemSelectionServiceImpl.java#L101) which is handled by the ExamItemSelection service

PR #231, opened 2017-06-07
Fixing MSB assessments - returning all ltemCandidates (instead of onl…

burzum619:
…y unsatisfied candidates)

Previously, satisfied segments were being filtered out of the list returned to the adaptive selector. Because of this, fixed form assessments were being processed incorrect as adaptive segments (since the first segment is always assumed to be adaptive) and the selection

This addresses the following:
TDS-1029
TDS-1030
TDS-1031

**This PR will NOT be merged into the develop for the release on 6/7/2017. This will be added as a patch post-release**

### DIFF
--- a/service/src/main/java/tds/exam/services/item/selection/ExamSegmentWrapper.java
+++ b/service/src/main/java/tds/exam/services/item/selection/ExamSegmentWrapper.java
@@ -14,7 +14,7 @@ import tds.exam.ExamSegment;
  * A Wrapper class that associates {@link tds.exam.ExamSegment} with its exams
  */
 class ExamSegmentWrapper {
-    private final ExamSegment examSegment;
+    private ExamSegment examSegment;
     private Set<UUID> pageIds = new HashSet<>();
     private List<ExamItem> items = new ArrayList<>();
 
@@ -65,5 +65,9 @@ class ExamSegmentWrapper {
      */
     public ExamSegment getExamSegment() {
         return examSegment;
+    }
+
+    public void setExamSegment(final ExamSegment examSegment) {
+        this.examSegment = examSegment;
     }
 }
--- a/service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


service/src/main/java/tds/exam/services/item/selection/ExamSegmentWrapper.java
```diff
@@ -66,4 +66,8 @@ public boolean isItemInSegment(final ExamItem examItem) {
     public ExamSegment getExamSegment() {
         return examSegment;
     }
+
+    public void setExamSegment(final ExamSegment examSegment) {
```
burzum619: Not a fan of breaking our immutability principal here, but it needed to be done to prevent some fairly-involved refactoring. This is done in order to give us a way to update the ExamSegment's "satisfied" flag in the wrapper before returning it tothe caller.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -92,7 +93,60 @@ public ItemCandidateServiceImpl(final ExpandableExamService expandableExamServic
     @Override
     public ItemCandidatesData getItemCandidates(UUID examId) throws ReturnStatusException {
         //Port of ItemSelectionDLL.AA_GetNextItemCandidates_SP
-        return getAllItemCandidates(examId).get(0);
+        ExpandableExam exam = expandableExamService.findExam(examId, ExpandableExamAttributes.EXAM_SEGMENTS, ExpandableExamAttributes.EXAM_PAGE_AND_ITEMS)
```
millerts: If anyone else reviews this code, the issue that @burzum619 and @hbgwhite found was that the original implementation was flawed because getting a single ItemCandidate has different functionality that getting all of them.  This method and the getAllItemCandidates are used heavily in the legacy item selection flow.  Ideally the getter would be much simpler than what we have here, but it would require a rewrite of item selection.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -92,7 +93,60 @@ public ItemCandidateServiceImpl(final ExpandableExamService expandableExamServic
...
+                && examSegment.getExamItemCount() == nonFieldTestItemCount
+                && fieldTestItemsSatisfied) {
+                satisfiedSegments.add(examSegmentWrapper);
+            } else if (examSegment.getAlgorithm().equals(Algorithm.FIXED_FORM)
+                && examSegment.getExamItemCount() == (nonFieldTestItemCount + fieldTestItemCount)) {
+                satisfiedSegments.add(examSegmentWrapper);
+            } else {
+                nonSatisfiedSegments.add(examSegmentWrapper);
+            }
+        }
```
millerts: Do you think we could extract lines 96-140 into a private method that could be used by this flow and the getAllItemCandidates?

```
private class ExamSegmentInfo {
  private List<ExamSegment> newlySatisfiedSegments;
  private List<ExamSegment> unsatisfiedSegments;
}

private ExamSegmentInfo findExamSegmentsNotYetSatisfied(UUID examId);
```

I know there are subtle differences between the overall methods but this block and the block below seem to be the same.  It is what happens after that information is fetched and parsed that the two methods deviate.

service/src/main/java/tds/exam/services/item/selection/ItemCandidateServiceImpl.java
```diff
@@ -135,26 +189,26 @@ public ItemCandidatesData getItemCandidates(UUID examId) throws ReturnStatusExce
...
             } else if (examSegment.getAlgorithm().equals(Algorithm.FIXED_FORM)
                 && examSegment.getExamItemCount() == (nonFieldTestItemCount + fieldTestItemCount)) {
-                satisfiedSegments.add(examSegment);
+                satisfiedSegments.add(examSegmentWrapper);
             } else {
                 nonSatisfiedSegments.add(examSegmentWrapper);
             }
         }
 
+        List<ExamSegmentWrapper> examSegments = Collections.emptyList();
```
millerts: Can we add a comment here explaining what is happening?  For example, after talking with you and @hbgwhite it became clear that this method needs to return all the ItemCandidates regardless if they are satisfied or not whereas the single item candidate service method only returns the first unsatisfied exam segment.

PR #237, opened 2017-06-08
Trimming quotes off browser user agent string

burzum619:
This user agent string is used by the TRT, and is expected to be unquoted

TDS-1034

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
@@ -451,8 +451,9 @@ class ExamServiceImpl implements ExamService {
         /* StudentDLL [5344] Skipping getInitialAbility() call here - the ability is retrieved in legacy but never set on TestConfig */
 
         if (exam.getStartedAt() == null) { // Start a new exam
-            // Initialize the segments in the exam and get the testlength.
-            Exam initializedExam = initializeExam(exam, assessment, browserUserAgent);
+            // Initialize the segments in the exam and get the testlength and trim the quotes off the user agent string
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -451,8 +451,9 @@ public void pauseAllExamsInSession(final UUID sessionId) {
         /* StudentDLL [5344] Skipping getInitialAbility() call here - the ability is retrieved in legacy but never set on TestConfig */
 
         if (exam.getStartedAt() == null) { // Start a new exam
-            // Initialize the segments in the exam and get the testlength.
-            Exam initializedExam = initializeExam(exam, assessment, browserUserAgent);
+            // Initialize the segments in the exam and get the testlength and trim the quotes off the user agent string
+            String unquotedUserAgent = browserUserAgent.replaceAll("^\"|\"$", "");
```
jtreuting: I guess I'm wondering why it is quoted in the first place?  Is it more appropriate to fix it at the source (depending on what that source is).

burzum619: This string is stored (in /openTest) and then retrieved from the student cookie on the TDS_Student side of things - I'm guessing that when its fetched back from the cookie, it comes back quoted. Do you think it makes more sense to fix this on the Student side instead?

millerts: My vote is to fix it in student as well.  I feel we can leave this here as a precaution but student shouldn't be sending the browser string as quoted

burzum619:
Closing this out to fix on student side

millerts:
Merge it.

PR #244, opened 2017-07-03
Feature/item document service

mtrupkin:


scoring/src/main/java/tds/score/services/impl/ContentServiceImpl.java
```diff
@@ -116,17 +115,16 @@ else if (rubricContentSource == RubricContentSource.ItemXML) {
...
-        try {
-            URI uri = new URI(xmlFilePath);
-            IITSDocument document = itemDocumentService.loadItemDocument(uri, accommodations.getValue(), true);
-            return document;
-        } catch (URISyntaxException e) {
-            throw new ReturnStatusException(e);
-        }
+    @Override
+    @Cacheable(CacheType.LONG_TERM)
+    public IITSDocument getContent(final String itemPath, final AccLookup accommodations) throws ReturnStatusException {
```
millerts: I think there should be a corresponding unit test change for this class.

millerts: I realize there may not be an existing test because we were cutting corners as we got close to the release, myself included, but we're post release so I think we should go back to writing tests.

scoring/src/main/java/tds/score/configuration/SupportApplicationConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required under applicable law or agreed to in writing, software distributed under the
+ * License is distributed in an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for specific language governing permissions
+ * and limitations under the license.
+ **************************************************************************************************/
+
+package tds.score.configuration;
+
+public class SupportApplicationConfiguration {
```
millerts: Why is this needed?  The defined constant is only used in the service impl?

scoring/src/main/java/tds/score/configuration/ScoringProperties.java
```diff
@@ -0,0 +1,29 @@
...
+
+package tds.score.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "exam-service")
+public class ScoringProperties {
+    private String contentUrl;
+
+    public String getContentUrl() {
```
millerts: Can you add a JavaDoc to this getter?

scoring/src/main/java/tds/score/configuration/ScoringConfiguration.java
```diff
@@ -33,6 +36,7 @@
 import tds.score.services.impl.ItemScoringServiceImpl;
 
 @Configuration
+@EnableConfigurationProperties(ScoringProperties.class)
```
millerts: Does Spring not inject the properties class directly once you add it to the getContentService `@Bean` annotated method?

scoring/src/main/java/tds/score/services/impl/ContentServiceImpl.java
mtrupkin: Agreed.  Was working on a unit test and got hung up.  I wanted to get the code reviewed while I get past the issues.

mtrupkin:
I see what you want with the ContentService mimicking the RubricService.  I can add that.

PR #249, opened 2017-08-16
Bugfix/npe fix

millerts:


millerts:
@jeffjohnson9046  and I paired and reviewed.

PR #255, opened 2017-09-01
Feature/itsdocument

mtrupkin:


scoring/src/test/java/tds/score/services/impl/ContentServiceImplTest.java
```diff
@@ -49,7 +49,7 @@ public void setup() {
     public void itShouldUseRepositoryToFindItemDocument() throws Exception {
         final String itemPath = "item/path.xml";
         final IITSDocument itemDocument = new ITSDocument();
-        itemDocument.setItemKey(1L);
+        itemDocument.setId(1L);
```
millerts: So why did this change?  Did the item document have Id and Itemkey and we consolidated?  Or is this a rename because we didn't want it to be itemkey?

mtrupkin: The concrete class ITSDocument override getItemKey and delegated to setId.

https://github.com/SmarterApp/TDS_ItemRenderer/blob/develop/TDS.ItemRenderer/src/main/java/tds/itemrenderer/data/ITSDocument.java#L33

PR #267, opened 2017-11-16
Handle the updates to the ChangeListener for exam updates

millerts:
This is a refactor that will make it easier to force submit exams which have a unique logic path when updating exams.  I needed a way to update multiple exams in a single go.  Rather than clutter the expiration logic in one massive PR this is a refactor to allow multiple updates

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
         <mysql.version>5.1.39</mysql.version>
         <joda-time.version>2.9.5</joda-time.version>
 
-        <tds-common.version>4.0.2-SNAPSHOT</tds-common.version>
+        <tds-common.version>4.0.2</tds-common.version>
         <tds-session.version>3.1.3.RELEASE</tds-session.version>
         <tds-assessment-client.version>4.0.0.RELEASE</tds-assessment-client.version>
         <tds-student-client.version>4.0.3.RELEASE</tds-student-client.version>
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
         <mysql.version>5.1.39</mysql.version>
         <joda-time.version>2.9.5</joda-time.version>
 
-        <tds-common.version>4.0.1.RELEASE</tds-common.version>
+        <tds-common.version>4.0.2-SNAPSHOT</tds-common.version>
         <tds-session.version>3.1.3.RELEASE</tds-session.version>
         <tds-assessment-client.version>4.0.0.RELEASE</tds-assessment-client.version>
         <tds-student-client.version>4.0.3.RELEASE</tds-student-client.version>
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -448,14 +448,6 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
         return jdbcTemplate.query(SQL, parameters, examRowMapper);
     }
 
-    public void findExamsToExpire() {
-        String QUERY_TO_GET_LATEST = "select * from exam.exam_event\n" +
-            "JOIN (select exam_id, max(id) as id\n" +
-            "from exam.exam_event \n" +
-            "group by exam_id) last_event on last_event.id = exam.exam_event.id\n" +
-            "where status != 'completed'";
-    }
-
     private static class AbilityRowMapper implements RowMapper<Ability> {
         @Override
         public Ability mapRow(ResultSet rs, int rowNum) throws SQLException {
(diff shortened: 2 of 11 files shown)


pom.xml
```diff
@@ -29,7 +29,7 @@
         <mysql.version>5.1.39</mysql.version>
         <joda-time.version>2.9.5</joda-time.version>
 
-        <tds-common.version>4.0.1.RELEASE</tds-common.version>
+        <tds-common.version>4.0.2-SNAPSHOT</tds-common.version>
```
millerts: I will release common and change this prior to merging

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -447,6 +448,14 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
         return jdbcTemplate.query(SQL, parameters, examRowMapper);
     }
 
+    public void findExamsToExpire() {
```
millerts: shoot, this should be in.  I'll remove this code

service/src/main/java/tds/exam/services/ExamService.java
```diff
@@ -122,4 +123,11 @@
      * @return The list of {@link tds.exam.Exam}s the student has taken
      */
     List<Exam> findAllExamsForStudent(final long studentId);
+
+    /**
+     * Updates exams
+     *
+     * @param examUpdates update exams
+     */
+    void updateExams(final List<EntityUpdate<Exam>> examUpdates);
```
millerts: Force submit will need the ability to update a bunch of exams at once

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
burzum619: Is this being used anywhere? Looks like this public method just declares some SQL that isn't used.

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1124,4 +1113,19 @@ boolean isResumable() {
...
         }
     }
+
+    private void updateExam(final EntityUpdate<Exam> examUpdate) {
+        updateExams(Collections.singletonList(examUpdate));
+    }
+
+    @Override
+    public void updateExams(final List<EntityUpdate<Exam>> examUpdates) {
+        List<Exam> examsToUpdate = examUpdates.stream()
```
millerts: When updating multiple exams we need an ability to track multiple old and new exams. For example, force submit will need to change the status to "complete" for `n` exams at one go.  The change listener needs the old and new.  The `EntityUpdate` object allows one to pair those two objects.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1687,7 +1688,7 @@ public void shouldReturnDeniedStatusIfExceededMaxAttempts() {
 
         final Response<List<ExamAssessmentMetadata>> response = examService.findExamAssessmentMetadata(studentId, sessionId, grade);
 
-        final List<ExamAssessmentMetadata> assessmentMetadata = response.getData().or(Collections.EMPTY_LIST);
+        final List<ExamAssessmentMetadata> assessmentMetadata = response.getData().or(Collections.emptyList());
```
millerts: I'm surprised the EMPTY_LIST constant is public but this removes a casting warning.

service/src/test/java/tds/exam/services/impl/ExamServiceImplTest.java
```diff
@@ -1670,7 +1671,7 @@ public void shouldReturnDeniedStatusIfExceededMaxAttempts() {
             .build();
 
         when(mockStudentService.findStudentPackageAttributes(eq(studentId), eq(currentSession.getClientName()), Matchers.<String>anyVararg()))
-            .thenReturn(Arrays.asList(new RtsStudentPackageAttribute(RtsStudentPackageAttribute.ELIGIBLE_ASSESSMENTS,
+            .thenReturn(Collections.singletonList(new RtsStudentPackageAttribute(RtsStudentPackageAttribute.ELIGIBLE_ASSESSMENTS,
```
millerts: Removed warning

service/src/main/java/tds/exam/services/ExamService.java
millerts: I chose to use a List rather than a varargs because of this [issue](https://stackoverflow.com/questions/12462079/potential-heap-pollution-via-varargs-parameter).

service/src/main/java/tds/exam/services/impl/ExamServiceImpl.java
```diff
@@ -1124,4 +1113,19 @@ boolean isResumable() {
             return Optional.ofNullable(validationError);
         }
     }
+
+    private void updateExam(final EntityUpdate<Exam> examUpdate) {
+        updateExams(Collections.singletonList(examUpdate));
+    }
+
+    @Override
```
millerts: This should be transactional

PR #268, opened 2017-11-17
Feature/force submit

millerts:


client/src/main/java/tds/exam/ExpiredExamInformation.java
```diff
@@ -0,0 +1,52 @@
+package tds.exam;
+
+import java.util.UUID;
+
+/**
+ * Information on the expired exam
+ */
+public class ExpiredExamInformation {
```
millerts: This is used as the response object when expiring exams.

service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
```diff
@@ -100,4 +100,21 @@
...
     List<Exam> findAllExamsForStudent(final long studentId);
+
+    /**
+     * Find all the exams to expire based on the status codes and the number of days since the last change event
+     *
+     * @param statusCodesToIgnore     the status codes to ignore
+     * @param daysFromLastChangeEvent the number of days since the last change event for a given exam
+     * @return the list of {@link tds.exam.Exam} to expire
+     */
+    List<Exam> findExamsToExpire(List<String> statusCodesToIgnore, int daysFromLastChangeEvent);
```
millerts: This one is no longer used but I wrote the code first so I just left it in.  I can remove it if people want.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    ExamExpirationServiceImpl(final ExamService examService, final ExamQueryRepository examQueryRepository, final TimeLimitConfigurationService timeLimitConfigurationService, final AssessmentService assessmentService) {
+        this.examService = examService;
+        this.examQueryRepository = examQueryRepository;
+        this.timeLimitConfigurationService = timeLimitConfigurationService;
+        this.assessmentService = assessmentService;
+    }
+
+    @Override
+    public Collection<ExpiredExamInformation> expireExams(final String clientName) {
+        //Retrieve the timelimits so we know how many days need to pass before expiration.  The system will be misconfigured
```
millerts: This is a port of the [DmDLL method](https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/develop/tds-dll-mysql/src/main/java/tds/dll/mysql/DmDLL.java#L63).  It isn't one to one since our data model differs from theirs.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+            return new HashSet<>();
+        }
+
+        //Get all the assessments for the exams to be expired since the "forceComplete" must be true
+       final  Map<String, Assessment> assessmentsByKey = examsToExpire.stream().map(Exam::getAssessmentKey)
+            .map(assessmentKey -> assessmentService.findAssessment(clientName, assessmentKey).get())
+            .collect(Collectors.toMap(Assessment::getKey, Function.identity()));
+
+        //Take the assessment ids from assessmentsByKey and get the timelimit configurations
+        final Map<String, TimeLimitConfiguration> assessmentIdToTimeLimits = assessmentsByKey.values().stream()
```
millerts: We need to get this because you can configure the time limit by client or narrow it more by assessment id.  We don't have the assessment ids we need until we get the exams to expire.  It is a bit of a pain to be honest.

There is a potential performance gain here if the config service had a way to get time limit configurations for a group of assessments rather than a single one.  Since this is a background process I didn't add it in.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        /*
+         - Filter out any exam that has an assessment if isn't force complete
+         - If time limit and last change aren't long enough apart
+         - create entity update
+         */
+        final Instant now = Instant.now();
+        List<EntityUpdate<Exam>> examUpdates = examsToExpire.stream()
+            .filter(exam -> {
+                Assessment assessment = assessmentsByKey.get(exam.getAssessmentKey());
+                return assessment.isForceComplete();
```
millerts: The `isForceComplete` is pulled from `configs.client_testproperties`

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        }).filter(exam -> {
+            TimeLimitConfiguration timeLimitConfiguration = assessmentIdToTimeLimits.getOrDefault(exam.getAssessmentId(), clientTimeLimitConfiguration);
+            return Days.daysBetween(exam.getChangedAt(), Instant.now()).isGreaterThan(Days.days(timeLimitConfiguration.getExamExpireDays()));
```
millerts: This is the day math.  Basically, time limits configure the number of days that need to pass prior to expiration.  We're checking the changed_at field to determine the last time the exam was modified.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+            return new EntityUpdate<>(exam, completedExam);
+        }).collect(Collectors.toList());
+
+        if(examUpdates.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        //Update the exams for completion
+        examService.updateExams(examUpdates);
```
millerts: We call this to set the completed at rather than the status method on exam service because we want to ignore the validation.  Basically, the client would like us to create the TRT's and not worry about the fact that the exam isn't 100% complete.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+                .withStatusChangeReason("Expiring exam via the force submittal process")
+                .build();
+
+            return new EntityUpdate<>(examEntityUpdate.getUpdatedEntity(), expiringExam);
+        }).collect(Collectors.toList());
+
+        //Update the exams and set them to expired
+        examService.updateExams(expiringExams);
+
+        return expiringExams.stream().map(examEntityUpdate -> {
```
millerts: So one thing that can happen with this flow is that the exam status can go from `expire` to `submit` or vice versa due to timing.  I think 99% of the time it'll go from `expire` to `submit` since I think this will execute faster than the rabbit and ERT will update the submit record.

service/src/main/java/tds/exam/web/endpoints/ExamExpirationController.java
```diff
@@ -0,0 +1,30 @@
...
+@RequestMapping("/exam")
+public class ExamExpirationController {
+    private final ExamExpirationService examExpirationService;
+
+    @Autowired
+    public ExamExpirationController(final ExamExpirationService examExpirationService) {
+        this.examExpirationService = examExpirationService;
+    }
+
+    @PutMapping(value = "expire/{clientName}", produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: So the url will be `exams/expire/SBAC_PT` or whatever client they choose.  I chose to make the SBAC_PT part of the path since it is required.  We could also make it a required request param.

PR #272, opened 2017-11-27
Fix an issue where the assessments were not being filtered to just th…

millerts:
…e unique keys

### DIFF
--- a/service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
+++ b/service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
@@ -58,12 +58,14 @@ public class ExamExpirationServiceImpl implements ExamExpirationService {
         List<Exam> examsToExpire = examQueryRepository.findExamsToExpire(STATUSES_TO_IGNORE_FOR_EXPIRATION);
 
         //None to expire return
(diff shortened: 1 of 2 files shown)


service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -58,12 +58,14 @@
...
-        if(examsToExpire.isEmpty()) {
+        if (examsToExpire.isEmpty()) {
             return new HashSet<>();
         }
 
         //Get all the assessments for the exams to be expired since the "forceComplete" must be true
-       final  Map<String, Assessment> assessmentsByKey = examsToExpire.stream().map(Exam::getAssessmentKey)
+        final Map<String, Assessment> assessmentsByKey = examsToExpire.stream()
+            .map(Exam::getAssessmentKey)
+            .distinct()
```
millerts: This was the change that fixed the 500 error.  I mistakenly thought `Collectors.toMap` would merge the list and not take duplicates but instead throws.  Now I distinct the assessmentKeys since there may be the same assessment key in many of the expired exams.

millerts:
Sorry for the formatting changes.  I didn't realize my original PR didn't follow our accepted coding standards.  I called out the change.

PR #280, opened 2017-12-27
Bugfix/make force submit work

millerts:
Due to my misunderstanding all force completed exams were completed then expired.  The logic we should have which this PR implements can be found [here](https://confluence.fairwaytech.com/display/SBAC1719/Force+Submit+Architecture).

### DIFF
--- a/client/src/main/java/tds/exam/ExpiredExamInformation.java
+++ b/client/src/main/java/tds/exam/ExpiredExamInformation.java
@@ -10,16 +10,18 @@ public class ExpiredExamInformation {
     private String assessmentKey;
     private String assessmentId;
     private UUID examId;
+    private String updatedExamStatus;
 
     //For frameworks
     ExpiredExamInformation() {
     }
 
-    public ExpiredExamInformation(final long studentId, final String assessmentKey, final String assessmentId, final UUID examId) {
+    public ExpiredExamInformation(final long studentId, final String assessmentKey, final String assessmentId, final UUID examId, final String updatedExamStatus) {
         this.studentId = studentId;
         this.assessmentKey = assessmentKey;
         this.assessmentId = assessmentId;
         this.examId = examId;
+        this.updatedExamStatus = updatedExamStatus;
     }
 
     /**
@@ -48,5 +50,12 @@ public class ExpiredExamInformation {
      */
     public UUID getExamId() {
         return examId;
+    }
+
+    /**
+     * @return the updated exam status after the process ran
+     */
+    public String getUpdatedExamStatus() {
+        return updatedExamStatus;
     }
 }
--- a/service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
+++ b/service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
@@ -15,7 +15,6 @@ package tds.exam.repositories;
 
 import org.joda.time.Instant;
 
-import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -107,8 +106,7 @@ public interface ExamQueryRepository {
      *
      * @param statusCodesToIgnore the status codes to ignore
      * @param expireExamLimit     the max number of exams to find
-     * @param assessmentIds       the assessment ids to include in the query
      * @return the list of {@link tds.exam.Exam} to expire
      */
-    List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit, Collection<String> assessmentIds);
+    List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit);
 }
--- a/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
@@ -27,7 +27,6 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -450,9 +449,8 @@ public class ExamQueryRepositoryImpl implements ExamQueryRepository {
     }
 
     @Override
-    public List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit, Collection<String> assessmentIds) {
+    public List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit) {
         final SqlParameterSource parameters = new MapSqlParameterSource("statusCodesToIgnore", statusCodesToIgnore)
-            .addValue("assessmentIds", assessmentIds)
             .addValue("expireExamLimit", expireExamLimit);
 
         /*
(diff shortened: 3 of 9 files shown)


client/src/main/java/tds/exam/ExpiredExamInformation.java
```diff
@@ -10,16 +10,18 @@
     private String assessmentKey;
     private String assessmentId;
     private UUID examId;
+    private String updatedExamStatus;
```
millerts: Added this field because we can now have "submitted" and "expired" exams.

service/src/main/java/tds/exam/repositories/ExamQueryRepository.java
```diff
@@ -107,8 +106,7 @@
      *
      * @param statusCodesToIgnore the status codes to ignore
      * @param expireExamLimit     the max number of exams to find
-     * @param assessmentIds       the assessment ids to include in the query
      * @return the list of {@link tds.exam.Exam} to expire
      */
-    List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit, Collection<String> assessmentIds);
+    List<Exam> findExamsToExpire(final List<String> statusCodesToIgnore, final int expireExamLimit);
```
millerts: A misunderstanding by me on the original logic had me thinking we only cared about exams that can be force submitted.  Exams that have been started but not completed can either be expired or completed based on the configuration.  Removed the assessment ids to get them all.

service/src/main/java/tds/exam/repositories/impl/ExamQueryRepositoryImpl.java
```diff
@@ -491,8 +489,6 @@ public ExamQueryRepositoryImpl(@Qualifier("queryJdbcTemplate") NamedParameterJdb
             "           AND type = 'Language' \n" +
             "       ORDER BY created_at DESC \n" +
             "       LIMIT 1) \n " +
-            "WHERE \n" +
```
millerts: This is the change.  I'm not sure what capitalized the SQL keywords.  Maybe a new feature of IntelliJ?

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -57,90 +58,104 @@
     @Override
     public ExpiredExamResponse expireExams(final String clientName) {
```
millerts: To understand what we're doing at a high level please look at the confluence page Overview section found [here](https://confluence.fairwaytech.com/display/SBAC1719/Force+Submit+Architecture).

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -57,90 +58,104 @@
...
 
-        //None to expire return
-        if (examsToExpire.isEmpty()) {
-            return new ExpiredExamResponse(false, Collections.emptyList());
-        }
+        //Get all assessment ids that have force complete enabled.
+        Collection<String> forceCompleteAssessmentIds = configService.findForceCompleteAssessmentIds(clientName);
+
+        //Get all the assessment ids for the exams so we can fetch the timelimit configurations for them
+        Set<String> expiringExamsAssessmentIds = examsToExpire.stream()
```
millerts: Now that we're considering all exams regardless if they can be force completed we need to get all the ids so we can get all the time limit configurations for those assessments.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -57,90 +58,104 @@
...
-            }).map(exam -> {
-                //Complete the exam
-                Exam completedExam = new Exam.Builder().fromExam(exam)
-                    .withCompletedAt(now)
-                    .withStatus(new ExamStatusCode(ExamStatusCode.STATUS_COMPLETED), now)
-                    .withStatusChangeReason("Completing exam via the force submit process")
+        Map<Boolean, List<Exam>> submitAndExpireExams = examsToExpire.stream().filter(exam -> {
+            TimeLimitConfiguration timeLimitConfiguration = assessmentIdToTimeLimits.getOrDefault(exam.getAssessmentId(), clientTimeLimitConfiguration);
+            return Days.daysBetween(exam.getChangedAt(), Instant.now()).isGreaterThan(Days.days(timeLimitConfiguration.getExamExpireDays()));
+        }).collect(Collectors.partitioningBy(exam -> forceCompleteAssessmentIds.contains(exam.getAssessmentId())));
```
millerts: The block above:
1. removes any exam that isn't ready to be expired based on it's assessment expire configuration
2. separates the exams into a list of exams that can be force completed an another that cannot

Exams that cannot be force completed are expired.

service/src/main/java/tds/exam/services/impl/ExamExpirationServiceImpl.java
```diff
@@ -57,90 +58,104 @@
...
-                exam.getAssessmentId(),
-                exam.getId());
-        }).collect(Collectors.toSet());
+        //Collect all the modified exams and send the list of data as a response
+        Collection<ExpiredExamInformation> expiredExamInformations = modifiedExams.stream().map((EntityUpdate<Exam> entityUpdate) -> new ExpiredExamInformation(
+            entityUpdate.getUpdatedEntity().getStudentId(),
+            entityUpdate.getUpdatedEntity().getAssessmentKey(),
+            entityUpdate.getUpdatedEntity().getAssessmentId(),
+            entityUpdate.getUpdatedEntity().getId(),
+            entityUpdate.getUpdatedEntity().getStatus().getCode())).collect(Collectors.toSet());
```
millerts: The main change in this block is that we now include the status.

service/src/main/java/tds/exam/utils/listeners/OnExpiredStatusExamChangeListener.java
```diff
@@ -0,0 +1,40 @@
...
+
+        // If the status has not changed between exam instances or the status has not already been set to "completed" on
+        // the new version of the exam, exit
+        if (oldExam.getStatus().equals(newExam.getStatus())
+            || !newExam.getStatus().getCode().equals(ExamStatusCode.STATUS_EXPIRED)) {
+            return;
+        }
+
+
+        messagingService.sendExamCompletion(newExam.getId());
```
millerts: This class was added because it seems that expired exams can also be submitted to the TRT process.  The legacy code is a bit hard to follow with all the temp tables but you can see it is submitting expired exams for the "QA Report", which is a TRT, [here](https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/develop/tds-dll-mysql/src/main/java/tds/dll/mysql/DmDLL.java#L139).

service/src/test/java/tds/exam/repositories/impl/ExamQueryRepositoryImplIntegrationTests.java
```diff
@@ -306,59 +306,6 @@ public void shouldFindExamsToExpireByStatusCodes() {
         assertThat(foundStarted).isTrue();
     }
 
-    @Test
-    public void shouldFindExamsToExpireByStatusCodesAndAssessmentId() {
```
millerts: This test is no longer needed because we no longer filter by assessment id which is what this test was testing.

client/src/main/java/tds/exam/ExpiredExamInformation.java
millerts: @jeffjohnson9046 is going to verify if this addition breaks Proctor.  @trapnine , I assume the python loader will just ignore or possibly print out this field without issue?

trapnine: Yes, the loader will just add an additional column to the csv output.

